PR #644, opened 2011-02-06
Add UCS-2 encoding support (aka the UTF-16 subset supported by JavaScript)

kkaefer:
It looks like v8 only supports little endian encoding (ä will be encoded as fc 00, not as 00 fc).


rolljee:
With Love :heart:

PR #1580, opened 2011-08-23
Fix for url.parse() leaving trailing ":" on the protocol/scheme

jordansissel:
Before:
    % node -e 'require("url").parse("http://www.google.com/").protocol'
    http:

After:
    % node -e 'require("url").parse("http://www.google.com/").protocol'
    http

### DIFF
--- a/lib/url.js
+++ b/lib/url.js
@@ -28,7 +28,7 @@ exports.format = urlFormat;
 
 // define these here so at least they only have to be
 // compiled once on the first module load.
-var protocolPattern = /^([a-z0-9]+:)/i,
+var protocolPattern = /^([a-z0-9]+):/i,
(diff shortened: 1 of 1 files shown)


billy-grates:
@github -1 on the ability to close comments after some period of time. As a firm believer in free speech, we should not let people stifle constructive conversation.

I've been a fan of using github as a source of high quality entertainment for over 5 years, you would be doing your key demographic a disservice in implementing said feature.

Also I feel like @gilesbowkett comments were on point.

revmischa:
thanks guys for going so deep into this colon issue. the library really needed some cleansing. felt like it was getting full of waste.
those colons can be a real pain in the you-know-what! too bad that this thread had to go down that dark hole.

PR #3036, opened 2012-03-30
Response object for HTTP Upgrade/CONNECT

stephank:
I started a thread on nodejs-dev about this, but it looks like my posts are still in the moderator queue. Will link it here once it gets through.

Briefly:
- The first commit is a test showing that a pipelined Upgrade request to a Node server can cause an out-of-order response. This is because the `upgrade` handler gets immediate access to the socket, while other requests may still be processing async.
- The second commit fixes this by providing a response object even for Upgrade, that is also queued properly along with other responses.
- In addition to fixing the test, I find it makes sense to have a response here. Per the spec, CONNECT and Upgrade **must** be followed by a proper response, before handover.

Regarding `make test`, I have one failing test both before and after this patch set, namely `simple/test-http-full-response`. I consider this a platform irk. `make test-debug` plain crashes the test suite on `simple/test-repl` (also before and after), but a fair amount of HTTP tests do pass.

### DIFF
--- a/doc/api/http.markdown
+++ b/doc/api/http.markdown
@@ -93,39 +93,37 @@ not be emitted.
 
 ### Event: 'connect'
 
-`function (request, socket, head) { }`
+`function (request, response) { }`
 
-Emitted each time a client requests a http CONNECT method. If this event isn't
-listened for, then clients requesting a CONNECT method will have their
-connections closed.
+Emitted each time a client requests a http CONNECT method. The arguments are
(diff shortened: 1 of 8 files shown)


jasnell:
Keeping this around and open for 2+ years after the original poster gave up on updating it isn't helpful. Looks like there's been recent albeit not particularly active discussion on it on the io.js side. Given that it's unlike this specific PR will ever land in v0.x, closing. Can reopen if necessary.

PR #3178, opened 2012-04-26
Adding support for .node_history and .noderc.js in $HOME and `pwd`

regality:
This adds support for .node_history, which behaves like a .bash_history for the node repl. When an interactive node quits, it writes ~/.node_history and will load it the next time you start the node repl.

This patch also allows a user to create a file called `.noderc.js` and place it in their home directory, or the directory they start node from.

When node is started with the `-i` flag or without arguments, it fires up the REPL.
At this point it will check if `~/.noderc.js` or `<pwd>/.noderc.js` exist. If they do, whatever objects they export will be copied into the repl context and be available for use.

A common use case for this will likely be loading debugging tools. For example I have the following as my noderc:

``` javascript
var util = require('util');

function inspect(obj) {
  console.log(util.inspect(obj, true, 10, true));
}

exports.p = inspect;
```

That way whenever I start the node repl, I can just type `p({some:'huge object'})` without having to retype basic boilerplate every time I start node.


robertkowalski:
I think this is fixed in https://github.com/nodejs/io.js/pull/1513

jasnell:
@robertkowalski ... should we backport or defer to the converged repo to pick it up?

Fishrock123:
Hmmm, might be best to at least choose one or the other with the combined TSC if you want to land here.

jasnell:
+1 ... at this point, recommend just closing this and picking up the io.js change in the converged stream.

cjihrig:
+1 to what @jasnell said.

robertkowalski:
+1 on closing

btw: if we would backport it it would be in node 0.13 right? and 0.13 does not really exist i think or did i miss somthing?

jasnell:
Yes, I'd assume v0.13. It's not yet entirely clear.

jasnell:
Ok, closing. If we end up doing a backport we'll pick it up in a different PR.

regality:
I haven't stayed up to date on all the node.js/io.js politics, so I don't know if it would be better to keep this pull or not, plus it is has been neglected and become very old. But it is worth noting that nodejs/io.js#1513 requires you to set `NODE_REPL_HISTORY` instead of just looking in your home directory, which seems silly to me. This pull request also creates the .noderc.js functionality, which was not replicated in the other pull that handles node history.

In short, closing this may be the best course of action, but moving some of the changes to a new pull request might also be a good idea.

Fishrock123:
Yeah, perhaps it;s best to open a new issue on io.js discussing that?

Do note getting the home directory is currently flawed until https://github.com/libuv/libuv/pull/350 lands.

PR #3315, opened 2012-05-24
Added that symlinks only work with windows vista or newer.

faridnsh:
Based on https://github.com/joyent/node/issues/3314#issuecomment-5899358

This doesn't have to be merged, but if something only works with certain versions, it should be in docs...


jasnell:
@piscisaureus ... is there anything from this we should land?

jasnell:
Closing. @alFReD-NSH if you feel this PR is still relevant, please rebase and update and we can reopen.

PR #3367, opened 2012-06-03
make it possible to locally configure cluster

tomyan:
The goal of the change is to make it possible to configure and use the cluster module without modifying global config. The changes leaves the current interface entirely in tact, but additionally allows configuration of a cluster with the following:

```
var cluster = require('cluster');

var master = new cluster.Master();
master.setupMaster({ exec: "some/file.js" });
```

Using the existing interface still works:

```
cluster.setupMaster({ exec: "some/other/file.js" });
```

..except that the global master instance has no effect on the other master, and vice versa. This is kinda similar to how you can configure a local http.Agent without affecting the global one.

The reason I wanted to propose this change was because I have a project that uses cluster to do some long running work. The project exposes a library interface and I would like users to be able to use that library interface without having side-effects other than the one they have initiated by using the library (i.e. does not modify global cluster config).


jasnell:
@tomyan ... I know it's been a while, but is this still something you want to pursue? If yes, the PR would need to be updated significantly. If not, we should likely go ahead and close.

jasnell:

PR #3499, opened 2012-06-20
Allow building node as a library

joelrbrandt:
Node currently builds great as a shared or static library (for embedding in other applications).

However, the build scripts don't make this easy.

This change adds an optional `--output-type` command line option to configure to allow specifying a static or shared library build. It also fixes up vcbuild to allow setting output type on Windows.

### DIFF
--- a/configure
+++ b/configure
@@ -165,6 +165,12 @@ parser.add_option("--with-arm-float-abi",
     help="Specifies which floating-point ABI to use. Valid values are: "
          "soft, softfp, hard")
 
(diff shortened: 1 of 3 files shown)


jasnell:
@joelrbrandt ... I know it's been a while, but is this still something you want to pursue? If yes, the PR would need to be updated significantly. If not, we should likely go ahead and close.

joelrbrandt:
@jasnell thanks for checking in. Brackets arrived at an alternate solution to leverage node. Closing.

PR #3643, opened 2012-07-05
Socket's setEncoding(null) removes Stream encoder.

Rush:
Also according to Isaac's tips I have added a test for the new behavior. `test/simple/test-net-socket-set-encoding-revert.js`


jasnell:
@Rush ... I know it's been a while, but is this still something you want to pursue? If yes, the PR would need to be updated significantly. If not, we should likely go ahead and close.

jasnell:
Closing this, but tagging it to keep track of. The PR is out of date. @Rush ... if this is something that you'd still like to pursue, please open a new, updated PR.

/cc @isaacs @bnoordhuis

PR #3696, opened 2012-07-12
json doc generator now generates _index.json as it creates files.

iapain:
`generate.js` now accepts `--output=<file>` which `tools/doc/json.js` uses to write the output. If output file is missing then it executes callback. This was required to get the directory for new `_index.json`. Ideally, I left it for you to decide what to call this `_index.json`. There is already an index.json file which should not be generated at all. Additionally it sets title as first heading node text if applicable.

TODO
- ~~Add exclude files to avoid generation of `index.json`~~
- Dependency map in Makefile (if subchapters like structure is required).


jasnell:
@iapain @chrisdickinson ... I know it's been a while, but is this still something you want to pursue? If yes, the PR would need to be updated significantly. If not, we should likely go ahead and close.

iapain:
@jasnell I will implement @chrisdickinson suggestions in neat way. Hold on.

jasnell:
@iapain :+1:

iapain:
@jasnell Ready to review.

iapain:
@jasnell I've to close this PR and will send you another one shortly. Looks like git rebase messed up with previous commits.

PR #4169, opened 2012-10-19
Implement dns.lookupAll() and use for net.connect()

### LINKED ISSUE #4168: dns/net: net.connect() only tries the first IP address found for a host
Common application behaviour when connecting to a remote host is to lookup that host in DNS (getaddrinfo) and then loop through the list of addresses returned to attempt to connect to one.

In node, net.connect() calls dns.lookup() which only ever returns the first A or AAAA address for the specified host.  This causes horrible problems when trying to connect to a host which returns both A and AAAA addresses when the client only has IPv4 or IPv6 (not both) connectivity;  worse still, this can be inconsistent as the return order of getaddrinfo can be arbitrary.

tomlanyon:
In hope that this fixes #4168, this implements dns.lookupAll() to return all A/AAAA RRs for a hostname retrieved from getaddrinfo() and subsequently sets net.connect() to use this when connecting to remote hosts.  This requires re-working of some of the other net connection code to retry with another host on failure and to setup a new TCP socket handle for each so hopefully it's all OK still.

This is my first node hacking; be gentle!

### DIFF
--- a/doc/api/dns.markdown
+++ b/doc/api/dns.markdown
@@ -3,12 +3,13 @@
     Stability: 3 - Stable
 
 Use `require('dns')` to access this module. All methods in the dns module
-use C-Ares except for `dns.lookup` which uses `getaddrinfo(3)` in a thread
-pool. C-Ares is much faster than `getaddrinfo` but the system resolver is
-more constant with how other programs operate. When a user does
-`net.connect(80, 'google.com')` or `http.get({ host: 'google.com' })` the
-`dns.lookup` method is used. Users who need to do a large number of look ups
-quickly should use the methods that go through C-Ares.
+use C-Ares except for `dns.lookup` and `dns.lookupAll` which use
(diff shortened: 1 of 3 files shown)


jasnell:
At this point, it's likely best to close this. There is work underway to refactor dns support entirely within io.js. That work would be picked up in the converged stream and this would need to be updated significantly.

@tomlanyon ... I know it's been a while, but is this still something you want to pursue? If yes, the PR would need to be updated significantly. If not, we should likely go ahead and close.

tomlanyon:
Happy for this to be closed.  I suggest that lookups in whatever DNS changes occur in io.js need to implement support for trying A and AAAA records, trying multiple records - potentially concurrently (happy eyeballs).

jasnell:
Noted! Closing

cjihrig:
I only read the title of this PR just now, so this could be something different, but `dns.lookup({all: true})` landed in io.js a while back.

PR #4249, opened 2012-11-07
Clarify comment about URI auth in mailto host

davidmurdoch:
The previous comment was difficult to understand upon first read and had a few typos.


jasnell:
Given that (a) this is targeted at master, (b) it's out of date and (c) code comments update are a low priority, I'm going to go ahead and close this. It's not critical to land in v0.12 or v0.10. If this is still something that should land, the right thing would be to open a PR against nodejs/node.

PR #4327, opened 2012-11-28
Add new http status code - 451

mbarinov:
HTTP Status Code for Legally-restricted Resources http://tools.ietf.org/html/draft-tbray-http-legally-restricted-status-00


jasnell:
Marking this as a "defer" to the converged repo. It's not likely to land here but going to keep it open for now.

jasnell:

PR #4338, opened 2012-11-30
Added debugging information to man page and help

gabrielf:
The man page for node and the help text when running node -h are now displaying
information about the --debug, --debug-brk and debug options.

Especially --debug and debug is important to distinguish since they look so
similar.


jasnell:
Closing because this does not appear to be relevant. Can reopen if that assumption is incorrect. Either way, it would need to be updated significantly

parshap:
The docs still don't seem to have any mention of `--debug-brk`. The contents of this PR seem relevant to me.

PR #4383, opened 2012-12-06
tls, https: add tls handshake timeout

bnoordhuis:
Don't allow connections to stall indefinitely if the SSL/TLS handshake does
not complete.

Adds a new tls.Server and https.Server configuration option, handshakeTimeout.

Fixes #4355.

Reviewers: @indutny @piscisaureus?


tangxinfa:
I use node.js v0.11.14, tls.Server not automatic close the underlaying tcp socket when handshake timeouted(https.Server and spdy.Server look's not have this problem), this caused my server has 60K dead tcp connections finally, to fix this problem you need destroy socket in the clientError event handler:

        tlsServer.on('clientError', function (exception, socket) {
            socket.destroy();
        });

Hopes this can help somebody.

PR #4644, opened 2013-01-22
dgram: add setMulticastInterface

austinf:
Depends on libuv pull request https://github.com/joyent/libuv/pull/694


jasnell:
Hmm... @austinf , looks like we're still waiting on review from @piscisaureus  ;-) this would need to be updated if you still wish to pursue. Going to close this but tagging it to keep track of.

PR #4773, opened 2013-02-14
Implement automatic out-of-tree-builds

iam-TJ:
This series of patches introduces automatic out-of-tree builds. Simply create and change to an alternate directory and call the configure script in the source tree directly. It will use the current working directory as the root of the build tree.

This series should include all the fixes to ensure the test suite works correctly in-tree or out-of-tree.


jasnell:

PR #4834, opened 2013-02-22
Adding support for util.extend(origin, [...]);

jimbojw:
The extend function is a staple of modern JavaScript
development. Yes, it's easy to implement, and everyone does it.
But functionality that everyone implements is exactly the kind
of thing that belongs in the core.

The extend function is so useful, in fact, that node already has
one, it just happens to be called _extend to make it secret.
Let's make it public and save everyone the burden of re-
inventing this particular wheel every time we make something.


osher:
@jimbojw that's an abuse of unecessary string manipulation. the bigger your object the more inefficient it is. I never recommend it.

PR #4882, opened 2013-03-01
Added interval and count options to setKeepAlive.

jameshartig:
See libuv pull request: https://github.com/joyent/libuv/pull/729


jasnell:
Given that (a) this is an API change that had been pushed out to v0.13 and (b) the PR needs updating before it can hope to land, I'm going to recommend closing this. If someone wishes to pursue this further, updating the PR and targeting it against either http://github.com/nodejs/io.js or http://github.com/nodejs/node would be the most appropriate. Closing for now but can reopen if there are objections. /cc @joyent/node-tsc

PR #4992, opened 2013-03-12
stream: Clear read buffer on unshift(null)

kanongil:
If you add an 'end' to the front of the read queue with unshift(null), logically the buffered data should be discarded, which this patch enables.


jasnell:

PR #5041, opened 2013-03-16
Added ability to run interactively with arguments

redchair123:
There currently is no way to run node interactively and pass arguments without
a file: `node -i foo bar baz` will instruct node to load `foo`.

This commit adds a `--` argument that instructs node to run interactive and not
to load any file in the argv:

```
$ node -i -- foo bar baz
> process.argv
[ '/the/path/to/node',
  '--',
  'foo',
  'bar',
  'baz' ]
```

The same thing works without the '-i':

```
$ node -- foo bar baz
> process.argv
[ '/the/path/to/node',
  '--',
  'foo',
  'bar',
  'baz' ]
```

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -2677,6 +2677,8 @@ static void ParseArgs(int argc, char **argv) {
       throw_deprecation = true;
     } else if (argv[i][0] != '-') {
       break;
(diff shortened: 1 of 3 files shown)


jasnell:

PR #5087, opened 2013-03-19
Dual stack for node.js

indutny:
Should be pulled only after joyent/libuv#749.

/cc @isaacs @bnoordhuis @piscisaureus.


jasnell:
@indutny ... any updates on this one?

indutny:
@jasnell not much, I'm still not sure if the happy eyeballs should be implemented in core, instead of the user-land module.

jasnell:
@indutny ... ok, well, given that active development has shifted over to nodejs/node, this is not likely to land here. Do you mind if we close this and if you want to pursue it then a new PR over there would make the most sense.

indutny:
@jasnell sure! I'd say it would be enough to just open an issue and cc me in nodejs/node.

BadgerBadgerBadgerBadger:
Where has this conversion been moved to? A link would help in keeping up with the status.

PR #5102, opened 2013-03-21
timers: fix setInterval() assert

bnoordhuis:
Test case:

```
var t = setInterval(function() {}, 1);
process.nextTick(t.unref);
```

Output:

```
Assertion failed: (args.Holder()->InternalFieldCount() > 0),
function Unref, file ../src/handle_wrap.cc, line 78.
```

setInterval() returns a binding layer object. Make it stop doing that,
wrap the raw process.binding('timer_wrap').Timer object in a Timeout
object.

Fixes #4261.

Reviewers: @isaacs @TooTallNate

### DIFF
--- a/lib/timers.js
+++ b/lib/timers.js
@@ -222,7 +222,7 @@ exports.setTimeout = function(callback, after) {
 exports.clearTimeout = function(timer) {
   if (timer && (timer.ontimeout || timer._onTimeout)) {
     timer.ontimeout = timer._onTimeout = null;
-    if (timer instanceof Timer || timer instanceof Timeout) {
+    if (timer instanceof Timeout) {
       timer.close(); // for after === 0
     } else {
       exports.unenroll(timer);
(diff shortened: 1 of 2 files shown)


tigran-avanesov:
Hello.

We experience fail of this test for our platform and I am currently trying to fix it, but I need to understand what is the expected behavior of this test.

__test/simple/test-timers-unref.js:60__
```
process.nextTick(t.unref.bind({}));
```
You are calling unref which expects __this__ to be __Timeout__ object, but bind an empty object instead. Which brings us to a situation where delay and _idleTimeout are __NaN__.
```JavaScript
  if (!this._handle) {
    var now = Date.now();
    if (!this._idleStart) this._idleStart = now;
    var delay = this._idleStart + this._idleTimeout - now;
    if (delay < 0) delay = 0;
    exports.unenroll(this);
    this._handle = new Timer();
    this._handle.ontimeout = this._onTimeout;
    this._handle.start(delay, 0);
    this._handle.domain = this.domain;
    this._handle.unref();
  } else {
    this._handle.unref();
  }
```

Could you please clarify?

Thanks in advance.

bnoordhuis:
`t.unref.bind({})` is nonsensical, it has no well-defined behavior, but it shouldn't SIGABRT the process.

I think it'd be alright instanceof checks in the right places in lib/timers.js.  If you do, target https://github.com/nodejs/node - this repo is effectively deprecated.

Issue #4932, opened 2013-03-06
many tests are too slow


isaacs:
There are a lot of tests that are too slow.  Some of them do 1000ms timeouts and stuff, but are perfectly fine to run much faster than that, especially when they should be responding to events rather than waiting on a setTimeout.  A full second is *way* too long for anything in test/simple/, since we run that very frequently.

It'd be great to have most tests run in under 50ms, or perhaps as much as 200ms for cases where timing is actually relevant to the test.

I'd suggest that some tests should just be retired, but every time I think that, some ancient test ends up catching a regression, so they should probably all stay.

jasnell:
Marking this as a potential iojs-backport, but closing. Can reopen if we feel this is still something worth pursuing in this stream.

PR #5214, opened 2013-04-03
test: Calculate timeout in milliseconds

paulfryzel:
This changes the timeout granularity from seconds to milliseconds. This is useful in issues such as #4932.

The default test timeout is now 30000ms (30s) per @isaacs suggestion, down from 60s.


jasnell:

PR #5274, opened 2013-04-11
node `-i` and `-` argument parsing

redchair123:
There currently is no way to run node interactively and pass arguments without
a file: `node -i foo bar baz` will instruct node to load `foo`.

This commit adds a `-` argument and modifies the behavior of the `-i` flag.
Informally, `-` means stdin and `-i` means run interactively.

When stdin is a tty (running directly without piping data into it):

1. `node - foo.js bar` will invoke the standard REPL without executing anything
beforehand.  `process.argv = ['node', '-', 'foo.js', 'bar']`

2. `node -i - foo.js bar` is the same as `node - foo.js bar`

3. `node -i foo.js bar` will run foo.js in the REPL context and then invoke the
standard REPL.  `process.argv = ['node', 'foo.js', 'bar']`

When stdin is a pipe or other file:

4. `| node - foo.js bar` will execute the commands piped into the shell.

5. `| node -i - foo.js bar` will execute the commands piped into the shell.

6. `| node -i foo.js bar` will first execute the commands in foo.js and then
execute the commands piped into the shell.

Note that this is slightly different from the behavior discussed in https://github.com/joyent/node/pull/5041.  It is possible to invoke behavior like the logic
in the `unbuffer` linux command, but that isn't implemented here.

### DIFF
--- a/lib/repl.js
+++ b/lib/repl.js
@@ -82,7 +82,7 @@ function REPLServer(prompt, stream, eval_, useGlobal, ignoreUndefined) {
 
   EventEmitter.call(this);
 
-  var options, input, output, dom;
+  var options, input, output, dom, paused = false;
   if (typeof prompt == 'object') {
     // an options object was given
     options = prompt;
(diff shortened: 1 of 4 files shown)


jasnell:

PR #5396, opened 2013-05-02
`process.pid` now using getter (calling `getpid()` internally)

me-andre:
`process.pid` has been internally changed to use a getter instead of being simple property which is set at app startup. This change enables us to use posix `fork` & `daemonize` and still get correct process id. While this change may be considered as a source of possible performance impact, I personally don't think so, because we usually need PID just once or twice per app.


jasnell:

PR #5481, opened 2013-05-15
readline: add setEcho function to readline

OJezu:
setEcho allows disabling (or enabling) input echoing (visibility) in readline as
user types it into the terminal

backward compatibility preserved


jasnell:

PR #5579, opened 2013-05-28
readline: colorized lines auto-completion support

nuh-temp:
Added support for using colorized strings (https://github.com/marak/colors.js) as options for completer.


jasnell:

Issue #5344, opened 2013-04-21
Need way to disable REPL ANSI escape codes


jart:
The repl.js module needs to be smarter about how it sends ansi escape codes. For example if I run "coffee -i" (a command which uses the standard repl module) inside a dumb terminal that doesn't support ansi escape sequences, I see the following:

```
^[[1G^[[0Jcoffee> ^[[9G
```

This is ugly! Those ansi escape codes are unnecessary. Furthermore the repl module is not respecting the fact that I set NODE_NO_READLINE=1 to try and get rid of those escape codes. Because I'm running the coffee command, I do not have the ability to pass the "terminal: false" parameter to the repl.start() function so I have no way of making these codes go away.

Therefore I propose:
1. If NODE_NO_READLINE is set, repl will not print ansi codes.
2. If process.env['TERM'] is 'dumb' then repl will not print ansi codes.

jonslenkooyala:
I'm running node v0.12.2 on Mac OS X 10.9.5, inside a bash instance inside emacs Ok, make fun of me). I am still getting (color) escape codes, I think??

tmp 509 $ echo $NODE_DISABLE_COLORS
1
tmp 510 $ node -v
v0.12.2
tmp 511 $ node
^[[1G^[[0J> ^[[3G
tmp 512 $ echo $TERM
dumb

PR #5602, opened 2013-05-30
set isTTY to false if TERM is 'dumb'

saljam:
This is to stop REPLs/logs from spitting ANSI control codes on terminals which don't support them. (issue #5344)

I'm not sure if this is the best approach (I'm new to node), but it does the job for me.

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -129,7 +129,7 @@
           useGlobal: true,
           ignoreUndefined: false
         };
-        if (parseInt(process.env['NODE_NO_READLINE'], 10)) {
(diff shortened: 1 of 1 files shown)


jasnell:
@saljam ... I know it's been a while, but is this still an issue as far as you can tell? And if so, are you interested in updating the PR?

saljam:
@jasnell seems like node still spits out control codes when TERM=dumb, so yeah I think this is still an issue.

I've since deleted my fork on GitHub, so I'm not sure how to update the commits in this one. Should I issue a new PR? It's just a rebase.

saljam:
OK abandoned in favour of #25506.

PR #5637, opened 2013-06-07
debugger: also exit when the repl emits 'exit'

### LINKED ISSUE #5631: Repl in node debug can only be exited using ctrl-c
Node debug allows you to drop into a repl at your current location using the `repl` command. To exit this, the only allowed method is to use `ctrl-c` since `.exit` does not work. As shown [here](https://github.com/joyent/node/blob/master/lib/_debugger.js#L1545), the repl listens to SIGINT, but sending it a sigint with kill, breaks the debugger. 

I'm not sure, but [digging into the repl source](https://github.com/joyent/node/blob/master/lib/repl.js#L169) seems to indicate that it only pays attention to SIGINT from the attached terminal, instead of in general. This is a problem if the debugger is a child process of a script, and SIGINT on the script causes an exit.

If `.exit` acted the same as `ctrl-c` during debugging, this problem would be solved.

fb55:
Fixes #5631

### DIFF
--- a/lib/_debugger.js
+++ b/lib/_debugger.js
@@ -1540,9 +1540,8 @@ Interface.prototype.repl = function() {
   // Don't display any default messages
   var listeners = this.repl.rli.listeners('SIGINT').slice(0);
   this.repl.rli.removeAllListeners('SIGINT');
-
-  // Exit debug repl on Ctrl + C
-  this.repl.rli.once('SIGINT', function() {
+  
+  function exitDebugRepl(){
     // Restore all listeners
     process.nextTick(function() {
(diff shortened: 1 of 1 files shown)


jasnell:
@cjihrig ... this is targeted at master. Do we want to push this off to nodejs/node or get this landed in v0.12?

cjihrig:
@jasnell let's take it to nodejs/node.

jasnell:
@fb55... I know it's been a while, but are you willing to update this and open it against master in http://github.com/nodejs/node?

fb55:
@jasnell /done

jasnell:
Thanks!

PR #5710, opened 2013-06-17
doc: add description of legacy buf.write() format

namhyung:
The buf.write() method also supports undocumented legacy style invocation.
As the example of buf.length property uses this form, a reader can be confused
with it.  Add description of these styles at the end.


jasnell:
Closing due to lack of activity. Can revisit if someone is interested in updating the PR or if someone feels this is still relevant.

PR #5717, opened 2013-06-19
http: forbid double 'close' event

egirshov:
In 0.8 branch there is this corner case when 'close' event is emitted twice from both ServerResponse and underlying Socket (response is written after socket is already closed by the client).
The same test case works fine in v0.10 or master without any fixes.


jasnell:

PR #5741, opened 2013-06-25
add ability to add more stdio channels in child_process.fork and cluster.fork

EugenDueck:
Export ability (partly) of `child_process.spawn()` to specify fds / channels between parent and child to `child_process.fork()` and `cluster.fork()`, in reference to issue #5727

### DIFF
--- a/doc/api/child_process.markdown
+++ b/doc/api/child_process.markdown
@@ -1,4 +1,4 @@
-# Child Process
+ Child Process
 
     Stability: 3 - Stable
 
@@ -544,6 +544,7 @@ leaner than `child_process.exec`. It has the same options.
   * `env` {Object} Environment key-value pairs
   * `encoding` {String} (Default: 'utf8')
   * `execPath` {String} Executable used to create the child process
+  * `addStdio` {Array} Additional child stdio fds / channels.
 * Return: ChildProcess object
 
 This is a special case of the `spawn()` functionality for spawning Node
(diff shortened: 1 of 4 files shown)


moll:
Is this issue still alive? I've got the same wish to set fork options when spawning new cluster workers.

sam-github:
Why was env made mandatory? Seems a pointless incompatibility which will break almost all uses of cluster.

moll:
Perhaps we're back to needing a user-space implementation of Cluster? The Cluster module could provide a few primitives, but that way the whole connection scheduling et al. features could be improved upon more frequently.

EugenDueck:
@sam-github Fair enough point. In actual fact, `env` is only mandatory (but can be null) when `options` is provided. But if env is not provided, behavior has not changed. Not sure how to document this. It's like two method overloads.

```
cluster.fork([env]) // overload 1
  ...
cluster.fork(env, options) // overload 2
  ...
```

sam-github:
@EugenDueck 

`cluster.fork([env[, options]])`

jasnell:
@EugenDueck @sam-github ... given that this is targeted at master and has conflicts, if this is still something that should land, it needs to be updated and reopened against nodejs/node. Recommend closing this here.

jasnell:
Closing this here. @EugenDueck, if you'd like to pursue this API change, please update and open a new PR against http://github.com/nodejs/node master. Thank you!

PR #5757, opened 2013-06-27
cluster: Multiple calls to worker.disconnect() should still wait for server close events

groner:
The current worker.disconnect() code includes a check for server._handle to determine if the server is already closed.

This PR also checks server._connections to see if there are open connections remaining.  It also extends the simple/test-cluster-worker-disconnect.js test to check that open connections keep the worker alive.


jasnell:
Closing due to lack of activity. Can revisit if someone is interested in updating the PR or if someone feels this is still relevant

PR #5771, opened 2013-07-01
http: Send GET request with a body as chunked

servant-42:
If a GET with a body is being sent and it is in implicit headers mode,
the body will be sent chunked by default (if a Content-Length is
not specified). Otherwise it would result in a request with a body and
no Content-Length header, which is invalid and causes a socket hang up
error.

The HTTP 1.1 spec allows sending a body with a GET request, but the
server isn't allowed to attach meaning to it because GET requests are
idempotent.

Fixes #5767 (more discussion there as well).


jasnell:

PR #5805, opened 2013-07-06
doc: update JSON output of API docs to unstable

luk-:
Minor doc change, moving from Experimental to Unstable for 
JSON output and removing reference to node v0.6.12.


jasnell:
@joyent/node-tsc ... this isn't exactly critical, but what is the status on the JSON output?

PR #5816, opened 2013-07-08
http: fix chunk length calculation when using a custom encoding

rustyconover:
Fix the calculation of http chunk length when writing a string with an encoding.

Added test.

Fixes issue #5777.


jasnell:
@trevnorris @isaacs ... what do we want to do with this one?

jasnell:
Looking at the code, it appears that some variation of this fix already landed at some point. Recommend closing.

jasnell:
Closing this. Some testing shows that the current behavior in nodejs/node master is correct that that this change is no longer necessary. It could still be possible that we need to backport to v0.12 or v0.10 but that would need to be done in a different PR.

PR #5818, opened 2013-07-09
http: fix chunk length calculation if write was called with encoding.

rustyconover:
The chunk length calculation when using chunked transfer encoding
and a string encoding of Base64 was incorrect.  This patch properly
calculates the chunk length, by getting the number of bytes when
the string is decoded rather then just the length of the string.

Patch is for v0.10.

Fixes issue: #5777


jasnell:
Looking at the code, I believe a variation on this fix was already landed. Given the conflicts, recommend closing. @joyent/node-tsc

jasnell:
Closing this here for now. Can revisit later.

PR #5869, opened 2013-07-19
child_process: Set additional fds as read/write

sethfitz:
When a `ChildProcess`'s `stdio` configuration includes additional file descriptors, those fds should be set as read/write to prevent `uv_shutdown from setting`ENOTCONN`when calling end() on the`Socket` wrapper.

Pre-libuv handling: https://github.com/joyent/node/blob/8417870f513e6f1ab98e07fe95d3430dbec48b87/lib/net.js#L851-L853

Where `uv_shutdown` sets `ENOTCONN`: https://github.com/joyent/libuv/blob/3ee4d3f183331a123ce35edd0d32268a2bb22aa5/src/unix/stream.c#L1090-L1095

This still needs tests, but I'm curious what y'all think (and to put my thinking down).

I'm seeing it when I pipe multiple `Buffer`s to ImageMagick via additional fds using `Socket#end()`.  (See "STDIN, STDOUT, and file descriptors" in http://www.imagemagick.org/script/command-line-processing.php)

### DIFF
--- a/lib/child_process.js
+++ b/lib/child_process.js
@@ -62,16 +62,11 @@ function createPipe(ipc) {
   return new handleWraps.Pipe(ipc);
 }
 
-function createSocket(pipe, readable) {
+function createSocket(pipe, readable, writable) {
   var s = new net.Socket({ handle: pipe });
 
-  if (readable) {
-    s.writable = false;
-    s.readable = true;
-  } else {
-    s.writable = true;
-    s.readable = false;
-  }
+  s.writable = writable;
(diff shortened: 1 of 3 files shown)


jasnell:
@mojodna ... given that (a) this targets master (which is no longer the active dev channel), (b) involves a change in behavior and (c) would need to be updated, I'm going to close this. If this is still something you'd like to pursue, please update and open a PR against http://github.com/nodejs/node. Thanks!

PR #5940, opened 2013-07-29
trace: remove macro usage for probe points

tjfontaine:
Unconditionally compile `src/node_dtrace.cc` but only export the
functions the configuration supports. Then `lib/_trace.js` replaces any
missing probe points with a nop function.

This is a companion commit to #5938


jasnell:
Closing due to lack of activity. Can revisit if someone is interested in updating the PR. (this may still be a good idea, just hard to determine and given that there's been zero activity....)

PR #6004, opened 2013-08-06
cluster: introduce `distribute` option

### LINKED ISSUE #6001: cluster: implement API for pluggable distribute() for round-robin scheduler
Looks usefull to provide API in the cluster module to define user own [`distribute`](https://github.com/joyent/node/blob/22c68fdc1dae40f0ed9c71a02f66e5b2c6353691/lib/cluster.js#L161) method.

Something like following:

``` javascript
cluster.setupMaster({
  schedulingPolicy: cluster.SCHED_RR,
  distributeRequest: function() {
    // ...
    return worker;
  }
});
```

Previous discussion with @bnoordhuis in the issue #4435 (after closing)

indutny:
`distribute` is a user-specified callback for asynchronous balancing of
incoming connections to cluster workers.

fix #6001

/cc @bnoordhuis @isaacs @kaero

### DIFF
--- a/doc/api/cluster.markdown
+++ b/doc/api/cluster.markdown
@@ -126,6 +126,12 @@ values are `"rr"` and `"none"`.
     (Default=`process.argv.slice(2)`)
   * `silent` {Boolean} whether or not to send output to parent's stdio.
     (Default=`false`)
+  * `distribute` {Function} user-specified server connection distribution
+    function. Invoked with only one argument - `callback`, which it should
+    invoke in a following way: `callback(err, worker)`, where the `worker`
+    argument is an instance of `Worker` class and specifies to which worker
+    a request should be scheduled.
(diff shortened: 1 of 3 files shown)


jasnell:
@indutny ... any further progress on this one? Is this still needed?

indutny:
cc @kaero ?

kaero:
Requests distribution problem is still here, but for a long time we successfully use dual-layer load balancing. RR at proxy level (nginx/haproxy/...) is used to distribute requests between small groups of workers and on node.js level, we just use default behavior – it works fine.
Dynamic RR between single node.js workers at proxy level is works well too.
So maybe it's not a node.js problem, to provide such a method.  `¯\_(ツ)_/¯`

jasnell:
Ok. With that in mind then, I'm going to close this. We can revisit if anything thinks having this in core is necessary.

PR #6011, opened 2013-08-07
implement domain-like hooks (asynclistener) for userland

trevnorris:
Here's the current API planned for implementation:

``` javascript
// Class that will store information about a specific request.
function Storage() { }

// This will be called every time asynchronous work is queued.
// The returned data will propagate to the callbackObject's callbacks.
function onCreation() {
  return new Storage();
}

// Set of methods that will run at specific points in time according to
// their cooresponding callbacks. These methods are always run FIFO if
// multiple are queued.
// "context" is the "this" of the request object.
var callbackObject = {
  before: function asyncBefore(context, storageValue) {
  },
  // "returnValue" is the value returned from the callback.
  after: function asyncAfter(context, storageValue) {
  },
  // If this callback returns "true" then the error handler will assume
  // the error was properly handled, and process will continue normally.
  // If multiple error handlers are queued, and any one of those returns
  // true, then Node will assume the error was properly handled.
  // This will not currently be passed the context (or "this") of the
  // callback that threw. A simple way of achieving this is currently
  // being investigated, and the feature will be added when one is found.
  error: function asyncError(storageValue, err) {
  }
};

/**
 * process.addAsyncListener(callback[, object[, value]]);
 *
 * Arguments:
 *
 * callback - Function that will be run when an asynchronous job is
 *    queued.
 *
 * object - Object with the optional callbacks set on:
 *    before - Callback called just before the asynchronous callback
 *        will be called.
 *    after - Callback called directly after the asynchronous callback.
 *    error - Callback called if there was an error.
 *
 * The returned key is an Object that serves as the unique key for the
 * call (much like how Timers work).
 */
var key = process.addAsyncListener(onAsync, callbackObject);


/**
 * process.createAsyncListener(callback[, object[, value]]);
 *
 * Adding an async listener will immediately add it to the queue and
 * being listening for events. If you wish to create the listener in
 * advance, to say attach to the returned value object, it's possible
 * to get the key and pass it to process.addAsyncListener() later.
 */
var key = process.createAsyncListener(onAsync, callbackObject, value);

// Then activate like so:
process.addAsyncListener(key);


/**
 * process.removeAsyncListener(key);
 *
 * Remove any async listeners and associated callbackObjects. All
 * listeners will live independent of each other, and there will be no
 * method given to clear all async listeners.
 */
process.removeAsyncListener(key);
```

### DIFF
--- a/doc/api/net.markdown
+++ b/doc/api/net.markdown
@@ -174,9 +174,9 @@ already been bound to a port or domain socket.
 Listening on a file descriptor is not supported on Windows.
 
 This function is asynchronous.  When the server has been bound,
-['listening'](#event_listening_) event will be emitted.
+['listening'][] event will be emitted.
 the last parameter `callback` will be added as an listener for the
-['listening'](#event_listening_) event.
+['listening'][] event.
 
 ### server.close([callback])
 
--- a/doc/api/process.markdown
+++ b/doc/api/process.markdown
@@ -647,4 +647,218 @@ a diff reading, useful for benchmarks and measuring intervals:
       // benchmark took 1000000527 nanoseconds
     }, 1000);
 
+
+## Async Listeners
+
+<!-- type=misc -->
+
+    Stability: 1 - Experimental
+
+The `AsyncListener` API is the JavaScript interface for the `AsyncWrap`
+class which allows developers to be notified about key events in the
+lifetime of an asynchronous event. Node performs a lot of asynchronous
+events internally, and significant use of this API will have a **dramatic
+performance impact** on your application.
+
+
+## process.createAsyncListener(asyncListener[, callbacksObj[, storageValue]])
+
+* `asyncListener` {Function} callback fired when an asynchronous event is
+instantiated.
+* `callbacksObj` {Object} optional callbacks that will fire at specific
+times in the lifetime of the asynchronous event.
+* `storageValue` {Value} a value that will be passed as the first argument
+when the `asyncListener` callback is run, and to all subsequent callback.
+
+Returns a constructed `AsyncListener` object.
+
+To begin capturing asynchronous events pass the object to
+[`process.addAsyncListener()`][]. The same `AsyncListener` instance can
+only be added once to the active queue, and subsequent attempts to add the
+instance will be ignored.
+
+To stop capturing pass the object to [`process.removeAsyncListener()`][].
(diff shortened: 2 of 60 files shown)


RobinQu:
Is it landed on any released version? 
I still cannot find it in the latest(v0.12.2)

trevnorris:
@RobinQu Never landed.

gobwas:
So, will node have this feature some day?

trevnorris:
@gobwas It partially exists as `process.binding('async_wrap')`. Though the ability to capture exceptions was removed. Became a massive hair ball. Would still like to do this, but time constraints.

gobwas:
Thank you, @trevnorris!

Does this feature present in iojs docs? 

Any way, I am asking about [async-listener](https://www.npmjs.com/package/async-listener) functionality at the moment..?

trevnorris:
Oh, that died a horrible death. After 3 months of work I still couldn't figure out a few key edge cases. So instead I've been working on the set of hooks necessary to implement a feature like async-listener. This way others can experiment and publish their own modules.

gergelyke:
Hi @trevnorris ,

could you give us an update on the current status of this? What would be your suggestion to use these days if we want to active the same functionality? 

Thanks!

cc @peteyy

trevnorris:
@gergelyke Unfortunately there isn't an alternative. Will have to continue using domains for the time being. Though discussion about reimplementing this has started back up. Hopefully it will lead to something.

gergelyke:
thanks!

naorye:
What is wrong with the Domain solution?

trevnorris:
@naorye It's a complete hack that injects itself all over core code. By design it can't properly catch everything. There are edge cases it's missing. This is because it's injected at the JS API level, not at the native level where the asynchronous requests are actually made.

Olegas:
What alternatives are now being discussed?

jiaqifeng:
@trevnorris Is there any other solutions based on native layer of this problem currently available? I am a Java coder, I think if this should be done on VM layer which make things easier and more efficient.

trevnorris:
@jiaqifeng In master there's a native API that can be used (see all functions that use `async_context` in [node.h](https://github.com/nodejs/node/blob/master/src/node.h)). This will be backported to v8.x, but no further.

Issue #5734, opened 2013-06-23
repl: don't auto-complete array indices


bnoordhuis:
```
$ node
> process.versions.node
'0.11.3-pre'
> var a = [1,2,3]
> a.<TAB>
a.__defineGetter__      a.__defineSetter__      a.__lookupGetter__      a.__lookupSetter__      a.__proto__
a.constructor           a.hasOwnProperty        a.isPrototypeOf         a.propertyIsEnumerable  a.toLocaleString
a.toString              a.valueOf               

a.concat                a.constructor           a.every                 a.filter                a.forEach
a.indexOf               a.join                  a.lastIndexOf           a.length                a.map
a.pop                   a.push                  a.reduce                a.reduceRight           a.reverse
a.shift                 a.slice                 a.some                  a.sort                  a.splice
a.toLocaleString        a.toString              a.unshift               

a.0                     a.1                     a.2                     a.length
> a.0
...
```

The current auto-completion is:
- not very useful because `a.0` is not a valid syntactic construct, and
- annoying when the array is large - the REPL lists all indices and you can't interrupt it

brendanashworth:
See: https://github.com/nodejs/node/pull/2409 which landed as https://github.com/nodejs/node/commit/ce145bf3581f3bcff71e9b069d2620783c2a6466. It is currently marked for a 0.10 and 0.12 backport, so I don't know if this issue is still relevant.

PR #6037, opened 2013-08-12
REPL tab complete for Arrays

seebees:
The tab complete for Arrays will
include any integer indexes.  This is
suboptimal because obj.21 is not
valid syntax and a large array will
contain a very cluttered list.

@bnoordhuis RE: #5734.  Ask and you shall receive.  2 months later, and perhaps not what you were looking for, but still, you've got to be impressed.

Note:  The test should pass, but when I run it with `assert.equal(false, true)` [here](https://github.com/seebees/node-1/compare/5734-REPL-tab-complete-array#L1R81) it still passes.  I don't think that has anything to do with my change though.

### DIFF
--- a/lib/repl.js
+++ b/lib/repl.js
@@ -580,7 +580,20 @@ REPLServer.prototype.complete = function(line, callback) {
           // if (e) console.log(e);
 
           if (obj != null) {
-            if (util.isObject(obj) || util.isFunction(obj)) {
+            if (util.isArray(obj)) {
+              // Array elements will contain many integer values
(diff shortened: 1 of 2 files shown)


jasnell:
Sigh.. 2013 and it looks like this was never reviewed. Just tested the patch against v0.12 and master and it LGTM... verified that this is still an issue in io.js too so we'll want to land there also. @joyent/node-tsc

jasnell:
Closing in favor of alternate: https://github.com/joyent/node/pull/25819

PR #6101, opened 2013-08-22
Proposal: move lazy initialization deeper in timers module

subzey:
There are wraps in `global.setTimeout`, `global.setInterval` etc. So if `setTimeout` is invoked, two excessive function calls are made. The reason, as far as I can understand, is lazy load.

There's a proposal: move lazy timers initialization into timers module itself.

But there's, uh, an side effect. Currently node startup is slightly faster, the main loop may be spontaneously (again, slightly) additional blocked in order to init timers. May I ask about the "strategy", what is the priority, to make node startup faster or to run smoother?


Fishrock123:
Fwiw io.js did this a while ago.

subzey:
That's just great. Thanks!

PR #6120, opened 2013-08-24
doc: reflect latest https feature

koichik:
Because global Agent now works well with TLS options, creating a custom Agent is no longer needed.


jasnell:
At this point this change isn't really necessary. A new PR can be opened against nodejs/node if someone wishes to pursue it.

PR #6314, opened 2013-10-07
Add the escapeShellArg function to child_process

mlocati:
When we call external programs via `exec()`, we should escape parameters that may contain special characters.

Let's implement a function that escapes shell arguments.

Supersedes https://github.com/joyent/node/pull/6308


jasnell:

PR #6332, opened 2013-10-10
build: install node and static library headers

tjfontaine:
fixes #5112


jasnell:
@misterdjules @orangemocha ... any reason to keep this one open? I can't imagine any.

jasnell:
Closing this here. If there is any reason to pursue this, a new PR would need to be opened in nodejs/master once the v0.10 branch moves over.

PR #6351, opened 2013-10-15
Clarified HTTP client callback handling

tazle:
Current HTTP client documentation doesn't specify what happens to the callback parameter of request(). Added documentation based on current code base.


jasnell:
Appears to already have landed in nodejs/node

PR #6367, opened 2013-10-17
crypto: Advanced test case for spkac functions

jas-:
Moved away from using file based PEM & SPKAC for tests to object based
comparisons of example SPKAC's using key sizes ranging from 1024 - 4096,
and hashing algorithms md4, md5, sha1, sha224, sha256, sha384, sha512 &
rmd160


jasnell:

PR #6368, opened 2013-10-17
Add support for toggling debug mode and accessor function for debug mode status flag.

abhinavraj:
Change list:
a) a C++ API function for checking whether the debugger is enabled
b) functionality to stop the debugger.


trevnorris:
@abhinavraj This will need a rebase. If you're still willing to do some work on it, ping back and I'll reopen.

PR #6380, opened 2013-10-20
src: enable fips support

arthurtsang:
FIPS 140-2 support for enterprise, especially government is very important.  OpenSSL supported FIPS with OpenSSL FIPS Object Module.  To enable NodeJS to support FIPS, all we need to do is to compile with OpenSSL shared library which built with OpenSSL FIPS Object Module and call FIPS_mode_set() to enable FIPS mode.

to test if NodeJS has turned on FIPS mode, we can use the script below.  since bf is an unsupported algorithm, node will core dumped.

```
#!/usr/bin/env node

'use strict';

var crypto = require('crypto');

var cipher = crypto.createCipher('bf','12345');
var encrypted = cipher.update('encrypt this very long string' , 'utf8', 'base64');
encrypted = encrypted + cipher.final('base64');

var decipher = crypto.createDecipher('bf','12345');
var decrypted = decipher.update(encrypted, 'base64', 'utf8');
decrypted = decrypted + decipher.final('utf8');

console.log( 'encrypted: ' + encrypted );
console.log( 'decrypted: ' + decrypted );
```

configure is modified to add --shared-openssl-fips to turn on FIPS mode.

more info on OpenSSl FIPS http://www.openssl.org/docs/fips/fipsnotes.html

### DIFF
--- a/configure
+++ b/configure
@@ -152,6 +152,11 @@ parser.add_option('--shared-openssl-libpath',
     dest='shared_openssl_libpath',
     help='a directory to search for the shared OpenSSL DLLs')
 
+parser.add_option('--shared-openssl-fips',
+    action='store_true',
+    dest='shared_openssl_fips',
+    help='enable FIPS mode for shared OpenSSL DLLs')
+
 parser.add_option('--shared-v8',
     action='store_true',
     dest='shared_v8',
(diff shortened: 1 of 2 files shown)


jasnell:
@joyent/node-tsc ... given that this targets master, would need to be updated, and is being revisited in #25463, I recommend closing this one. That's not to say we shouldn't land FIPS support in v0.12, but this particular PR likely wouldn't be the way we do it.

jasnell:
Closing this here. There is an ongoing discussion at https://github.com/nodejs/node/issues/2242. FIPS enablement would need to happen in nodejs/node master and cherry picked back to the v0.12 or v0.10 as appropriate.

PR #6421, opened 2013-10-27
stream: End writables before emitting 'end' event

### LINKED ISSUE #5603: Writing to ended write stream continues without error
Not sure if this is a bug but it seems odd.

Running the following testcase:
```javascript
var fs = require('fs');

var files = [
	__dirname + '/1.txt',
	__dirname + '/2.txt',
	__dirname + '/3.txt'
];

var outFile = __dirname + '/out.txt';

concat(files, outFile, function() {
	console.log(fs.readFileSync(outFile).toString());
});

function concat(files, outFile, cb) {

	var writer = fs.createWriteStream(outFile);
	writer.on('error', function(err) {
		console.log(err);
	});

	function iterate() {
		console.log('iterating');
		if (files.length === 0) {
			return cb();
		}

		var path = files.shift();
		var reader = fs.createReadStream(path);
		reader.on('error', function(err) {
			console.log(err);
		});
		reader.on('end', function() {
			console.log('end');
			iterate();
		});
		reader.pipe(writer);
	}

	iterate();

}
```

produced an output of:
```
iterating
end
iterating
```

I eventually figured out why things weren't working as I expected (`pipe()` ends by default) and fixed it but I thought it was odd that the script continued while trying to `pipe()` for the second time with no error.

Should the WriteStream have emitted `error` in this case?

Node.js version 0.10.8

joybro:
There is a timing issue b/w the internal callback for ending writable
and the user callback of 'end' event because both of them are using
the same 'end' event.

So this fails:

```
reader.pipe(writer);
reader.on('end', function() {
  writer.end('Goodbye\n');  // should be failed
});
```

But this succeeds:

```
reader.on('end', function() {
  writer.end('Goodbye\n');  // should be failed
});
reader.pipe(writer);
```

This commit ends writables not using 'end' event.

Fixes #5603

### DIFF
--- a/lib/_stream_readable.js
+++ b/lib/_stream_readable.js
@@ -458,6 +458,15 @@ Readable.prototype._read = function(n) {
   this.emit('error', new Error('not implemented'));
 };
 
+function registerEndFn(src, endFn) {
+  if (!src._endFns)
+    src._endFns = endFn;
+  else if (util.isObject(src._endFns))
+    src._endFns.push(endFn);
+  else
+    src._endFns = [src._endFns, endFn];
+}
+
 Readable.prototype.pipe = function(dest, pipeOpts) {
   var src = this;
   var state = this._readableState;
@@ -484,7 +493,7 @@ Readable.prototype.pipe = function(dest, pipeOpts) {
   if (state.endEmitted)
     process.nextTick(endFn);
   else
-    src.once('end', endFn);
+    registerEndFn(src, endFn);
 
   dest.on('unpipe', onunpipe);
   function onunpipe(readable) {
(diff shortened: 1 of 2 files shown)


jasnell:
Dang, appears this was never reviewed. @trevnorris ... any thoughts on this one?

jasnell:
@chrisdickinson

jasnell:
Unfortunately the timing of the events here is part of the implicit API and it's not at all clear whether this would break existing apps. Given the potential API change, this would need to be updated and reopened against master on http://github.com/nodejs/node. Closing this PR here but can reopen if necessary.

PR #6434, opened 2013-10-29
#3977 Node.js crash on Windows when no read permissions for the root

napernik:
Fixing an error that was occurring if node was hosted on Azure or shared hosting, and the website user didn't have read NTFS persmissions on the root folder.

Examples:
https://github.com/joyent/node/issues/3977
http://stackoverflow.com/questions/12718488/error-eperm-operation-not-permitted-with-node-js-and-etherpad-lite


jasnell:

PR #6442, opened 2013-10-31
Add support for concatenated gzip files.

eendeego:


misterdjules:
@luismreis @chrisdickinson Should we re-open this PR so that we can fix the regression it caused (see https://github.com/joyent/node/issues/8962) and land it again in a later milestone?

In the meantime, I'm removing it from the 0.11.15 milestone.

luismreis:
@misterdjules I don't really know what to say... According to the (gzip) spec we should be restarting the gunzip process each time there's more data, but, apparently, npm depends on the behavior of terminating on the first terminating gzip stream.

PR #6455, opened 2013-11-02
added --enable-shared support for building libnode.so

unbit:
This patch adds a new 'configure' flags, named --enable-shared that build a libnode.so library and a 'node' executable using it.

The main purpose is allowing easy embedding of node in external apps:

```
#include <node.h>

void callback(uv_timer_s* t, int status) {
        printf("ciao %d\n", status);
}

void hello(uv_idle_t* handle, int status) {
        printf("hello %d\n", status);
        uv_idle_stop(handle);
}

int main() {
        int argc = 3;
        char *argv[4];
        argv[0] = (char *) "pipponode";
        argv[1] = (char *) "-e";
        argv[2] = (char *) "0";
        argv[3] = NULL;

        uv_timer_t timer_req;
        uv_idle_t check;

        uv_timer_init(uv_default_loop(), &timer_req);
        uv_timer_start(&timer_req, callback, 5000, 2000);

        uv_idle_init(uv_default_loop(), &check);
        uv_idle_start(&check, hello);

        node::Start(argc, argv);
}

```

the patch is already in use for my company attempt to embed node.js in uWSGI

Some note:

when building libso the main/wmail symbols are rewritten as node_main/node_wmain and exposed as C calls (this simplify management).

-rpath on node_prefix/lib is added to easily allow the node binary to find libnode.so

### DIFF
--- a/configure
+++ b/configure
@@ -44,6 +44,11 @@ parser.add_option('--gdb',
     dest='gdb',
     help='add gdb support')
 
+parser.add_option('--enable-shared',
+    action='store_true',
+    dest='enable_shared',
+    help='build a node shared library')
+
 parser.add_option('--ninja',
     action='store_true',
     dest='use_ninja',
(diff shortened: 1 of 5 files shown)


jasnell:

PR #6462, opened 2013-11-04
Include optional HOME path for group policies

bsawyer:
subsystem: Enhancement for providing an optional environment variable declaration; Useful as a workaround for network group policies.


jasnell:
@orangemocha ... what do you think?

jasnell:
Closing this here. If there's reason to pursue, a new PR against nodejs/node master should be opened

Issue #6400, opened 2013-10-23
assert: doesNotThrow doesn't show message


pkorzeniewski:
In the documentation, `assert.doesNotThrow` takes two arguments: `(block, [message])`, but calling

```javascript
assert.doesNotThrow(function() {
    // some code throwing an error
}, "Error occured");
```

won't display the "Error occured" message, but the thrown error instead. I've inspected the `assert.js` file, and the problem is probably caused by line 304:

```javascript
if (!shouldThrow && expectedException(actual, expected)) {
```

`expectedException` will always return `false`, because `expected` is always set to `null`

```javascript
 if (util.isString(expected)) {
    message = expected;
    expected = null;
  }
```

jasnell:
There is a new PR opened in nodejs/node tracking this issue. Closing here.

PR #6470, opened 2013-11-06
assert: respect assert.doesNotThrow message.

diversario:
Addresses #6400.
Special handling to detect when user has supplied a custom message.
Added a test for user message.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -299,7 +299,11 @@ function _throws(shouldThrow, block, expected, message) {
     fail(actual, expected, 'Missing expected exception' + message);
   }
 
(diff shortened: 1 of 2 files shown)


jasnell:
@diversario ... unfortunately it doesn't look like this was ever reviewed further. Given that this involves a change in functionality, it would likely be best to close this PR here and open a new one against the active dev channel at http://github.com/nodejs/node if this is something you'd still like to pursue.

jasnell:
New issue in nodejs/node created to track this. The change would need to be made there and potentially cherry picked to v0.12 or v0.10. Closing this here.

PR #6502, opened 2013-11-12
events: add event observers

trevnorris:
This patch adds an API that is similar to AsyncListeners. It will allow users to add "observers" that will be called at key times in the EventEmitter's life cycle.


trevnorris:
@tjfontaine @cjihrig @misterdjules Any of you think this could still land, or can I close this?

trevnorris:
Welp, not going to happen. Was a good idea though.

PR #6504, opened 2013-11-13
build: Simplify the logic in vcbuild.bat

JayBeavers:
While adding support for Visual Studio 2013, I noticed duplicate
logic for VS detection in gyp (MSVS*.py), nodejs (vcbuild.bat),
and node-gyp (lib/configure.js). The detection logic in gyp is
sufficient for our needs, so I am simplifing the logic in nodejs
and node-gyp.

This change requires node to be built from a VS Command Prompt
as it checks for the existence of VS1x0COMNTOOLS and relies on
msbuild being on the path.

This change also updates the README to specify using a command
prompt, note the support of VS 2013, and add hyperlinks to the
downloads.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,11 @@ non-standard name, run the following instead:
 
 Prerequisites (Windows only):
 
-    * Python 2.6 or 2.7
-    * Visual Studio 2010 or 2012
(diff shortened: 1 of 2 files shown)


jasnell:
Closing due to lack of activity. Can revisit if someone is interested in updating the PR or if anyone feels this is still relevant

PR #6525, opened 2013-11-15
fix(zlib) add check is callback is function

coderaiser:
Add check is callback is function to prevent next error.

``` js
zlib.js:160
    callback(err);
    ^
TypeError: undefined is not a function
    at Unzip.onError (zlib.js:160:5)
    at Unzip.EventEmitter.emit (events.js:95:17)
    at Zlib._binding.onerror (zlib.js:298:10)
```

I have one suggestion. Could be used construction like this:

``` js
function exec (callback, params) {
    if (util.isFunction(callback))
       callback(params);
}
```

Like it's done [here](https://github.com/coderaiser/util.io/blob/9370aad67e5bd79a4588bdf209c326a1687d071a/lib/util.js#L851). So there is would be less code.

### DIFF
--- a/lib/zlib.js
+++ b/lib/zlib.js
@@ -185,13 +185,18 @@ function zlibBuffer(engine, buffer, callback) {
   function onError(err) {
     engine.removeListener('end', onEnd);
     engine.removeListener('readable', flow);
-    callback(err);
+    if (util.isFunction(callback)) {
+      callback(err);
+    }
   }
 
   function onEnd() {
     var buf = Buffer.concat(buffers, nread);
     buffers = [];
-    callback(null, buf);
+    if (util.isFunction(callback)) {
+      callback(null, buf);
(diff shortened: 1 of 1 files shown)


jasnell:
Well.. never seems to have landed :-( @indutny ... any other thoughts on this?

@coderaiser ... if you're still interested in getting this landed, unless @joyent/node-tsc feel that this is something we should land in v0.12, it may be best to close this here and open it against master in http://github.com/nodejs/node (which is now the active development stream)

indutny:
@jasnell after a year I could only say that it needs a test. Other than that - everything is still looking good (except those braces)

jasnell:
Closing this here as its not going to be able to land in master. If this is going to land, it needs to be a new PR against http://github.com/nodejs/node. I've opened an issue to track.

PR #6580, opened 2013-11-24
docs(assert) call signature of doesNotThrow should be equivalent to thro...

basarat:
...ws.

The body is identical:

```
assert.throws = function(block, /*optional*/error, /*optional*/message) {
  _throws.apply(this, [true].concat(pSlice.call(arguments)));
};

// EXTENSION! This is annoying to write outside this module.
assert.doesNotThrow = function(block, /*optional*/error, /*optional*/message) {
  _throws.apply(this, [false].concat(pSlice.call(arguments)));
};
```

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -319,7 +319,7 @@ assert.throws = function(block, /*optional*/error, /*optional*/message) {
 };
 
 // EXTENSION! This is annoying to write outside this module.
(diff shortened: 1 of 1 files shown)


jasnell:
This PR is not going to be able to land. I opened an issue on nodejs/node to track the issue. Closing here.

PR #6670, opened 2013-12-10
Update http.markdown

liveinjs:
response.finished

### DIFF
--- a/doc/api/http.markdown
+++ b/doc/api/http.markdown
@@ -330,6 +330,12 @@ Example:
 After response header was sent to the client, this property indicates the
 status code which was sent out.
 
+### response.finished
+
+Default `response.finished = false` , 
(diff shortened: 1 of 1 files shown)


jasnell:
Hmm... unfortunately it doesn't look like this one was ever reviewed. @liveinjs , thank you for submitting this, sorry it's taken so long for someone to look at it. There are a couple of grammatical issues that should be addressed before it's landed tho; and we should verify that the `finished` property is meant to be part of the API. @cjihrig @trevnorris ... does this look fine to you?

cjihrig:
I'm not 100% sure. Given that it's not underscored, it seems like it should be documented. If we want to document it, this needs a bit of cleanup.

trevnorris:
Agree w/ @cjihrig

jasnell:
New cleaned up PR opened. Closing this here.

PR #6712, opened 2013-12-15
doc: fix error in removeListener event description

bsnote:


jasnell:
@sam-github ... I'm inclined to close this without further action. Do you agree or think there's something here we need to capture?

brendanashworth:
+1 for close

sam-github:
@jasnell I agree. Wording is improved in 3.x, and the behaviour is no longer unspecified.

PR #6725, opened 2013-12-17
doc: clarify process.nextTick vs. setImmediate (was: Fixes #6718)

### LINKED ISSUE #6718: nextTick documentation and warning message are confusing
A user of node-vasync reported this warning message emitted by Node on the latest v0.10:

```
Trace: (node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
    at maxTickWarn (node.js:377:17)
    at process.nextTick (node.js:480:9)
    at onwrite (_stream_writable.js:260:15)
    at WritableState.onwrite (_stream_writable.js:97:5)
    at WriteStream.Socket._write (net.js:651:5)
    at doWrite (_stream_writable.js:221:10)
    at writeOrBuffer (_stream_writable.js:211:5)
    at WriteStream.Writable.write (_stream_writable.js:180:11)
    at WriteStream.Socket.write (net.js:613:40)
    at Console.log (console.js:53:16)
    at WorkQueue.worker (/Users/linda/jpc2/jpc-report-db/scripts/zuora_verify_usage_load:102:11)
    at /Users/linda/jpc2/jpc-report-db/node_modules/vasync/lib/vasync.js:317:6
    at process._tickCallback (node.js:415:13)
```

This is a classic misuse of process.nextTick: the caller is using a vasync Queue with a synchronous worker function, and the Queue dispatches each new task using process.nextTick.  As a result, once the first task completes, all subsequent ones are dispatched and completed on the same tick.  If that number is more than 1000, we get this warning (though the program is arguably operating correctly, as this is not an infinite loop).

While this is sort of a programmer error, I think there are two concrete Node changes that would improve the situation.

First is the warning message.  There's no recursion going on here.  It's just a loop.  The code refers to "depth", but it's not stack depth -- it's just a loop counter (as the code comment says).  I get the sense in which it feels recursive, in that the function is triggering a call back into itself, but it's confusing to say that recursion is the problem because the _real_ problem (which is only _potentially_ detected) is event loop starvation, and that's not normally a result of recursion.  I think this would be a clearer error message:

```
warning: infinite loop suspected after 1000 nextTick handlers were processed in the same tick.  Remaining handlers have been deferred, but future versions of Node will continue executing handlers, which may result in an infinite loop.  Please use setImmediate for deferral instead of nextTick.
```

That makes it much clearer what actually happened, what Node did about it, why that's still a problem for the developer, and what to do about it.

The second improvement is for the documentation for process.nextTick, which I think leads people down the wrong path.  There appear to be two common use cases for nextTick():
1. It's used _within_ a subsystem to defer execution of a known function until the current "tick" completes, typically so that the caller can finish updating the state of an object and process the rest of the event later.
2. It's used to defer execution of an _arbitrary_ callback (i.e. to some other module).

I take as given that case 1 is useful, but case 2 is toxic because in case 2 there's no way to be sure to avoid the recursion problem.  Users must always use setImmediate() for case 2.  Since I think case 2 is overwhelmingly more common than case 1, I think the docs should strongly downplay any use of process.nextTick in favor of setImmediate.  Specifically:
- The first sentence says: "On the next loop around the event loop call this callback."  That sounds too much like what people want when they really want setImmediate.  nextTick should be qualified, with something like: "Call this callback the next time nextTick handlers are processed.  This is normally at the start of the next loop around the event loop, but it may be much sooner if nextTick handlers are currently being processed.  When nextTick handlers are processed, they are processed until none remain.  As a result, calling nextTick from a nextTick handler extends the loop.  Repeatedly doing so can thus starve the event loop, preventing any other events (e.g., network or filesystem I/O) from being processed until all handlers are processed."
- The example of a "hazardous" API should not be using nextTick at all.  If the API function is being called in a nextTick() handler, the program may run into the recursion problem.
- The process.nextTick docs should explicitly refer users to setImmediate().

I'd be happy to draft these changes, if this sounds good.

davepacheco:
This documentation change clarifies the potential issue with process.nextTick and steers users towards setImmediate instead for arbitrary callbacks.

### DIFF
--- a/doc/api/process.markdown
+++ b/doc/api/process.markdown
@@ -564,12 +564,23 @@ This will generate:
 
 * `callback` {Function}
 
-Once the current event loop turn runs to completion, call the callback
-function.
+Call this callback the next time nextTick handlers are processed.  This is
+normally once the current event loop turn runs to completion, but it may be much
+sooner if nextTick handlers are currently being processed (i.e., if the caller
+is part of a nextTick handler).
 
-This is *not* a simple alias to `setTimeout(fn, 0)`, it's much more
-efficient.  It runs before any additional I/O events (including
-timers) fire in subsequent ticks of the event loop.
+At the end of each tick the `process.nextTick` handlers are executed until there
+are none left.  Because of this, it's dangerous to invoke `process.nextTick`
+*from* a nextTick handler.  If you do this enough times, you can starve the main
+event loop, meaning that filesystem and network I/O events will not be processed
+(because nextTick handlers keep being added).
+
+**You should not use `process.nextTick` to invoke an arbitrary callback that
+you've been given by a caller.**  For that, use `setImmediate`, which avoids
+this starvation problem.  `process.nextTick` should only be used when the
+callback *must* be invoked before other I/O and timer events are processed, and
+you know that calling that callback won't itself start a long chain of
+`process.nextTick` calls.  If you're not sure, use `setImmediate`.
 
     console.log('start');
     process.nextTick(function() {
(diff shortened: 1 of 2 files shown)


jasnell:

davepacheco:
This is disappointing.  People are still confused by this.  What work needs to be done to make this able to be integrated?

jasnell:
Basically just someone willing to do the work. The last update on it was Dec *2013*.

davepacheco:
When I last updated this in 2013, I thought it _was_ done.  What's "the work"?  Is it that it doesn't merge cleanly now?

jasnell:
Generally, yes. (1) make sure the changes are still relevant (i.e. make sure that there haven't been other changes landed that make this irrelevant) and (2) make sure they still apply cleanly. Beyond that, I'd recommend also making sure that a similar change doesn't need to be landed on the io.js side of things so that when things merge there's less additional work to do.

jasnell:
If you're willing to make sure it's updated, I'll reopen.

davepacheco:
Yes, the changes are still relevant.  The issue that this fixes is still open (#6718), and even has recent comments.  [People are clearly still confused by the existing documentation.](https://twitter.com/rauchg/status/624696153750122496)  And as far as I can tell, the documentation is in exactly the same state as when I submitted the PR: the text that I removed in the PR is still there, the text that I moved is still present in its old place, and nothing has been added to clarify things.  (I'm not sure why it doesn't merge cleanly.)

Isn't it inappropriate to _close_ this PR without having answered those questions first?

jasnell:
@davepacheco as I said, I'm willing to reopen. The point to closing this is exactly to see if anyone is willing to step up to make sure the works gets completed.

jasnell:
New issue opened in nodejs/docs. Closing this here but still need to revisit.

PR #6726, opened 2013-12-17
Use the version of keepAlive which can be set externally

zackbloom:


jasnell:

PR #6744, opened 2013-12-20
build: Adds node_lib target.

zcbenz:
Adds a node_lib target that builds everything except the node_main.cc as a static library. And the node target would be depended on node_lib and compiles out the executable as usual.

With this other projects would be possible to include node as a third party library and embed a node engine in their applications.


jasnell:

PR #6774, opened 2013-12-28
windows: case-insensitive lookup in module cache

orangemocha:
The module cache key is the resolved filename which should
be treated as case-insensitive on Windows.

Fixes test-module-loading.js on Windows.

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -271,8 +271,11 @@ Module._load = function(request, parent, isMain) {
   }
 
   var filename = Module._resolveFilename(request, parent);
+  var cacheKey = process.platform === 'win32' ?
+      filename.toLowerCase() :
+      filename;
 
-  var cachedModule = Module._cache[filename];
+  var cachedModule = Module._cache[cacheKey];
   if (cachedModule) {
     return cachedModule.exports;
   }
@@ -297,7 +300,7 @@ Module._load = function(request, parent, isMain) {
     module.id = '.';
   }
 
-  Module._cache[filename] = module;
+  Module._cache[cacheKey] = module;
 
   var hadException = true;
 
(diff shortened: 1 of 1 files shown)


jasnell:
@orangemocha ... is this still needed/relevant?

orangemocha:
@jasnell : still relevant, but since it's been a while I need to take a second look and make sure it's up to date and the best fix for the underlying issue.

jasnell:
@orangemocha ... ok, it looks like @misterdjules submitted an alternate fix to the same problem... https://github.com/joyent/node/pull/8145

orangemocha:
@jasnell : thanks for the link. I still think we need case normalization, which path.normalize doesn't do. While trying to load the same module with different casing would be most likely a user error, the module will still load, but twice.

jasnell:
+1 .. definitely agree that the case normalization is important.

FransGH:
Old topic but still a pain if you waste valuable hours of your life troubleshooting non-working singletons..

This is not just a Windows issue, Please have a look at this pull request #8145  that nails it.

Note that there are two issues with require() case sensitivity, #8145 addresses the 2nd:
1) on case sensitive files systems, some modules might not load (throwing an error)
2) on case insensitive file systems, calling require() for the same module but with different case loads the module multiple times (breaking singletons)

Jeff-Lewis:
Please forgive me... I can't believe this actually is how node works on Windows. I just spent several hours debugging to find these issues have been written up for a couple years. 

I look at my require.cache and I have hundreds of `C:\modules...` and hundreds of same `c:\modules...`. loaded twice.

Please just get someone at MSFT to make a PR....

(edited)
Continuting issue here [Repro of Node loading modules twice on Windows](https://github.com/nodejs/node/issues/6978)

PR #6802, opened 2014-01-04
net: WIP run close callbacks in correct eloop phase

trevnorris:
Instead of running the close callbacks seemingly synchronously instead
of when the handle has actually been closed by libuv, instead run the
callbacks in the uv__run_closing_handles() phase of the eloop.

There's an issue with the cluster module overriding the `close()` callback that prevents the callbacks from being executed at the correct time.

**NOTE**: This is missing tests and not fully functional. Don't merge yet.

### DIFF
--- a/lib/child_process.js
+++ b/lib/child_process.js
@@ -355,6 +355,8 @@ function setupChannel(target, channel) {
       target.disconnect();
       channel.onread = nop;
       channel.close();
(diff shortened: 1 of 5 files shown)


trevnorris:
Still think it should be done this way, but not taking the time to update the code.

PR #6855, opened 2014-01-13
Fix path.relative on windows network shares

ForbesLindesay:
Array.prototype.slice takes `(startIndex, endIndex)` rather than `(startIndex, length)`.  If your path starts with `//`  or `\\` there will be two empty spaces at the start, which would lead to the last two parts of the path being chopped off.

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -269,7 +269,7 @@ if (isWindows) {
       }
 
       if (start > end) return [];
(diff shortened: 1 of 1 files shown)


jasnell:
Closing due to lack of activity. Can revisit if someone is interested in updating the PR or feels it is still relevant.

PR #6890, opened 2014-01-17
cluster: pass settings.execPath to fork

chulkilee:
With 70ad9bb - it would be nice to be able to set execPath with `cluster.fork`.

It's safe to have null for execPath since `child_process.fork` will handle it.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -285,6 +285,7 @@ function masterInit() {
     worker.process = fork(settings.exec, settings.args, {
       env: workerEnv,
       silent: settings.silent,
+      execPath: settings.execPath,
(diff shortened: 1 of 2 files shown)


jasnell:
Hmm... @chulkilee, given that this PR introduces a change in behavior, targets master (which is no longer the active development branch) and has conflicts that need to be resolved, if this is still something that you'd like to pursue, it would be best if this was closed here and a new, updated PR opened against http://github.com/nodejs/node.

jasnell:
Closing this here. @chulkilee if this is a change you'd like to pursue, please update and open a new PR against master on http://github.com/nodejs/node

PR #6948, opened 2014-01-23
doc: readline document TTY utils

indutny:
fix #6933


eugenejscray-zz:
Dumb Question but what is meant by the position x and y? Lets say I had a file would the y position be my line? and would the x position be the character index within the line?

PR #6978, opened 2014-01-26
Use unminified shjs language definition

kapouer:
It's a better open-source practice to distribute the preferred form of the work for making
modifications to it.
Typically, if that file needs a patch, it is easier to do it on the unminified version,
and it automatically fits GPL-3 requirements.

Copy taken from
http://shjs.sourceforge.net/lang/sh_javascript.js


jasnell:
Closing due to lack of activity. Can reopen if folks feel it's necessary.

PR #6980, opened 2014-01-28
Update fs.readdir with JSDoc comment

gandazgul:
If I do this kind of comments on all public methods would you merge them in?
You can generate documentation out of them plus they provide nice code hinting and completion on a lot of editors.


jasnell:
This is not likely to happen in v0.12 or v0.10. Recommend that if someone wants to move this further, they open a PR against http://github.com/nodejs/node or http://github.com/nodejs/io.js

PR #6926, opened 2014-01-29
Out of memory when inspect large sparse array

litmit:
util.inspect crashes Node when I  inspect large sparse array (even empty!)

Try in Node's REPL:

```
> new Array(0xffffffff)
```

Output:

```
FATAL ERROR: JS Allocation failed - process out of memory
```

But this work:

```
var a = new Array(0xFFFFFFFF)
a[0xFFFFFFFF]=1
console.log(a[0xFFFFFFFF],a.length);
```

It is possible to detect large arrays in util.inspect and not output them at whole or replace them with some stub like this:

```
[ 1, 20, 6,
... 2000000 elements skipped ...
34, 68, 30]
```

?

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -19,6 +19,7 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+var MAX_SPARSE_ARRAY_GAP = 10;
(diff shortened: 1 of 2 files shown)


jasnell:
@trevnorris ... what do you think about this one? Obviously it was never rebased. I'm fine with closing it. A new PR against nodejs/node master would likely be the best option.

jasnell:
@litmit ... at this point, this would need to land against http://github.com/nodejs/node master. If you want to pursue, please update and open a new PR.

Issue #6943, opened 2014-01-22
util.inspect() on process.config() and os.networkInterfaces() contains escaped line breaks "\n" 


HansHammel:
1. problem:  log entries contain "\n" instead of line breaks 
2. problem: while on windows, line break should be "\r\n"

reproduced like so:

```
Logger.info('PROCESS: ' + util.inspect(process.config, {
  showHidden: false,
  depth: null,
  colors: false,
  customInspect: false
}));
```

outputs in log file:

```
{"level":"info","message":"PROCESS: { target_defaults: \n   { cflags: [],\n     default_configuration:      [...]
```

should be:

```
{"level":"info","message":"NETWORK: { target_defaults: 
   { cflags: [],
     default_configuration:      [...]
```

Logger is a Winston File Logger
while e.g. `util.inspect(process.memoryUsage()));` and others are fine, the output of `util.inspect(os.networkInterfaces())` and `util.inspect(process.config())` are broken (^)

Maybe a Winston problem (Winston uses unix line breaks only) !?

jasnell:
Marking this as a feature request as it does not actually appear to be a bug. Rather, the discussion aims towards making the line length configurable somehow.

PR #6991, opened 2014-01-29
lib: Add "wrap" option to util.inspect to disable or set line length

thomseddon:
Previously inspect would always wrap if the string exceeded 60 chars.

Now the "wrap" option can either be set to false, to disable wrapping, to an explicit number to enable wrapping at that length, or true to revert to the previous default wrapping length of 60 (default behaviour).

Follows on from #6943


jasnell:
@trevnorris ... any further thoughts on this? I'm inclined to think this should be closed here and reopened against http://github.com/nodejs/node.

jasnell:
@thomseddon ... at this point, because this is an API change, if you'd like to pursue this, please open a new PR against master on http://github.com/nodejs/node.

PR #7094, opened 2014-02-10
module: fix addition of filetypes, if require ends with ..

robertkowalski:
Given my home-directory is `/Users/rocko` - and I have a file named
`npm.json` in it and also a repository with name `npm`, which is a
node-module.

When try to require the `/Users/rocko/npm/index.js` two direcotry
levels down in the npm folder (e.g. `/Users/rocko/npm/test/tap`)
with `require("../../")` node will load `/Users/rocko/npm/index.js`.

When I use `require("../..")` node will load `/Users/rocko/npm.json`
which is fixed by this commit.

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -167,20 +167,21 @@ Module._findPath = function(request, paths) {
   // For each path
   for (var i = 0, PL = paths.length; i < PL; i++) {
     var basePath = path.resolve(paths[i], request);
+    var stats = statPath(basePath);
     var filename;
 
     if (!trailingSlash) {
       // try to join the request to the path
       filename = tryFile(basePath);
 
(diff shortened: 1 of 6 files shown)


jasnell:
@robertkowalski ... still need/want this?

robertkowalski:
it landed in io.js along with other improvements to the module loader

jasnell:
Do you see a reason to land it in v0.12? Or can we close this and pick it up under the converged stream?

robertkowalski:
it is a bugfix, but a breaking fix. the idea was to land it in 0.11 and get it into the 0.12 release. i am happy if it lands in the next versions, so we can close it. :)

PR #6328, opened 2014-02-14
Add path.common(path1/*, ...pathn*/);

medikoo:
``` javascript
var path = require('path');

path.common('/lorem/ipsum/foo/bar', '/lorem/ipsum/raz/dwa', '/lorem/elo/foo/bar'); //  => '/lorem'

path.common('/foo/bar', '/raz/dwa'); // => '/'

path.common('C:\\\\foo\\bar', 'D:\\\\foo\\bar'); // => null
```

I have use case for it in few of my modules (using counterparts for now), and I wonder why it's not part of `path` API.

I'll be happy to provide a pull request, but before that I'd like to be sure that you're open for such add on.

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -84,6 +84,16 @@ if (isWindows) {
     return '\\\\' + device.replace(/^[\\\/]+/, '').replace(/[\\\/]+/g, '\\');
   };
 
+  var normalizeForCommon = function (path) {
+    if (!util.isString(path)) {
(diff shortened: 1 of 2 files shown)


jasnell:

medikoo:
@jasnell at some point I'll probably propose it for http://github.com/nodejs/node at this point as there are two projects, it hardly make sense

Issue #6415, opened 2013-10-25
readline with multi-line question (\n) repeats whole question on backspace/delete


Rob--W:
When a question passed to `readline.question` contains a `\n` character, pressing Enter/Backspace causes the previous questions to be repeated.

Example:

```javascript
var rl = require('readline').createInterface({
    input: process.stdin,
    output: process.stdout
}); 
rl.question('First\nSecond\nThird ', function(answer) {
    rl.close();
    console.log('Result = ' + answer);
});
```

Output:
```
First
Second
Third |
```

After pressing any key, say, "xX"

```
First
Second
Third xX|
```

Now, press backspace to observe the bug:

```
First
Second
First
Second
Third x|
```

As you can see, the last line disappeared, and the whole question is repeated.

Node v0.10.21 on Linux.

jasnell:
Unable to reproduce. Closing. Can reopen if necessary

PR #7128, opened 2014-02-15
lib/readline.js: Fix backspace for multi-line(\n)

Ben-Waters:
Add additional parameter to cursorTo in _refreshLine to ensure that
the cursor moves to the top left row as well as the far left column
This prevents multi-line output from repeating previous lines.

Calculate the cursor position relative to the last line by splitting
the prompt text on newline characters.

Fixes joyent/node#6415

### DIFF
--- a/lib/readline.js
+++ b/lib/readline.js
@@ -228,7 +228,7 @@ Interface.prototype._refreshLine = function() {
   }
 
   // Cursor to left edge.
-  exports.cursorTo(this.output, 0);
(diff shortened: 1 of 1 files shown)


jasnell:
@balupton ... just wanted to check on this. @indutny had asked for a test case to be added before landing. It appears to have stalled out at that point. If this is something that you'd still like to pursue, it would likely be best to (a) add the test case then (b) close this here and open a new PR against master on http://github.com/nodejs/node (the current active development branch).

PR #7132, opened 2014-02-16
http: Improve HTTP Expect header handling

### LINKED ISSUE #4651: Improve HTTP Expect header handling
Handling of the HTTP `Expect` header was implemented in #316.  However, the http module simply [checks](https://github.com/joyent/node/blob/v0.8.18/lib/http.js#L1878-1890) to see if the header's value is `100-continue`.

If it is, it applies logic to either raise the `checkContinue` event or simply send a `HTTP/1.1 100 Continue`.

However, if it is anything else, http simply ignores the header.  This is a violation of [the spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.20), which [allows](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.20) for values other than `100-continue`.

> A server that does not understand or is unable to comply with any of the expectation values in the Expect field of a request MUST respond with appropriate error status. The server MUST respond with a 417 (Expectation Failed) status if any of the expectations cannot be met or, if there are other problems with the request, some other 4xx status.
> 
> This header field is defined with extensible syntax to allow for future extensions. If a server receives a request containing an Expect field that includes an expectation-extension that it does not support, it MUST respond with a 417 (Expectation Failed) status.

Application authors should not have to manually check the `Expect` header just to properly handle a protocol detail.  http should be able to send a 417 automatically.

Of course, the flip side of this is that application authors should be able to handle custom `Expect` values.

I propose:
1. By default, respond with a 417 error for unknown `Expect` values.
2. Extend the current handling of the `Expect` header to allow applications to hook in and prevent the default response.

For example, we could add a `checkExpectation` event that works somewhat like `checkContinue`: if there are no listeners, respond with a 417.  If there are listeners, just pass the request and response objects to the listeners.

Alternatively, we could add an option to `http.Server` that controls whether or not unknown expectations are automatically handled.  If disabled, we'd leave it up to the app's normal request handling code, which would have to explicitly check the `Expect` header's value.

faridnsh:
Just following the spec:
http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.20

Fixes #4651

Note that, Now we have the following pattern 3 times in the _http_server.js:

```
if (EventEmitter.listenerCount(self, 'someEventName') > 0) {
    self.emit('someEventName');
} else {
     // Do something as default
}
```

I think maybe we should move it to a function, though didn't have any good idea for the name of the function. Something like this. If you guys give me a good name, I'll refactor it.

I also see this pattern in userland too. Maybe it should be a event emitter prototype method name `setDefaultListener` that will set the default action when there's no listeners but won't do anything when there are listeners to a particular event.

### DIFF
--- a/doc/api/http.markdown
+++ b/doc/api/http.markdown
@@ -115,6 +115,17 @@ request body.
 Note that when this event is emitted and handled, the `request` event will
 not be emitted.
 
+### Event: 'checkExpectation'
+
+`function (request, response) { }`
+
+Emitted each time a request with an http Expect header is received, where the
+value is not 100-continue. If this event isn't listened for, the server will
+automatically respond with a 417 Expectation Failed as appropriate.
+
+Note that when this event is emitted and handled, the `request` event will
+not be emitted.
+
 ### Event: 'connect'
 
 `function (request, socket, head) { }`
--- a/lib/_http_server.js
+++ b/lib/_http_server.js
@@ -492,14 +492,22 @@ function connectionListener(socket) {
     }
 
     if (!util.isUndefined(req.headers.expect) &&
-        (req.httpVersionMajor == 1 && req.httpVersionMinor == 1) &&
-        continueExpression.test(req.headers['expect'])) {
-      res._expect_continue = true;
-      if (EventEmitter.listenerCount(self, 'checkContinue') > 0) {
-        self.emit('checkContinue', req, res);
+        (req.httpVersionMajor == 1 && req.httpVersionMinor == 1)) {
+      if (continueExpression.test(req.headers['expect'])) {
+        res._expect_continue = true;
+        if (EventEmitter.listenerCount(self, 'checkContinue') > 0) {
+          self.emit('checkContinue', req, res);
+        } else {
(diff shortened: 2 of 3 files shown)


jasnell:
@alFReD-NSH ... still want this?

alFReD-NSH:
I actually don't need this for myself anymore. But should check with #4651

jasnell:
Ok, given the change I'm not sure if this would land in here or if it should be retargeted to nodejs/node. @alFReD-NSH , are you willing to take another look at this?

alFReD-NSH:
I don't need this anymore. If you think this should be merged, then merge it if not, if it needs any changes, then close this, because at this moment I don't have time for this.

jasnell:
For now I'll mark it as a low priority and assign myself to it. If I'm able to get back to it I will, otherwise it'll just have to sit a while longer. Thanks @alFReD-NSH !

jasnell:
Opened a new issue in nodejs/node to track this. Will close this PR as it's not going to be able to land.

PR #7137, opened 2014-02-17
domain: remove multiple of the same from stack

### LINKED ISSUE #7136: Domain stays in effect after exit() if called in nextTick()
``` javascript

var d = domain.createDomain();
d.on('error', function (err) {
    console.log(err.message);
});

d.enter();

process.nextTick(function () {
    d.exit();
    throw new Error('Eh?');
});
```

Reported originally in https://github.com/spumko/hapi/issues/1299

trevnorris:
If a domain is active when the callback is called, then the domain is
entered. When the user manually enters a domain before the callback is
run then there will be a duplicate copy of the domain in the stack.

Instead remove to the first instance of the domain pushed to the stack.

This will fix #7136

### DIFF
--- a/lib/domain.js
+++ b/lib/domain.js
@@ -138,7 +138,7 @@ Domain.prototype.enter = function() {
 Domain.prototype.exit = function() {
   // skip disposed domains, as usual, but also don't do anything if this
   // domain is not on the stack.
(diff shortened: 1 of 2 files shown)


trevnorris:
yup, nope.

PR #7162, opened 2014-02-21
doc: fix grammar in fs.writeFile

rudolf:


jasnell:
Already landed in nodejs/node

PR #7237, opened 2014-03-04
url: url.parse requires URL encoded strings

Swaagie:
Related to https://github.com/joyent/node/pull/7236 and https://github.com/joyent/node/pull/7234


jasnell:
-1 on landing this. `url.parse` works without the input being encoded (although it ought to be). The existing documentation is accurate. Closing but can reopen if necessary. /cc @joyent/node-tsc

PR #7242, opened 2014-03-04
domain: fix parents tracking

xaka:
We cannot use current stack to keep track of domains chain and their
parents for exceptions handling, we need an additional property to do
so. The reason behind it is an asynchronous code that throws an
exception and so when domain is handling that exception, the stack is
empty and there is no way to get the parent.


jasnell:
At this point, this PR definitely will not land. If there are changes here that need to be captured, then they need to be updated and a new PR opened in nodejs/node

PR #7243, opened 2014-03-04
stream: emit pipe and unpipe on readable

tjfontaine:
Prior to this commit, 'pipe' and 'unpipe' events would only be emitted
on the writable, i.e. the destination to where the readable was being
piped. Now 'pipe' and 'unpipe' will be emitted on both sides of a
pipeline.

Relatedly, if the unpipe was triggered by an error in the writable that
error is now passed to the unpipe event.

### DIFF
--- a/doc/api/stream.markdown
+++ b/doc/api/stream.markdown
@@ -209,6 +209,43 @@ descriptor) has been closed. Not all streams will emit this.
 
 Emitted if there was an error receiving data.
 
+#### Event: 'pipe'
+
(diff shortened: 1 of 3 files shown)


jasnell:
@indutny @chrisdickinson @trevnorris ... any further thoughts on this one? land, close or move to nodejs/node?

jasnell:
If this is a change we'd like to pursue, a new PR against master on nodejs/node would need to be opened. Closing this here.

PR #7260, opened 2014-03-06
http: correct fencepost error when freeing sockets

othiym23:
This only becomes apparent when using a small maxSockets for pipelining requests on a keep-alive connection. Bug introduced in 65f6f06a.

This change doesn't break `make test jslint`, which indicates that there need to be some more tests written around this stuff. Will get to that when I get a chance.

This only affects master, as the keep-alive behavior is very different in 0.10 and earlier.

### DIFF
--- a/lib/_http_agent.js
+++ b/lib/_http_agent.js
@@ -84,7 +84,7 @@ function Agent(options) {
         if (self.sockets[name])
           count += self.sockets[name].length;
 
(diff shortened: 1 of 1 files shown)


jasnell:
@othiym23 ... is this still needed? If yes, is it something that should land in v0.12 or can it be reopened on nodejs/node master?

jasnell:
Issue is still relevant in nodejs/node master. Opened a new PR so it can land there. Closing here.

PR #7292, opened 2014-03-12
android: pass extra arguments to configure script

javihernandez:
With this patch we are able to pass extra arguments to the general
'configure' script, ie: --prefix=/home/john/opt/.

### DIFF
--- a/android-configure
+++ b/android-configure
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+args=( "$@" )
+extra_arguments=""
+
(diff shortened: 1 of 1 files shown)


jasnell:
@orangemocha ... any thoughts?

jasnell:
@javihernandez ... now that we're moving things to nodejs/node, this is something that should be directed to https://github.com/nodejs/build. Going to close this here but encourage you to open a new PR if this is still needed.

Issue #7305, opened 2014-03-13
http: `undefined` headers should not be sent


bminer:
If an HTTP header value is `undefined`, that particular header should not be transmitted to the server.

Example: 

```
var http = require("http");
http.request({
    "host": "google.com",
    "headers": {"If-None-Match": undefined}
});
```

In the case above, the `If-None-Match` header is set to the string `"undefined"`.  Instead, this header should not be sent at all.

The string conversion happens here... see https://github.com/joyent/node/blob/2eddd7411234bf633df6a7bdb9c120ed89853324/lib/_http_outgoing.js#L301

jasnell:
Appears to have been resolved in v0.12, but still happens in v0.10. @joyent/node-coreteam ... should we do a backport or just leave it?

trevnorris:
@jasnell if the fix is trivial then I'd say backport.

jasnell:
At this point, with v0.10 in maintenance, backporting is likely unnecessary. It's fixed in v0.12 and v3.x.

PR #7329, opened 2014-03-19
http: Don't send empty, null or undefined header to server

ssuda:
if client sets any header null or undefined or empty string
  don't send that header to server. fixes issue #7305


trevnorris:
@ssuda Going to close this. Ping or update the PR and I'll reopen.

Issue #7213, opened 2014-02-28
Feature - Expose ZLib crc32() function


JavaScriptDude:
I have a need to do cross platform simple file validation using Node.js. I am presently building a backup file distribution and synchronization system using Node.js and need a solid and fast way of generating a check sum of the files on the sender and receiver sides.

The files are on servers I own so I don't need crypto hashes for my purpose.

I know that zlib has a crc32 function under the hood and it would be great to have this available in JavaScript.

I know that others have written crc32 implementations in JavaScript for Node but IMO, that is not the correct approach.

Any reason why this should not be implemented?

JavaScriptDude:
For the record, I have added a new issue to open a discussion to [Incorporate suite of hash utilities into node #2444](https://github.com/nodejs/node/issues/2444)

PR #7372, opened 2014-03-28
zlib: add adler32 and crc32

seishun:
Continuing from #7213.

I couldn't think of a way to make a streaming interface without over-complicating the API, and it seems like something that could be easily implemented by a user-land library. But if someone can do some benchmarking and show that marshaling the running checksum between JS and C++ causes a significant performance loss, it will be a good reason to reconsider.

The proposed API is pretty straightforward. If it LGTY, I'll add docs and tests.

### DIFF
--- a/doc/api/zlib.markdown
+++ b/doc/api/zlib.markdown
@@ -6,9 +6,9 @@ You can access this module with:
 
     var zlib = require('zlib');
 
-This provides bindings to Gzip/Gunzip, Deflate/Inflate, and
-DeflateRaw/InflateRaw classes.  Each class takes the same options, and
-is a readable/writable Stream.
+This provides bindings to Gzip/Gunzip, Deflate/Inflate, DeflateRaw/InflateRaw,
+as well as Adler32 and CRC32 classes. Each class is a readable/writable Stream,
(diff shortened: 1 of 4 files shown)


jasnell:
@indutny @chrisdickinson @trevnorris ... unfortunately this one seems to have stalled out. Any further thoughts on this one?

JavaScriptDude:
Sure would be great to see some cpu efficient checksum tools in the node.js base for those non-crypto usecases like file change checking algorithms.

jasnell:
@JavaScriptDude ... I agree. At this point tho, I'm just trying to get a sense of where it should land. It's probably not something we'd add to v0.12 at this point so I'm thinking this should be closed here and reopened against the active development stream at http://github.com/nodejs/node

jasnell:
Going to close this here. The PR would never be able to land in it's current state and an updated PR opened against http://github.com/nodejs/node master would be needed. That's not to comment on the validity of this.

JavaScriptDude:
For the record, I have added a new issue to open a discussion to [Incorporate suite of hash utilities into node #2444](https://github.com/nodejs/node/issues/2444)

PR #7377, opened 2014-03-30
util: format add %t placeholder for calling inspect to serialize

Bacra:
`util.format` add `%t` placeholder for calling `inspect` to serialize object
- This is very convenient for calling `util.inspect` method in `util.format`.
- Compared with json(`%j`), it can be serialized width `customInspect` opt. It is useful for serializing other Object (eg: `Data`) in format.


jasnell:
@Bacra .. given that (a) this introduces new functionality, (b) needs to be updated and (c) requires some additional commits as requested, if this is something that you'd like to pursue further, it would be best to close this here, update, and open a new PR against the current active development stream http://github.com/nodejs/node.

Closing this but can reopen if necessary.

PR #7438, opened 2014-04-09
Fix querystring.stringify( {array: []} ). Now it returns "array="

ioncreature:
I found strange behavior of `QueryString.stringify( {array: []} ) === ""`
Other strange cases usually returns something like `"key="`
For example:

``` javascript
QueryString.stringify( {a: [], b: undefined, c: null, d: NaN, e: ""} ); 
// "&b=&c=&d=&e="
```

So for conveniece it should be:

``` javascript
QueryString.stringify( {array: []} )
// "array="
```


jasnell:
At this point, this isn't likely to land here and needs to be updated. Closing this. If you'd like to pursue, please open a new PR against master on http://github.com/nodejs/node.

PR #7508, opened 2014-04-24
util: document the undocumented isXYZ functions

seishun:
@bnoordhuis's [comment](https://github.com/joyent/node/pull/6341#issuecomment-26245786) implies that they aren't meant to be used outside of node core, but I don't see why – they are useful and have well-defined behavior.

Also, currently `util.isObject` returns `false` for functions, which are full-fledged objects. It could be easily fixed by changing the check to `Object(arg) === arg`. If there is a specific reason why functions need to be excluded, then I'll change the docs to mention it.

### DIFF
--- a/doc/api/util.markdown
+++ b/doc/api/util.markdown
@@ -181,6 +181,102 @@ Returns `true` if the given "object" is an `Array`. `false` otherwise.
       // false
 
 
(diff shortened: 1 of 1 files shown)


jasnell:
Already documented in nodejs/node. Closing this here.

PR #7554, opened 2014-05-04
lib/path.js: Use '===' instead of '==' for comparison

stites:
path: Use '===' instead of '==' for comparison

I've started rewriting node modules to practice typing and noticed that
you're evaluating this `if (samePartsLength == 0) {` condition without
type-checking. I thought I might submit a quick pull request to fix
this issue.

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -286,7 +286,7 @@ if (isWindows) {
       }
     }
 
-    if (samePartsLength == 0) {
+    if (samePartsLength === 0) {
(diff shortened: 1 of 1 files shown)


jasnell:
Basic change looks good but PR is out of date and the change is a lower priority. Flagging it to get back to but marking as a low priority

jasnell:
Closing this here. New PR opened https://github.com/nodejs/node/pull/2388

PR #7612, opened 2014-05-13
New failing test: 'finish' firing before _flush is done in async case.

markstos:
The 'finish' event is documented to fire with flushing is done: When "all
data has been flushed to the underlying system".

However, it appears be fired too early in the case of Transform streams
which make an async call. process.nextTick is used here but I ran into this
in a real-world case of making a call to a Mongo database.

This new failing test illustrates the apparent bug. When the code is
_flush is synchronous, there's no problem.

It fails with both 0.10.x and the latest from the 0.11 branch as well.

### DIFF
--- a/test/simple/test-stream2-transform.js
+++ b/test/simple/test-stream2-transform.js
@@ -520,3 +520,39 @@ test('object transform (json stringify)', function(t) {
     t.end();
   })
 });
+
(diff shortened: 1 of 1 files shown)


yinrong:
+1 can't understand after reading the doc

markstos:
@yinrong there is ongoing related discussion on a proposed PR here: https://github.com/nodejs/node/pull/2314

Issue #7348, opened 2014-03-24
Standardize a way for writable streams to do work after all data has been consumed


evansolomon:
Transform streams can optionally have a `_flush` method that lets the stream do something after all incoming data has been consumed, but before the stream's readable side fires its 'end' event. It would be great if a similar convention existed for writable streams.

The particular use case I have in mind here is a writable stream that uploads data to S3 ([source](https://github.com/Obvious/canoe)) that has to tell S3 that the upload is finished, then wait for a response.  In our case, we chose to fire a 'close' event once that work is done.  This is annoying for users though, because they have to know that special event exists.  If they bind to the standard 'finish' event they may try to use the data produced by the writable stream before it's actually ready. Chaos ensues and tears are shed.

The two solutions I've thought about are (1) allowing writable streams to implement a `_flush` method or (2) indicating when the last data is being written so that the stream can delay calling the last callback until it's finished with it's post-processing work.

What do you think?

TomFrost:
To the folks finding this via Google and looking for the best workaround until there's a native _flush implementation, here's my monkey patch. It's a strange approach, but since it doesn't actually patch the stream itself, it's immune to any changes that may be made in reference to this.  I'm using this in production to stream to S3, in a framework where 'finish' absolutely must mean that no part of the stream is still active.

```js
var EventEmitter = require('events').EventEmitter,
    Writable = require('stream').Writable,
    util = require('util');

function MyWriteStream() {
    Writable.call(this, {highWaterMark: 500});
}
util.inherits(MyWriteStream, Writable);

MyWriteStream.prototype.emit = function(evt) {
    if (!Writable.prototype._flush && evt === 'finish') {
        this._flush(function(err) {
            if (err)
                EventEmitter.prototype.emit.call(this, 'error', err);
            else
                EventEmitter.prototype.emit.call(this, 'finish');
        }.bind(this));
    }
    else {
        var args = Array.prototype.slice.call(arguments);
        EventEmitter.prototype.emit.apply(this, args);
    }
};

MyWriteStream.prototype._flush = function(cb) {
    // This now executes identically to Transform._flush.
    // 'finish' will be emitted when cb() is called.
};

module.exports = MyWriteStream;
```

triccardi-systran:
I ran into the exact same need: AWS S3 multipart-upload requires a final HTTP request to *commit* the upload, after all data has been sent.
[...]

PR #7613, opened 2014-05-13
Streams: Add _flush capability to writable streams

evansolomon:
Allows writable streams to define a `_flush()` method that can delay the `finish` event until it completes.

See #7348

### DIFF
--- a/lib/_stream_transform.js
+++ b/lib/_stream_transform.js
@@ -126,12 +126,7 @@ function Transform(options) {
   this._readableState.sync = false;
 
   this.once('prefinish', function() {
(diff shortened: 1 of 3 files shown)


jasnell:
@chrisdickinson @trevnorris ... this never seems to have been completed. Any thoughts on whether this is needed and whether it would need to land in v0.12?

calvinmetcalf:
there is a more comlete pull to do this at nodejs/node#2314

jasnell:
Excellent. Closing this then in favor of that. Can reopen this if it becomes necessary

PR #7623, opened 2014-05-14
missing close event after EPIPE error

juliangruber:
The documentation for a `net.Socket` states [here](http://nodejs.org/api/net.html#net_event_error_1):

> Event: 'error'#
> Error object
> Emitted when an error occurs. The 'close' event will be called directly following this event.

This does however not happen after an `EPIPE` error. I made the `net-write-after-close` test fail by waiting for that event.

I'm not sure how to fix this, of course we could just `self.emit('close')` [here](https://github.com/joyent/node/blob/master/lib/net.js#L278) in `writeAfterFIN`, but it seems like this should rather happen somewhere around [here](https://github.com/joyent/node/blob/master/lib/net.js#L261) in `onSocketEnd()` or `destroySoon()`.

### DIFF
--- a/doc/api/net.markdown
+++ b/doc/api/net.markdown
@@ -336,7 +336,8 @@ The `connectListener` parameter will be added as an listener for the
 
 ### socket.bufferSize
 
-`net.Socket` has the property that `socket.write()` always works. This is to
+`net.Socket` has the property that `socket.write()` always works - as long
+as the socket is still open. This is to
 help users get up and running quickly. The computer cannot always keep up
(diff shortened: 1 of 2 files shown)


jasnell:
@chrisdickinson @trevnorris ... any thoughts on this one?

jasnell:
@juliangruber ... at this point, mainly because of the conflicts, if you still want to pursue this then a new PR opened against master on http://github.com/nodejs/node would be best. Closing this here.

PR #7659, opened 2014-05-21
crypto: fix memleak on non-buffer

mscdex:


jasnell:
Looking at v0.12, it looks like this was fixed at some point. Going to close. @mscdex, can reopen this if necessary.

PR #7669, opened 2014-05-23
npm: upgrade to 1.4.10

isaacs:
- Don't set referer if already set
- fetch: Send referer and npm-session headers
- run-script: Support --parseable and --json
- list runnable scripts (Evan Lucas)
- Use marked instead of ronn for html docs


jasnell:
It's likely safe to assume it's ok to close this one ;-)

PR #7680, opened 2014-05-23
net: throw on write after close

juliangruber:
`socket.write()` on a closed socket will either throw or pass an error to the callback.

see https://github.com/joyent/node/pull/7623#issuecomment-43895947

### DIFF
--- a/doc/api/net.markdown
+++ b/doc/api/net.markdown
@@ -371,6 +371,9 @@ buffer. Returns `false` if all or part of the data was queued in user memory.
 The optional `callback` parameter will be executed when the data is finally
 written out - this may not be immediately.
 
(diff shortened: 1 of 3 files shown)


juliangruber:
any updates?

jasnell:
@chrisdickinson @indutny .. any further thoughts on this one?

indutny:
I'm not sure how well it plays with the streams API that we are exposing now. Probably not that well at all (take a look at `pipe` implementation)

jasnell:
Ok. @juliangruber , given how things have evolved in the streams API, this would need to be revisited if it's still an issue you want to pursue. The right place would be either at http://github.com/nodejs/node or http://github.com/nodejs/readable-stream. Closing this here.

PR #7687, opened 2014-05-25
Console.debug()

piranna:
Add support for Console.debug() function, as latest versions of Chrome & Firefox has. In a logical order, debug() has a lower priority than log(), although it's only intended for developers and it's not part of the normal output of the program, due to this I output it on stderr like warn() & error().

### DIFF
--- a/lib/console.js
+++ b/lib/console.js
@@ -63,6 +63,7 @@ Console.prototype.warn = function() {
 
 
 Console.prototype.error = Console.prototype.warn;
+Console.prototype.debug = Console.prototype.warn;
 
 
 Console.prototype.dir = function(object) {
--- a/lib/console.js
+++ b/lib/console.js
@@ -54,6 +54,7 @@ Console.prototype.log = function() {
 };
 
 
+Console.prototype.debug = Console.prototype.log;
 Console.prototype.info = Console.prototype.log;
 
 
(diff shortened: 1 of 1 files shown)


jasnell:
Given that this would introduce new behavior, if it's still relevant and something that you'd like to pursue, it should likely be closed here and reopened against http://github.com/nodejs/node

piranna:
> Given that this would introduce new behavior, if it's still relevant and something that you'd like to pursue, it should likely be closed here and reopened against http://github.com/nodejs/node

Are you talking about open an issue there pointing here, or about re-creating the pull-request there?

piranna:
There's a formalization in progress of the [Console API](https://github.com/DeveloperToolsWG/console-object/blob/master/api.md) that also takes in account the differences and current lacks on Node.js and other implementations. I think Node.js console should adhere to it and fully implement the API.

jasnell:
Both perhaps. This PR won't be able to land here and you're going to need a
new PR anyway to land anything in the new repo. However, it always that
there's is more to discuss possibly on this so opening an issue may be good
to stimulate the discussion.

jasnell:
As mentioned, this would need a new PR opened against master on http://github.com/nodejs/node. Closing this here!

Issue #4846, opened 2013-02-25
UDP throws unhandled error, even if callback is provided to `send`.


jwalton:
If you create a udp4 socket and send some data over it to an invalid hostname (as in the Coffee-Script example below), then the callback will be called to indicate that an error occurred.  Unfortunately, the udp socket will ALSO throw a top-level exception which kills the entire app.

Sample:

    socket = (require 'dgram').createSocket 'udp4'
    #socket.on "error", (err) ->
    #  console.log "Error handler: #{err}"
    
    buffer = new Buffer("foo")
    callback = (err, bytes) ->
      if err
        console.log("Error sending: #{err}")
      else
        console.log("Sent #{bytes}")
    socket.send(buffer, 0, buffer.length, 100, "asdkasd", callback)

Run with "coffee sample.coffee".  You'll get the following output:

    Error sending: Error: getaddrinfo ENOTFOUND
    
    events.js:71
            throw arguments[1]; // Unhandled 'error' event
                       ^
    Error: getaddrinfo ENOTFOUND
        at errnoException (dns.js:37:11)
        at Object.onanswer [as oncomplete] (dns.js:124:16)

And the problem If you uncomment the two lines at the top, and define a "on error" handler, then the callback will be called with the error, and the error handler will also be called, and the program will not terminate (since the socket is never closed.)

I would expect that either the presence of a callback *or* a defined error handler would prevent killing the program, and not require both?

jasnell:
@bnoordhuis @indutny ... do we still want to do something with this one?

mcollina:
@jasnell there is a PR here: https://github.com/joyent/node/pull/7738

cjihrig:
This is now fixed on the io.js `next` branch in 77266d7fadd8dfefb107ccb1e3fe97f9620f1288 and 90daf870a00305dcb4b7bf474d8eafaed0117242.

jasnell:
Do we want to backport?

cjihrig:
It is a semver major change, so if so, it needs to go into master.

jasnell:
I'd say no backport then. We can pick it up later in the converged stream.

cjihrig:
Works for me.

mcollina:
Great!!

PR #7738, opened 2014-06-05
dgram: make .send cb act as an "error" event handler.

chrisdickinson:
This allows users to provide a callback that handles potential
errors resulting from a `socket.send` call. The original behavior
of emitting the error event on the socket is preserved.

Fixes joyent/node#4846.

### DIFF
--- a/doc/api/dgram.markdown
+++ b/doc/api/dgram.markdown
@@ -100,7 +100,9 @@ assigned a random port number and is bound to the "all interfaces" address
 An optional callback may be specified to detect DNS errors or for determining
 when it's safe to reuse the `buf` object.  Note that DNS lookups delay the time
 to send for at least one tick.  The only way to know for sure that the datagram
-has been sent is by using a callback.
+has been sent is by using a callback. If an error occurs and a callback is given,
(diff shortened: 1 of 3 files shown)


mcollina:
Putting some more though on this one, I still think that if there is a callback, then the error should not be emitted, as they will have to 'trap' that error and ignore it, causing the exact same issue described above.
A safe way to keep backward compatibility might be to check the `callback.length`, as if it has an argument then we can assume the application is going to handle the error. What do you think?

sam-github:
I think we should break compatibility in node v3.x, or else have a different send that does the right thing.

mcollina:
I agree with @sam-github. Do you think it might make sense to send a PR to io.js, or I should wait after the merge happens?

sam-github:
I don't think its worth waiting, and the io.js repo is going to move, to my knowledge, to the new org.

cjihrig:
Closing this. The fix can be back ported from https://github.com/nodejs/io.js/commit/77266d7fadd8dfefb107ccb1e3fe97f9620f1288 and https://github.com/nodejs/io.js/commit/90daf870a00305dcb4b7bf474d8eafaed0117242.

PR #7791, opened 2014-06-16
http: fix http agent legacy support

nitroduna:
ClientRequest calls addRequest with 'localAddress' in V0.10,
(not 'path'). And getName looks for 'localAddress' in the
options object, so it is important to use the localAddress
as part of the key for  "socket" and "freeSocket" maps.

https://github.com/joyent/node/blob/v0.10/lib/http.js#L1416


jasnell:
@trevnorris ... do you see reason to keep this one open?

trevnorris:
not really.

Issue #7031, opened 2014-02-02
path.join() changes casing of drive letter starting with 0.11.0


AlexZeitler:
Running this piece of code on Windows and Node.js 0.10.25 results as shown below.

Code:

```
var path = require('path');
var join = path.join;
var root = __dirname;
var relativePath = '/relativepath'
var joinedPath = join(root, relativePath);
console.log("root:",__dirname);
console.log("relative path:",relativePath);
console.log("joined path:", joinedPath);
```

Result (Node.js 0.10.25 on Windows 7 Pro x64):

```
root: D:\Code\PDMLab\nodetests
relative path: /relativepath
joined path: D:\Code\PDMLab\nodetests\relativepath
```

Result running the same code on Node.js 0.11.0 and later on Windows 7 Pro x64:

```
root: D:\Code\PDMLab\nodetests
relative path: /relativepath
joined path: d:\Code\PDMLab\nodetests\relativepath
```

As you can see, the drive letter now is lower case.

This behavior breaks modules like [koa-send](https://github.com/koajs/send/blob/master/index.js#L67) when it comes to using joined path instances.

PR #7799, opened 2014-06-18
path: `resolve()` doesn't normalize drive letter on Windows

mrfabbri:
Calling `resolve()` on Windows (in Node v0.11.0+) result in a path where drive
letter case isn't always lowercase (e.g. for a relative path like '.') while
`normalize()` always returns a path with a lower case drive letter.
This eventually breaks comparisons between results of `resolve()` and
`join()` (which relies on `normalize()`) for the same path.

Fixes joyent/node#7031 also resolves joyent/node#7806

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -149,6 +149,11 @@ if (isWindows) {
     if (isUnc) {
       resolvedDevice = normalizeUNCRoot(resolvedDevice);
     }
+    // If `resolvedDevice` is a drive letter, we'll normalize to lower case.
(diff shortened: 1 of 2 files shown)


jasnell:
@orangemocha ... thoughts?

orangemocha:
I believe this was fixed in https://github.com/joyent/node/commit/f6e574018090ed4d63596b8a3bb614f8f48b6267. The behavior was later changed in https://github.com/joyent/node/commit/016e08458cc38d7324c907c6594f2b56a5a19719, to avoid modifying the case of the drive letter.

PR #7856, opened 2014-06-28
stream: only end reading on null, not undefined

reem:
The [Stream documentation for .push](http://nodejs.org/api/stream.html#stream_readable_push_chunk_encoding)
explicitly states multiple times that null is a special cased value
that indicates the end of a stream. It is confusing and undocumented
that undefined _also_ ends the stream, even though in object mode
there is a distinct and important difference.

The docs for Object-Mode also explicitly mention null as the _only_
special cased value, making no mention of undefined.

### DIFF
--- a/lib/_stream_readable.js
+++ b/lib/_stream_readable.js
@@ -139,7 +139,7 @@ function readableAddChunk(stream, state, chunk, encoding, addToFront) {
   var er = chunkInvalid(state, chunk);
   if (er) {
     stream.emit('error', er);
-  } else if (util.isNullOrUndefined(chunk)) {
+  } else if (chunk === null) {
     state.reading = false;
     if (!state.ended)
       onEofChunk(stream, state);
--- a/test/simple/test-stream2-readable-non-empty-end.js
+++ b/test/simple/test-stream2-readable-non-empty-end.js
@@ -35,7 +35,7 @@ var n = 0;
 test._read = function(size) {
   var chunk = chunks[n++];
   setTimeout(function() {
-    test.push(chunk);
+    test.push(chunk === undefined ? null : chunk);
   });
 };
 


brycebaril:
Ok, just spent some time debugging some modules which broke because of this change. I also am not a fan.

brycebaril:
This also diverges from http://npm.im/readable-stream which does not have this patch -- @rvagg or @chrisdickinson any thoughts here?

rvagg:
Need @chrisdickenson to weigh in on this but messing with the external API for streams is a huge risk at this stage and shouldn't be done lightly. There's a ton of code in npm that depend on minor details of the actual implementation and changes like this _may_ be best seen as documentation problems rather than implementation problems. 

That said, if you want stability, use readable-stream@1.0.x, this is a 1.1.x detail for streams3.

chrisdickinson:
I am filled with unease by the thought that streams2 streams will still roam the earth despite streams3 being available. If this change is going to encourage that outcome, I'd almost prefer to back it out. On the other hand, we may be moving to a world where `undefined` and `null` are not "out of alphabet" for streams in the future, and this change is a first step in that direction (if unintentionally).

I think I have to come down on the side of backing this change out. If we're going to get to a world where `undefined` and `null` are in-alphabet, we have to do so in a way that is messaged in big bold letters to everyone it affects, with ample time to make the necessary changes.

@brycebaril I am curious, for future reference: how did the bug manifest?

brycebaril:
@chrisdickinson in http://npm.im/stream-spigot (a Readable stream helper) there is an Array helper that creates streams from arrays e.g. `spigot.array(["these", "will", "be", "stream", "chunks"])`

It is (roughly) implemented as `this.push(array.shift())` and thus the `undefined` was signalling the end.

The specific manifestation was in an objectMode stream causing an undefined property access error.

Issue #7006, opened 2014-01-30
Cluster module can lose track of connections if child processes die


msmithrdio:
Our application uses the cluster module, and periodically we kill worker processes - the master re-spawns them, and new connections can go to the new workers.

We're now trying to run 0.11.10, and occasionally, a connection is 'lost' - it's in limbo, not disconnected, nor processed by a child.

I believe this happens in the following scenario:
 - Master accepts a connection
 - Master passes new connection to worker with sendmsg
 - Before worker can recvmsg to get the passed file descriptor...
 - Worker process is terminated
 - Master process leaves the socket open

I will attach a simple minimized test script that reproduces this problem for me (it works fine in 0.10.x).

jasnell:
@cjihrig @msmithrdio ... is this still an issue? Do we need to backport anything to v0.10?

rdio-reviewboard:
Unsure if this is an issue on recent 0.12 (I haven't tried updating/removing our workarounds) - but this was a regression in 0.11, so backporting to 0.10 is definitely not required.

jasnell:
I'm unable to reproduce in v0.12 or io.js so going to close this for now then. Can reopen if new information is received.

PR #7928, opened 2014-07-10
test: regression test based on #7006

rmg:
A test based on the description in #7006
1. Start cluster master, worker
2. Connect to cluster with separate client
3. Worker dies badly
   4a. Client gets disconnected (reliably on v0.10)
   4b. Client does not get disconnected (frequent on v0.11)

Seems to always pass on v0.10, but intermittently fail on v0.11. Having random worker deaths result in hung clients makes cluster a lot less appealing.


jasnell:
@rmg ... if you'd still like to land this, can I ask you to open a new PR against master on http://github.com/nodejs/node ?

rmg:
Sure, I'll open a new PR for it later.

PR #7944, opened 2014-07-14
Update fs.markdown to account for correct uses of fs.exists()

oconnore:
`fs.exists()` is not an anti-pattern in itself. It becomes an anti-pattern when used in a particular way. Checking the existence of a file is quite useful when, for example, checking filesystem based locks or status files (of course, _grabbing_ a lock is the same problem, and should use `fs.open()` directly), and the documentation shouldn't be so narrow in its assumptions.


jasnell:
Closing this here. A new PR has been opened in http://github.com/nodejs/node

PR #7988, opened 2014-07-23
Update fs.markdown

ghost:
Inscrease informations : "mainly if you watch a file but for a folder, it is unlikely."

On mac (Darwin Kernel Version 13.2.0) : 
## Watcher on directory :

```
require('fs').watch("./", function() {
    console.log(arguments);
});
```

Output : `{ '0': 'change', '1': 'toto.js.lk' }`
## Watcher on file :

```
require('fs').watch("./toto.js.lk", function() {
    console.log(arguments);
});
```

Output : `{ '0': 'change', '1': null }`

### DIFF
--- a/doc/api/fs.markdown
+++ b/doc/api/fs.markdown
@@ -624,9 +624,10 @@ less reliable.
 
 <!--type=misc-->
 
-Providing `filename` argument in the callback is not supported
-on every platform (currently it's only supported on Linux and Windows).  Even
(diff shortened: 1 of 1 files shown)


jasnell:
I'm -1 on this change. Don't view it as being necessary and there are a number of grammatical issues. Also, given that it's been sitting here for a while, going to go ahead and close. /cc @joyent/node-tsc

PR #8002, opened 2014-07-26
fs: Add DuplexStream and fs.createDuplexStream

benjamincburns:
This adds a `DuplexStream` and matching `createDuplexStream` function to the fs module. Similar to how `Duplex` uses parasitic inheritance from the `Writable` class, I've implemented this by inheriting prototypically from `Duplex` and parasitically from `ReadStream` and `WriteStream`.

Please note that I had to modify slightly the `ReadStream` class to make this work. The only modification necessary was to prefix the `end` member (undocumented and used in a private context), so it is now `_end`. This prevents this field from clashing with the `end` method specified by the `Duplex` stream.

### DIFF
--- a/doc/api/fs.markdown
+++ b/doc/api/fs.markdown
@@ -797,6 +797,45 @@ Emitted when the WriteStream's file is opened.
 The number of bytes written so far. Does not include data that is still queued
 for writing.
 
+
+## fs.createDuplexStream(path, [options])
+
+Returns a new DuplexStream object (See `Duplex Stream`).
+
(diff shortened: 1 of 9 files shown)


jasnell:
Given that there's been no further motion on this and it looks to introduce a significant new feature, I'm inclined to close this here and ask that if it's still something that is wanted, please update and open a new PR against master on http://github.com/nodejs/node. /cc @joyent/node-tsc

benjamincburns:
@jasnell - yes, this is something that is still wanted. I'll do as suggested.

PR #8010, opened 2014-07-28
test: fix leak detection for non-enumerables

geek:
- Non-Enumerable global leaks are now detected
- Check by key name not by reference


jasnell:
@geek not sure if this was ever reviewed. It's also not sure what problem this was trying to solve. If you'd like to pursue this, closing this here and opening a new PR against master on http://github.com/nodejs/node would be best.

PR #8021, opened 2014-07-29
tls: Server should honor previous "timeout" listeners

### LINKED ISSUE #7764: https: Node HTTPS server has no socket timeout by default
The **HTTP** (not HTTPS) documentation reads:

```
By default, the Server's timeout value is 2 minutes, and sockets are
destroyed automatically if they time out. However, if you assign a
callback to the Server's 'timeout' event, then you are responsible
for handling socket timeouts.
```

This is not true for **HTTPS** servers.  As you can see from [this snippet](https://github.com/joyent/node/blob/v0.10.22/lib/tls.js#L1166-L1171), `tls.js` is overriding the socket timeout during the TLS handshake (setting it to `options.handshakeTimeout || (120 * 1000)`, as appropriate).  Then, after the TLS handshake, the socket timeout is disabled completely.

This behavior should be changed, or the documentation for the HTTPS server should be updated.

This particular issue has caused me massive headaches, as our Node HTTPS server has been crashing due to hanging connections (we have a lot of 3G modems connecting to our server).

As a workaround, I am using the following to restore the default 2 minute timeout:

```
server.on("secureConnection", function(clearText) {
    clearText.setTimeout(2 * 60 * 1000);
});
```

**EDIT**: I should also note that `clearText.setTimeout` is also **NOT** documented on the Node API.

bminer:
This fixes #7764 and allows HTTPS servers to follow the same socket timeout behavior as HTTP servers.

### DIFF
--- a/lib/_tls_wrap.js
+++ b/lib/_tls_wrap.js
@@ -635,6 +635,8 @@ function Server(/* [options], listener */) {
     });
 
     socket.on('secure', function() {
+      // enable normal socket timeout after handshake
+      socket.setTimeout(self.timeout || 0);
       if (socket._requestCert) {
         var verifyError = socket.ssl.verifyError();
         if (verifyError) {
--- a/lib/_tls_wrap.js
+++ b/lib/_tls_wrap.js
@@ -636,7 +636,7 @@ function Server(/* [options], listener */) {
 
     socket.on('secure', function() {
       // enable normal socket timeout after handshake
-      socket.setTimeout(self.timeout || 0);
+      socket.setTimeout(self.timeout | 0);
       if (socket._requestCert) {
         var verifyError = socket.ssl.verifyError();
         if (verifyError) {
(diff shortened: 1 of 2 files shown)


jasnell:
@bminer @indutny ... what's the current status on this?

bminer:
I am attempting a one-liner change, and I don't know how to write a test for it.

This is an attempt to fix #7764.

jasnell:
Ok @bminer ... this is definitely not going to be able to land here in it's current state. Given that the PR needs to be updated anyway, I'm going to go ahead and close this. Once you've got things cleaned up on your end, please open a new PR against master on http://github.com/nodejs/master (which is the new active development branch). If you need any further assistance on it, feel free to ping.

bminer:
This is really a bummer.  I mean... this PR is a one-liner bug fix, and the amount of dialogue for it is a tad intense.  The related issue is #7764.  I'm giving up and letting someone else fix the problem.  :)

PR #8090, opened 2014-08-06
src: expose libuv handle/req id on process object

bnoordhuis:
Assign numeric identifiers to libuv handle and request wrappers and
expose them to JS land as process._uvContextId[0].  Provides an ersatz
"thread-local" state for tracking execution across asynchronous calls.

Future optimization: reuse identifiers so they keep fitting in a SMI?

R=@trevnorris?

This PR is a conversation starter.  The goal is to have a cheap way of tracking the flow of execution across callbacks for instrumentation purposes.

### DIFF
--- a/src/async-wrap-inl.h
+++ b/src/async-wrap-inl.h
@@ -35,12 +35,29 @@
 
 namespace node {
 
+inline AsyncWrap::ScopedExposeId::ScopedExposeId(AsyncWrap* wrap)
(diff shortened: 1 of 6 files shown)


jasnell:
@bnoordhuis ... any further thoughts on this? is it still necessary?

bnoordhuis:
We can revisit this, it's useful, but I'll close it for now.

Issue #8105, opened 2014-08-08
setTimeout drifts after busy wait


bjouhier:
if you call `setTimeout` after a busy wait the delay can be much longer than expected.

I created a gist: https://gist.github.com/bjouhier/c7adb5f54caf9fc86d6e

OSX 10.9.4, node 0.10.27

misterdjules:
Fixed by d38e865fce93c00cb128034202fb1e26987efdbc. Released in v0.10.39 and v0.12.5.

natbro:
nice fix - thanks & sorry i didn't spot it.

misterdjules:
@natbro No worries, and thank you for investigating and reporting your findings :+1:

Issue #8066, opened 2014-08-03
setInterval doesn't comply with DOM-version


loyd:
Checked on x86_64, v0.10.30 and v0.11.14-pre, linux.

Firslty, `setInterval` determines delay between _executions_. For `setInterval(sleep(20), 100)`:

```
------|====----------------|====----------------|====---
     0    20              120  140             240  260
```

DOM-version `setInterval` determines delay between _runs_:

```
------|====-----------|====-----------|====-----------|=
     0    20         100  120        200  220        300
```

Use the following snippet to check it out:

``` javascript
var prev = Date.now();
setInterval(function() {
  var s = Date.now();
  while(Date.now() - s < 20);
  var e = Date.now();
  console.log('Start', s, 'End', e, 'Delta', s - prev);
  prev = e;
}, 100);
```

Node v0.11.14-pre 'Delta': 102, 124, 119, 120, 120, ...
Node v0.10.30 'Delta': 101, 102, 100, 100, 100, ...
DOM 'Delta': 100, 80, 81, 80, 80, 81, ...

Secondly, overlay of calls:

``` javascript
var prev = Date.now();
setInterval(function() {
  var s = Date.now();
  while(Date.now() - s < 200);
  var e = Date.now();
  console.log('Start', s, 'End', e, 'Delta', s - prev);
  prev = e;
}, 100);
```

Node v0.11.14-pre 'Delta': 101, 302, 300, 300, 300, ...
Node v0.10.30 'Delta': 101, 101, 100, 100, ...
DOM 'Delta': 101, 0, 0, 1, 0, 1, 0, 0, ...

PR #8120, opened 2014-08-09
Fix timers

loyd:
Fixed:
1. `this` within unref'd timer point to `_handle`, not timer.
2. Now `listOnTimeout` updates `now` for every element of the list. It gives independence from the time of callbacks. See #8133 and #8105.
3. Behavior of timers doesn't comply with browser version. See #8066.


jasnell:
@trevnorris ... any further thoughts on this? Any reason to keep this open?

jasnell:
Given the nature of this change, a new PR would need to be opened against master on http://github.com/nodejs/node if this is something that should be pursued. Closing this here.

PR #8145, opened 2014-08-12
module: normalize path when computing cache key.

### LINKED ISSUE #7031: path.join() changes casing of drive letter starting with 0.11.0

misterdjules:
When adding a new module to the cache, and also when querying the cache,
normalize the path so that modules don't get loaded more than once when:
- They are referenced by the same file name but with different case for
  the drive letter on Windows.
- They are referenced by two different paths that point to the same
  file.

Tested on Windows, Ubuntu 12.04, MacOS X and SmartOS.

Fixes #7031.

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -276,8 +276,9 @@ Module._load = function(request, parent, isMain) {
   }
 
   var filename = Module._resolveFilename(request, parent);
+  var cacheKey = path.normalize(filename);
 
-  var cachedModule = Module._cache[filename];
+  var cachedModule = Module._cache[cacheKey];
   if (cachedModule) {
     return cachedModule.exports;
   }
@@ -302,7 +303,7 @@ Module._load = function(request, parent, isMain) {
     module.id = '.';
   }
 
-  Module._cache[filename] = module;
+  Module._cache[cacheKey] = module;
 
   var hadException = true;
 
@@ -311,7 +312,7 @@ Module._load = function(request, parent, isMain) {
     hadException = false;
   } finally {
     if (hadException) {
-      delete Module._cache[filename];
+      delete Module._cache[cacheKey];
     }
   }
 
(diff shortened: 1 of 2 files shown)


Jokero:
When are you going to merge it? I think it is critical issue for Windows users

jasnell:
@misterdjules ... are you able to get back to this?

FransGH:
There seem to be two core issues coming up in the various discussions around case sensitivity of require():
1) on case sensitive files systems, some modules might not load (throwing an error)
2) on case insensitive file systems, calling require() for the same module but with different case loads the module multiple times (breaking singletons)

The 2nd issue really sounds like a bug that this pull request elegantly resolves.

What would be required to get this merged?

FransGH:
Workaround for singletons:

In your module add the following code at the top:

if(global.__didimport_MyModuleName_) {
    throw("MyModuleName already imported - check if all filenames have equal case");
}

global.__didimport_MyModuleName_ =true;

PR #8147, opened 2014-08-12
doc: more methods from the util module.

majimboo:
Something that should be available on the documentation because most people are manually doing it not knowing they can easily use the util module.


jasnell:
Documentation for these has already landed in nodejs/node. Closing this.

PR #8156, opened 2014-08-13
lib: remove variables root and GLOBAL

JacksonTian:
There is no document and API use the root and GLOBAL variables. I think we can remove it from core.


jasnell:
@joyent/node-tsc ... thoughts? are these actually used?

jasnell:
@chrisdickinson

cjihrig:
I think we should close this. Since August 2014, when this was opened, we've been bitten a number of times by thinking it was OK to just remove things.

PR #8178, opened 2014-08-14
fd leak in ReadStream.destroy()

rlidwka:
defensive programming is bad sometimes :)

this leaks:

``` js
setInterval(function() {
  var x = require('fs').createReadStream('/etc/passwd');
  x.destroy();
}, 1000);
```

this doesn't:

``` js
setInterval(function() {
  var x = require('fs').createReadStream('/etc/passwd');
  x.close();
}, 1000);
```

all details are here somewhere: https://github.com/visionmedia/send/issues/55

I wonder how do I write tests for those...

### DIFF
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -1588,9 +1588,7 @@ ReadStream.prototype.destroy = function() {
   if (this.destroyed)
     return;
   this.destroyed = true;
-
-  if (util.isNumber(this.fd))
-    this.close();
+  this.close();
 };
 
 


jasnell:
@piscisaureus @indutny ... looks like this went stale but nothing was ever changed in v0.10 or v0.12.

Fishrock123:
@jasnell this is fixed in io.js, I'll try to dig up the commit later.

jasnell:
Yep saw that. We may want to backport the fix. But this particular PR can be closed.

PR #8180, opened 2014-08-15
net: make Socket.connect() consistently asynchronous

### LINKED ISSUE #8140: net.createConnection seems to assume IPv4 localhost address when no host given
It seems that Socket.prototype.connect assumes an IPv4 localhost address when no host is given. Here's the code in `lib/net.js`:
```
} else if (!options.host) {
    debug('connect: missing host');
    self._host = '127.0.0.1';
    connect(self, self._host, options.port, 4);
```
Should we instead set options.host to 'localhost' and fall through so that both use cases (host set in options or not) resolve to the same IP address?

There are other places where it seems IPv4 addresses are assumed:

1) In `Socket.prototype.listen`, when no IP is passed, IPv4 is hardcoded:
```
} else if (util.isUndefined(arguments[1]) ||
             util.isFunction(arguments[1]) ||
             util.isNumber(arguments[1])) {
    // The first argument is the port, no IP given.
    listen(self, null, port, 4, backlog);
```

2) There's another case in Socket.prototype.connect when we hard-code IPv4, I'm not sure what use case it covers:
```
} else {
        timers._unrefActive(self);

        addressType = addressType || 4;

        // node_net.cc handles null host names graciously but user land
        // expects remoteAddress to have a meaningful value
        ip = ip || (addressType === 4 ? '127.0.0.1' : '0:0:0:0:0:0:0:1');

        connect(self,
                ip,
                options.port,
                addressType,
                options.localAddress,
                options.localPort);
      }
```

/cc @cjihrig

cjihrig:
`Socket.prototype.connect()` follows a synchronous code path in some cases and an asynchronous one in others. This commit makes it consistently asynchronous. This commit also removes hard coded IPv4 addresses, and throws on bad inputs immediately, instead of potentially after an asynchronous operation. Closes #8140

EDIT: I might have worded this slightly wrong. The connection is made asynchronously. I was referring to the fact that one current code path involved a DNS lookup, while the other two completed synchronously.

### DIFF
--- a/lib/net.js
+++ b/lib/net.js
@@ -769,34 +769,18 @@ function connect(self, address, port, addressType, localAddress, localPort) {
   assert.ok(self._connecting);
 
   var err;
+
   if (localAddress || localPort) {
-    if (localAddress && !exports.isIP(localAddress))
-      err = new TypeError(
-          'localAddress should be a valid IP: ' + localAddress);
-
-    if (localPort && !util.isNumber(localPort))
-      err = new TypeError('localPort should be a number: ' + localPort);
-
     var bind;
 
-    switch (addressType) {
-      case 4:
-        if (!localAddress)
-          localAddress = '0.0.0.0';
-        bind = self._handle.bind;
-        break;
-      case 6:
-        if (!localAddress)
-          localAddress = '::';
-        bind = self._handle.bind6;
-        break;
-      default:
-        err = new TypeError('Invalid addressType: ' + addressType);
-        break;
-    }
-
-    if (err) {
-      self._destroy(err);
+    if (addressType === 4) {
+      localAddress = localAddress || '0.0.0.0';
+      bind = self._handle.bind;
+    } else if (addressType === 6) {
+      localAddress = localAddress || '::';
+      bind = self._handle.bind6;
+    } else {
+      self._destroy(new TypeError('Invalid addressType: ' + addressType));
       return;
     }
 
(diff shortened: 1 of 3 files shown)


cjihrig:
@trevnorris Friday ping

lib/net.js
```diff
@@ -814,16 +798,11 @@ function connect(self, address, port, addressType, localAddress, localPort) {
   }
 
   var req = { oncomplete: afterConnect };
-  if (addressType === 6 || addressType === 4) {
-    port = port | 0;
-    if (port <= 0 || port > 65535)
-      throw new RangeError('Port should be > 0 and < 65536');
```
trevnorris: Sorry I didn't notice this before. Why removal of the port range check?

lib/net.js
```diff
@@ -879,19 +858,26 @@ Socket.prototype.connect = function(options, cb) {
...
     };
 
+    if (localAddress && !exports.isIP(localAddress))
+      throw new TypeError('localAddress must be a valid IP: ' + localAddress);
+
+    if (localPort && !util.isNumber(localPort))
+      throw new TypeError('localPort should be a number: ' + localPort);
+
+    if (port <= 0 || port > 65535)
+      throw new RangeError('port should be > 0 and < 65536: ' + port);
```
trevnorris: Ah, this must replace the above check.

test/simple/test-net-localerror.js
```diff
@@ -23,39 +23,30 @@ var common = require('../common');
...
+    host: 'localhost',
+    port: 65536
+  }, 'port should be > 0 and < 65536: 65536');
 
-  test1(test2);
-})
+  connect({
+    host: 'localhost',
+    port: 0
+  }, 'port should be > 0 and < 65536: 0');
```
trevnorris: I assume this is the port range error check that @tjfontaine requested.

trevnorris:
@cjihrig Things LGTM. Go ahead and squash it, add the usual commit fields and land it. Thanks. :)

lib/net.js
cjihrig: Correct.

test/simple/test-net-localerror.js
cjihrig: Yep

cjihrig:
Landed in b636ba8186d191c52ee36f2f2b1aebbbb4d95575

PR #8189, opened 2014-08-15
Create less IO errors to improve efficiency of require

errendir:
In some environment (especially when coffescript is involved) module.js
will spend a lot of time looking up nonexisting files. This is caused by the
lengthy error creation. Those errors are then immediately caught.

This change makes the lookup much faster, by avoiding the unnecessary error
creation.

module::statPath now uses the new second parameter of fs.statSync function of the
fs.js. There is a similar change for the fs.stat, adding the third parameter.

### DIFF
--- a/doc/api/fs.markdown
+++ b/doc/api/fs.markdown
@@ -153,11 +153,14 @@ Only available on Mac OS X.
 
 Synchronous lchmod(2).
 
-## fs.stat(path, callback)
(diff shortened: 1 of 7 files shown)


jasnell:
@trevnorris ... any further thoughts on this?

jasnell:
Actually... closing this in favor of https://github.com/joyent/node/pull/8312. Can reopen if necessary.

PR #8212, opened 2014-08-20
events: Added support for custom context

### LINKED ISSUE #8210: events: Introducing a third "context" argument to eventemitter.on
#### Problem

The `EventEmitter` calls the supplied functions with it's own context. If you want to pass in a function to an event emitter from your own class you either need to wrap or bind your function which causes loss of performance. See the following example:

``` js
function Foo() {}
Foo.prototype.handle = function handle(req, res) {
  this.headers(res);
  res.end('bar')
};
Foo.prototype.headers = function headers(res) {
  res.setHeader('foo', 'bar');
}

var http = require('http').createServer(),
   foo = new Foo();

http.on('request', foo.handle.bind(foo));
// or
http.on('request', function (req, res) { foo.handle(req, res) })
```

This isn't just a problem with http modules but everything in node that inherits from the eventemitter.
#### Solution

The solution is to add an extra `context` argument to the `EventEmitter.on` method which will 
ensure that the eventlistener is called with the supplied context instead of the context of the eventemitter. It would be fully backwards compatible as the third argument is not used:

``` js
http.on('request', foo.handle, foo);
```

An example implementation of this could be found in the `eventemitter3` module https://github.com/3rd-Eden/EventEmitter3/blob/master/index.js#L32-L102

As for the performance gain:

```
Node's EventEmitter using bind
Count (66192), Cycles (6), Elapsed (5.485), Hz (1272665.3236835978)

EventEmitter3 with using the 3rd argument for context
Count (363111), Cycles (4), Elapsed (5.412), Hz (6888645.436669402)

EventEmitter3 using bind
Count (70391), Cycles (6), Elapsed (5.4), Hz (1339867.5996333812)
```

3rd-Eden:
The custom context argument allows you to specify with what context
your supplied event handler is executed. It defaults to the
EventEmitter class.

Fixes #8210

### DIFF
--- a/doc/api/events.markdown
+++ b/doc/api/events.markdown
@@ -30,8 +30,8 @@ trace and exit the program.
 All EventEmitters emit the event `'newListener'` when new listeners are
 added and `'removeListener'` when a listener is removed.
 
-### emitter.addListener(event, listener)
-### emitter.on(event, listener)
+### emitter.addListener(event, listener, [context])
+### emitter.on(event, listener, [context])
 
 Adds a listener to the end of the listeners array for the specified event.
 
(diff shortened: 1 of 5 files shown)


jasnell:
@trevnorris @indutny ... hmmm this seems to have been intended for v0.12 but never actually landed :-/

@3rd-Eden , if this is something you'd still like to pursue, it would need to be closed here and reopened against master on http://github.com/nodejs/node (which is the current active development branch). Would you be willing to get that updated?

3rd-Eden:
@jasnell If there intention is there to land this feature then yes, i'll update the pull request and make one against nodejs/node

jasnell:
I can't say for certain that it would land yet but it certainly would get the appropriate review from a much larger audience.

jasnell:
As previously mentioned, this is not going to be able to land in v0.12 so I'm going to close it off here. @3rd-Eden , I would encourage you to open a new PR against master on nodejs/node

PR #8277, opened 2014-08-27
child_process: add type checking for "args" param

### LINKED ISSUE #8249: Child_process execFile() silently failing
This is not a bug, more of an inconvenience and pain to debug.

If you call Child_process's `execFile(file, [args], [options], [callback])` with arg being a string (rather than an array like the API specifies), it will fail silently.

What I mean is that the process will hang and nothing will come out of either `error`, `stdout` or `stderr`. So not output to work with like a `[Error: Command failed: ]` or `ENOENT`.

I'd imagine that other, more recent versions of Node.js as well as other `child_process` module functions would be impacted by this issue.

Node version: v0.10.28

chrisdickinson:
This commit ensures that the "args" parameter is either:
1. Undefined or null (defaulting "args" to []).
2. The callback parameter (defaulting "args" to []).
3. An array.

All other values will throw a TypeError.

Fixes #8249.

### DIFF
--- a/lib/child_process.js
+++ b/lib/child_process.js
@@ -610,6 +610,10 @@ exports.execFile = function(file /* args, options, callback */) {
   if (Array.isArray(arguments[1])) {
     args = arguments[1];
     options = util._extend(options, arguments[2]);
+  } else if (arguments[1] && callback !== arguments[1]) {
+    throw new TypeError(
+        "Parameter 'args' must be an array, not a " + typeof arguments[1]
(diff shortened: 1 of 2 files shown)


jasnell:
@chrisdickinson ... given that this was opened against v0.10, any reason to keep this open?

jasnell:
@chrisdickinson .. going to close this here. given the current status, there's really no reason to land this in v0.10.

PR #8294, opened 2014-08-30
Build fixes from MSYS2/MinGW-w64

mingwandroid:
I work on the MSYS2 project and node.js is one of our packages. I'd like upstream my patches and would like to have a discussion about this set of changes first.


jasnell:
Closing since it appears this already landed in io.js. We'll pick it up there. cc @joyent/node-tsc

PR #8306, opened 2014-09-02
Pipeline Errors - Forwarding errors down stream pipelines

chrisdickinson:
**Note:** This is by way of proposal -- I'm opening this now in its current form to get more comments. In  particular, I have the following questions:

Would it be better to distinguish the error re-throwing behavior from the unpiping behavior? This would entail
splitting up `handleError` into `handleError` (to prevent re-throwing the error) and `preventDefault` (to prevent unpiping) -- or alternatively, should the presence of a pipelineError handler be enough to prevent the re-throwing of errors?

cc @phated @contra @tjfontaine @hayes @dominictarr @substack
# pipeline errors

Writing correct streaming programs is difficult. A single stream missing an
error handler in a pipeline has the ability to crash your entire program, often
with a cryptic error message. Even when an error listener is added, the stream
system has already unpiped by the time client code is called. This makes it
difficult to generically gather pipeline information on error events.
Finally, error events are not necessarily fatal to the pipeline, but the user
has no way to communicate this intent back to the stream subsystem -- any error
emitted _will_ destroy whatever pipeline it is emitted in, at present.
To reiterate: error handling in pipelines requires users to remember to add
error listeners to every constituent stream, does not allow users to prevent
undesired default behavior, and in the event of an error, does not give users
enough information to debug the situation.

The goals and constraints of this proposal are as follows:
- Make correct error handling easier to implement for end users.
- The solution must be polyfillable for older versions of node.
- The solution must work in browser (for browserify).
- Work with extant stream solutions (through, through2).
- Provide public methods into internal state for being able to traverse
  up and down pipelines.
- Implement the solution in terms of public methods, inasfar as possible.

To that end, this proposal adds six methods (in three pairs) to the Stream class.
- `stream.next()` and `stream.prev()`: Get an array of the set 
  of _immediate_ destination or source streams.
- `stream.nextAll()` and `stream.prevAll()`: Get an array of the 
  set of all destination or source streams, with no duplicates or cycles.
- `stream.addPipelineErrorHandler(fn)` and `stream.removePipelineErrorHandler(fn)`: 
  Add a pipeline error handler to a stream.
## User-facing API

``` javascript
var A = fs.createReadStream('input');
var B = new stream.PassThrough();
var C = fs.createWriteStream('output');

A.name = 'A';
B.name = 'B';
C.name = 'C';

C.addPipelineErrorHandler(function(ev) {
  console.log(ev.error.message, ev.stream.name);
  ev.handleError();
});

A.pipe(B).pipe(C);
A.emit('error', new Error('hello!')); // logs "hello!".
```

For example, in the above program, someone has maliciously injected an error
event into our stream pipeline. The error originates on `A`, but is handled at
`C`. The error handler indicates that the user wishes to manually handle this error
event and prevent the unpiping behavior with `ev.handleError()`. The original error
and the originating stream are available at `ev.error` and `ev.stream`, respectively.

The error event may only be marked "handled" on the same event loop turn as it was
emitted. Multiple pipeline error handlers may intercept an error -- it propagates downstream,
hitting handlers "A" then "B" then "C". Once an error event is "handled", it stops propagating
entirely. For complex pipelines -- where there may be cycles, or multiple branches, it propagates
down the "oldest" pipelines first, stopping when it hits duplicates. A few examples:

```
A - B - A     error at A, propagates to A, then B.
*

A - B - A     error at B, propagates to B, then A.
    *

A - B - C     error at A, propagates to A, B, then C.
*

A - B - C     error at B, propagates to B, then C.
    *

       C - D  error at A, propagates to A, B, C, D, X, then Y.
      /       NB: handlers on *either* branch have the opportunity
A - B         to "handle" the error and thus stop propagation.
*     \
       X - Y
```

A user may remove a handler by calling `stream.removePipelineErrorHandler(fn)` with the handler
function they originally provided.
## Implementation

The sequence of events during an error is as follows:
1. A.emit("error")
2. stream .pipe "error" handler (or `_defaultErrorHandler`).
3. Internal "_preError" event.
4. pipelineErrorHandlers for A, and all downstream streams (gathered with `nextAll`) until error is handled or streams exhausted.
5. if the error is handled, skip to step 8.
6. unpipe().
7. re-emit error if no other error handlers exist.
8. The rest of the "error" handlers on A.

In order to implement this sequence of events, the following modifications and
additions were made:
### Added `_nextStreams` and `_prevStreams` attributes to base Stream class

The stream constructor grew two new arrays -- `_nextStreams` and
`_prevStreams`. These are arrays tracking the next and previous streams, and
are managed internally by the `.pipe` method (both `Readable#pipe` and
`Stream#pipe` may affect these arrays. Since "correct" subclassing isn't
guaranteed (see crypto's `LazyTransform`), the attributes are [lazily](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/_stream_readable.js#L466-L467) [added](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L230-L231) 
by `*#pipe` if they are not present. These attributes are there to support 
the `{next,prev}{All,}()` methods. **`_nextStreams`** is a slight duplication
of information in the `Readable` case, but it is only ever represented by an
Array, while `_readableState.pipes` may be null, a single Stream, or an array
of Streams.
### Added `{next,prev}{All,}()` methods

To give users better tools to debug errors within streams, the `next()`,
`prev()`, `nextAll()`, and `prevAll()` methods were added to publicly expose
what was previously private or implicit state. `nextAll` and `prevAll` are implemented
in terms of `iterateStream(stream, attributeName)` ([here](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L61-L67) and [here](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L202-L225)), which does a (non-recursive) depth
first traversal of the `_nextStreams` (or `_prevStreams`) property. One potentially
controversial bit is the [use of a `WeakSet`](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L92) to ensure no duplicates are added to the
output -- this could be replaced with an array of "seen" streams in environments that
don't support `WeakSet`.
### Modified `onerror` in `{Readable,Stream}#pipe`

This is the primary change:

``` javascript
function onerror(er) {
  var handled = {handled: false};
  this.emit('_preError', er, handled);

  if (handled.handled) {
    return;
  }

  // ...
```

This has been added for [old streams](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L279-L284) as
well as [new streams](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/_stream_readable.js#L563-L568). The goal
was to keep all of the pipeline error event handling **out** of the pipe functions. Earlier attempts, where the pipelineError event was
emitted as a "normal" EE-style event, and the handling logic was embedded in `.pipe` became complicated _very_ quickly. There's bookkeeping
that needs to be done with regards to default error handlers, deciding which error handler needs to propagate the PipelineErrorEvent, etc.,
and I wanted to keep that as separate from `.pipe` as possible. In order to do this, I added an internal event -- `"_preError"` -- that emits
the error and an object that can be modified by listeners of `_preError` to indicate whether the error is handled or not.

One benefit of this approach is that the feature can be added piecemeal -- adding the navigational components can be one piece, adding `_preError`
can be another, and we can iterate on what the public API for `_preError` should be (if the current approach is deemed unsatisfactory).
### Added `_errorHandlers` attribute to base Stream class

This is pretty straightforward. Like `_{next,prev}Streams`, it's an attribute that's
lazily added if not present; otherwise it contains a list of `{handler, uninstall}` 
objects. `{add,remove}PipelineErrorHandler` manipulates this attribute.
### Added `addPipelineErrorEvent`

This is the [real meat](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L81-L154) of the change. It installs listeners onto the current
stream as well as all streams returned by `prevAll()`. Installation adds
listeners for `unpipe`, `pipe`, `_preError`, and, if there are no `error`
handlers present, `error` as well.

The `error` event handler is called `_defaultPipelineErrorHandler` ([here](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L156-L171)). Because streams2+
do not add an error event handler to Readable streams on `.pipe`, this ensures consistent
pipeline error behavior -- namely, that errors emitted on Readables are forwarded down the
pipeline.

Because upstream may change over time (after the initial
`addPipelineErrorHandler` call), a listener is added for `pipe` and
`unpipe`. [Once piped to](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L121-L125), a pipeline-error-handling stream will remove the
`_defaultPipelineErrorHandler`, and install the 3 to 4 listeners on the new
source and all of its ancestors. [If unpiped](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L127-L130), to keep things simple, the handler
simply uninstalls and reinstalls itself, re-propagating the error handlers out
appropriately.

Uninstallation involves [removing all installed listeners](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L137-L140), and [splicing out](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L142-L147) the
errorHandler from the list of error handlers if the stream is the source
stream. Since the streams installed are tracked by the `installedOnStreams`
array (deduplicated with the aid of a `WeakSet`), uninstallation is as simple as
calling `uninstall` on [each of the streams](https://github.com/chrisdickinson/node/blob/feature/adderrorhandler/lib/stream.js#L150-L153).

### DIFF
--- a/doc/api/stream.markdown
+++ b/doc/api/stream.markdown
@@ -647,6 +647,168 @@ Examples of Transform streams include:
 * [crypto streams][]
 
 
+### Common Stream Methods
+
+**All** streams have the following methods, be they Readable, Writable, Duplex,
+Transform, PassThrough, or classic.
+
+#### stream.addPipelineErrorHandler(handlerFunction)
+
+Adds an error handler to the stream. This is *distinct* from adding an 
+"error" listener: all errors that occur upstream of this handler will be forwarded
+to this handler. This handler fires before the streams are unpiped, and has the 
+option of preventing the default error handling behavior. The handlerFunction is
+provided a single argument, a [PipelineErrorEvent][] object that provides the original
+error object and the stream the error originated on.
+
+```javascript
+
+var A = createReadableSomehow();
+var B = createTransform();
+var C = createWritable();
+
+A.name = 'A';
+B.name = 'B';
+C.name = 'C';
+
+C.addPipelineErrorHandler(function(errorEvent) {
+  console.log(errorEvent.stream.name, errorEvent.error);
+
+  // prevent the stream from unpiping or exploding if there are
+  // no "error" listeners.
+  errorEvent.handleError();
+})
+
+A.pipe(B).pipe(C);
+
+A.emit('error', new Error('problem!'));   // logs, "A problem!"
+B.emit('error', new Error('problem!'));   // logs, "B problem!"
+C.emit('error', new Error('problem!'));   // logs, "C problem!"
+```
+
+If the error is not handled on the current turn of the event loop, the streams
+will be unpiped and if there are no error listeners on the stream that
+generated the error, the error will be thrown. For example:
+
+```javascript
+A.pipe(B).pipe(C).addPipelineErrorHandler(function (ev) {
+  console.log(ev.error.message);
+});
+
+A.emit('error', new Error('beep boop')); // logs "beep boop", and then the process crashes
+```
+
+Error events will propagate downward from the stream that emitted them. Handling
+the error event will also stop its propagation downstream.
+
+```javascript
+
+A.pipe(B).addPipelineErrorHandler(function (ev) {
+    console.log('B');
+  })
+  .pipe(C).addPipelineErrorHandler(function (ev) {
+    console.log('C');
+    ev.handleError();
+  })
+  .pipe(D).addPipelineErrorHandler(function (ev) {
+    console.log('D')
+  });
+
+A.emit('error', new Error('hello!')); // logs "B" and "C", but *not* "D"
+```
+
+#### stream.removePipelineErrorHandler(handlerFunction)
+
+Remove an error handler from a stream.
+
+```javascript
+
+var handler = function (ev) {
+  ev.handleError();
+};
+aStream.addPipelineErrorHandler(handler);
+aStream.removePipelineErrorHandler(handler);
+```
+
+#### stream.next()
+
+Returns an array of streams that this stream is piped to currently.
+
+```javascript
+A.pipe(B).pipe(C);
+A.next(); // [B]
+A.pipe(X);
+A.next(); // [B, X]
+```
+
+#### stream.prev()
+
+Returns an array of streams that are piping to this stream.
+
+```javascript
+A.pipe(B).pipe(C);
+C.prev(); // [B]
+X.pipe(C);
+C.prev(); // [B, X]
+```
+
+#### stream.nextAll()
+
+Returns an array of streams representing all downstream streams from this stream,
+with no cycles or duplicates, not including the current stream.
+
+```javascript
+A.pipe(B).pipe(C).pipe(D);
+A.pipe(X).pipe(Y);
+
+A.nextAll() // [B, C, D, X, Y]
+
+U.pipe(V).pipe(U);
+U.nextAll() // [V]
+```
+
+#### stream.prevAll()
+
+Returns an array of streams representing all upstream streams from this stream,
+with no cycles or duplicates, not including the current stream.
+
+```javascript
+A.pipe(B).pipe(C).pipe(D);
+X.pipe(Y).pipe(D);
+
+D.prevAll() // [C, B, A, Y, X]
+```
+
+### Class: PipelineErrorEvent
+
+This is the object that is provided to pipeline error handlers. It
+contains details about the error -- the original error object, as well
+as the stream the error was initially emitted on. Additionally, within
+an error event handler, it can be used to stop propagation of the error
+event and prevent the default unpiping behavior. Users should never need
+to instantiate this class themselves.
+
+```javascript
+
+fs.createReadStream('does/not/exist')
+  .pipe(http.request('http://localhost:8124'))
+  .addPipelineErrorHandler(function (errorEvent) {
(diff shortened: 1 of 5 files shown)


jasnell:
@chrisdickinson ... any reason to keep this open still?

chrisdickinson:
Nope! Feel free to close this.

contra:
Just for reference - what was the resolution here?

stevemao:
What @dominictarr is the resolution?
> If the error doesn't break the stream, it's not really an error, not from the point of view of the stream.
From the point of view of the stream, it's data. It's like, there wen't enough chicken meals for all the passengers, and some people had to have fish. It's not like the wings fall off the plane, that would most certainly be a fatal error. So what if gulp just emitted a non-fatal-error event, that did not cause streams to be unpiped? it could be converted into a data event hand passed along the stream via the normal mechanism. if you add a parallel channel for errors, then now you have two streams, one made of normal matter, and a weird dark-matter stream that rarely interacts with the normal world.

PR #8312, opened 2014-09-03
Create less IO errors to improve efficiency of require (forwardport of #8189)

errendir:
In some environment (especially when coffescript is involved) module.js
will spend a lot of time looking up nonexisting files. This is caused by the
lengthy error creation. Those errors are then immediately caught.

This change makes the lookup much faster, by avoiding the unnecessary error
creation.

module::statPath now uses the new second parameter of fs.statSync function of the
fs.js. There is a similar change for the fs.stat, adding the third parameter.

This is the forwardport of https://github.com/joyent/node/pull/8189

### DIFF
--- a/doc/api/fs.markdown
+++ b/doc/api/fs.markdown
@@ -168,11 +168,14 @@ Only available on Mac OS X.
 
 Synchronous lchmod(2).
 
-## fs.stat(path, callback)
+## fs.stat(path, callback, [throwSafe])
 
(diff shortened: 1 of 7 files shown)


jasnell:
@errendir ... Unfortunately this seems to have stalled out. At this point, the PR will need to be updated and reopened against master on http://github.com/nodejs/node. Closing this here (which is not to say that improvements aren't needed, just that this can't land the way it is currently)

PR #8418, opened 2014-09-21
fs: deprecate exists() and existsSync()

cjihrig:
`fs.exists()` does not currently support Node's standard callback signature. This commit adds errback support while maintaining backwards compatibility. Closes #8369.


jasnell:
@cjihrig @trevnorris .... any reason to keep this open?

cjihrig:
No. People have asked for this to be deprecated. We deprecated it in io.js, and people asked for it to be un-deprecated.

PR #8439, opened 2014-09-24
Handle trailing bytes for chunked encoding more gracefully

tjfontaine:
http requests may encounter a response stream that outputs extra bytes -- for instance see 1fddc1fee84785d296e37337dcb0dc12c13a049f where Node.js itself was sending too much. In that case our parser will currently give a parse error, which results in us unilaterally destroying the socket and raising an error on the `request` (not the `response`).

This change adds an event `chunkedRemainingBytes` which will be emitted on the `response` object, that callback will be passed the bytes that we were unable to parse. If there is no listener for this event we will fall back to the previous behavior. The documentation includes some mediation for detecting this case, but it cannot interrupt the socket from being destroyed.

As part of this PR, I also have a commit that partially reverts 1fddc1fee84785d296e37337dcb0dc12c13a049f, it was opting out of the hot path detection code for responses that have a body, resulting in the many pipeline test taking too long and timing out particularly on linux.

/cc @indutny @trevnorris

### DIFF
--- a/doc/api/http.markdown
+++ b/doc/api/http.markdown
@@ -246,6 +246,30 @@ does not imply that the client has received anything yet.
 
 After this event, no more events will be emitted on the response object.
 
+### Event: 'chunkedRemainingBytes'
+
+`function (buff) { }`
+
+Fired when the incoming stream is `Transfer-Encoding: chunked` and the stream
+has been parsed successfully but there are trailing bytes at the end of the
+response. The callback is passed the bytes that were not successfully parsed.
+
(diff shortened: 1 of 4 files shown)


jasnell:
@trevnorris @indutny @chrisdickinson ... any further thoughts on this one?

trevnorris:
Not from me.

jasnell:
Ok, closing. There's no pressing reason to land this in v0.10. If it's still something that needs to be done in master, then let's get a new PR opened there.

PR #8447, opened 2014-09-25
Support windows sdk on Windows 8.1 x64

WaffleSouffle:
Update configure script to improve check for Windows SDK by ignoring environment variable name case, and checking in both bin and bin\x86 directories.

### DIFF
--- a/configure
+++ b/configure
@@ -2,7 +2,6 @@
 import optparse
 import os
 import pprint
-import re
 import shlex
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


jasnell:
@orangemocha

jasnell:
@WaffleSouffle ... at this point, because of the transition to nodejs/node, this is likely better done in http://github.com/nodejs/build. Closing this here because it cannot land here.

PR #8452, opened 2014-09-25
url: do not stringify query if this.search exist

dead-horse:
it is unnecessary to stringify `this.query` when `this.search` exist.

what's worse, `querystring.stringify` will throw when:

```
var query = 'q=%CE%ED%BB%AF%C6%F7';
var parsed = qs.parse(query);
qs.stringify(parsed); // URIError: URI malformed throwed
```

so `url.parse` will throw when `url.parse('/index?q=%CE%ED%BB%AF%C6%F7', true)`.

BTW, can we fix `querystring.stringify` ?

### DIFF
--- a/lib/url.js
+++ b/lib/url.js
@@ -392,7 +392,8 @@ Url.prototype.format = function() {
     }
   }
 
-  if (this.query &&
+  if (!this.search &&
+      this.query &&
       util.isObject(this.query) &&
       Object.keys(this.query).length) {
     query = querystring.stringify(this.query);


jasnell:
At this point, given that v0.12 and v0.10 APIs are locked down and the discussion that's happened in io.js around changing these kinds of behaviors, I'm inclined to close this without further action. Any changes on this would need to be done in http://github.com/nodejs/node master. Can reopen if necessary. /cc @joyent/node-tsc

PR #8459, opened 2014-09-26
Fixed url encoding for slash switching emulation.

eborden:
Fixes #8458


GUI:
Just a heads up for anyone that might run into this and find this issue: As mentioned earlier, this affects v0.11.14, but this also actually affects Node v0.10 as of the v0.10.27 release (commit https://github.com/joyent/node/commit/9520adeb37f5ebe02a68669ec97770f4869705bb).

Issue #8324, opened 2014-09-06
please mention AtExit in the documentation


kapouer:
Hi,
it would be nice to have [AtExit](https://github.com/joyent/node/commit/e4a8d2617b49727c4de09ad9ca9bc8fa270adf1e) mentioned in
https://github.com/joyent/node/blob/master/doc/api/addons.markdown

It would save developers a lot of time ;|

jasnell:
Closing since the PR was closed and never landed.

PR #8461, opened 2014-09-28
Update addons.markdown

ghost:
Updates addons.markdown to mention AtExit() function.  Addresses issue #8324.


JCMais:
What is the status here? Why it was not merged?

trevnorris:
IIRC it was because it `make test-addon` failed with this patch, but no one from core closed this issue. We were simply waiting for the patch to be fixed.

PR #8479, opened 2014-10-01
Added missing StringDecoder to readline.js

ISNIT0:


jasnell:
@ISNIT0 ... this change looks fine in general but, as @indutny and @trevnorris indicate, it's not clear what this is fixing and a test would be required. If this is something that you'd like to pursue, opening a new PR against master in http://github.com/nodejs/node would be the right approach. Going to close this for now as it's not something that will be able to land here in it's current state.

PR #8485, opened 2014-10-02
build: fix or suppress warnings

refack:
- activated "treat warnings as errors"


jasnell:
@orangemocha ... do you see reason to keep this one open?

orangemocha:
Well, now we do have a CI that tests every commit. It would be nice to flush all warnings out and treat warnings as errors. This PR would have to be revisited though, and it would make more sense to do it in https://github.com/nodejs/node

PR #8486, opened 2014-10-02
build: added `verbose` argument to build tools

refack:


jasnell:
@orangemocha ... any reason to keep this one open?

PR #8487, opened 2014-10-02
tools: update cpplint.py

refack:
- filter out new errors for now


jasnell:
@indutny @orangemocha @misterdjules ... do any of you see a reason to keep this one open?

PR #8491, opened 2014-10-02
doc: update CONTRIBUTING.md

diosney:
Update the procedures by using the source at the Node.js website, since they
clear some doubts on source/destination branches for contributions.

Node.js source @ http://www.nodejs.org/documentation/contributing/


jasnell:
No longer relevant.

PR #8507, opened 2014-10-03
include error argument in thrown Error msg

aheckmann:
when a non-Error is emitted, include it in the constructed Error message.

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -84,7 +84,7 @@ EventEmitter.prototype.emit = function emit(type) {
     } else if (er instanceof Error) {
       throw er; // Unhandled 'error' event
     } else {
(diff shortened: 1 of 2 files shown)


jasnell:
While simple, I don't think this is something we'd land now in v0.12 or v0.10. @aheckmann, I know it's been a while, but if this is something that you'd like to pursue, it's likely best to close this PR, update, and open a new PR against the active development branch at http://github.com/nodejs/node

jasnell:
A more useful approach would be to set the exception as a property on the new Error object then possibly. e.g. 
```
var err = Error('Uncaught, unspecified "error" event');
err.exception = er;
throw err;
```
Going ahead and closing as this is not a change that we'll be able to land here.

PR #8522, opened 2014-10-06
fs: Add file descriptor support to *File() funcs

jwueller:
These changes affect the following functions and their synchronous counterparts:
- `fs.readFile()`
- `fs.writeFile()`
- `fs.appendFile()`

If the first parameter is a `uint >= 0`, it is treated as a file descriptor. In all other cases, the original implementation is used to ensure backwards compatibility. File descriptor ownership is never taken from the user.

The documentation was adjusted to reflect these API changes. A note was added to make the user aware of file descriptor ownership and the conditions under which a file descriptor can be used by each of these functions.

Tests were extended to test for file descriptor parameters under the conditions noted in the relevant documentation.

### DIFF
--- a/doc/api/fs.markdown
+++ b/doc/api/fs.markdown
@@ -460,9 +460,9 @@ The callback is given the three arguments, `(err, bytesRead, buffer)`.
 
 Synchronous version of `fs.read`. Returns the number of `bytesRead`.
 
-## fs.readFile(filename[, options], callback)
+## fs.readFile(file[, options], callback)
 
-* `filename` {String}
+* `file` {String | Integer} filename or file descriptor
 * `options` {Object}
   * `encoding` {String | Null} default = `null`
   * `flag` {String} default = `'r'`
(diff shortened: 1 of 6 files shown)


doc/api/fs.markdown
```diff
@@ -527,7 +535,10 @@ Example:
       console.log('The "data to append" was appended to file!');
     });
 
-## fs.appendFileSync(filename, data, [options])
+Note that any specified file descriptor needs to be openend for appending.
```
misterdjules: Typo: `openend` should be `opened`.

test/simple/test-fs-write-file.js
```diff
@@ -97,12 +97,46 @@ fs.writeFile(filename3, n, { mode: m }, function(e) {
...
+var buf = new Buffer(s, 'utf8');
+common.error('writing to ' + filename4);
+
+fs.open(filename4, 'w+', function(e, fd) {
+  if (e) throw e;
+
+  ncallbacks++;
+  common.error('file4 opened');
+
+  fs.writeFile(fd, buf, function(e) {
```
misterdjules: It seems that we're missing a test for `fs.writeFileSync(fd, ...)`, could you please add one? Thanks!

misterdjules:
@jwueller @trevnorris What do we need to do to move this PR forward?

trevnorris:
@misterdjules probably a rebase. Last I looked it was fine. Once it's rebased I'll take another glance and sign off.

jwueller:
@trevnorris Oh my, I just noticed that this PR is still open and pushed a rebase. Do we need anything else?

jwueller:
@trevnorris @misterdjules Should I add a test for `fs.writeFileSync(fd, ...)` in this PR, or do you want that one separately?

jasnell:
@jwueller ... well, thank you for being so patient on this one. At this point, given the API change, this is something that would need to be closed here then re-targeted at master with a new PR on http://github.com/nodejs/node.

jwueller:
@jasnell I'll do that, thank you for the update.

PR #8564, opened 2014-10-17
net: Allow custom DNS resolvers

### LINKED ISSUE #8475: net: Allow custom DNS resolvers
Per @bnoordhuis's suggestion in #2868:

> [...] a change that makes the DNS resolver configurable in the call to net.connect() might be acceptable. [...]
> The API I have in mind is something like this:
>
> ```javascript
> net.connect({
>  host: "nodejs.org",
>  port: 80,
>  resolver: function(host, type, callback) {
>    return dns.resolve(host, type, callback);
>  }
> });
> ```
>
> It's up for debate if the 'type' argument should be called 'family' instead and be a number, like how dns.lookup() works. I'm open to well-reasoned suggestions.

The PR should be against master -- this would be a good first feature for someone looking to contribute (be sure to include tests, doc updates, and to follow the [guidelines for contributing](http://nodejs.org/documentation/contributing/)).

cdlewis:
net.connect now checks whether the user has provided a dns
resolution function, this arguments and expected output
for which mirror the existing dns.lookup implementation.

Fixes #8475

### DIFF
--- a/doc/api/net.markdown
+++ b/doc/api/net.markdown
@@ -70,7 +70,9 @@ For TCP sockets, `options` argument should be an object which specifies:
 
   - `localPort`: Local port to bind to for network connections.
 
-  - `family` : Version of IP stack. Defaults to `4`.
+  - `family`: Version of IP stack. Defaults to `4`.
+
+  - `resolver`: Function to resolve domain names. Defaults to `dns.lookup`.
 
 For local domain sockets, `options` argument should be an object which
 specifies:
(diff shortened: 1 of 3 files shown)


jasnell:
Unfortunately this appears to have stalled out. At this point this is something that would likely need to be closed here and reopened with a new PR at http://github.com/nodejs/node. @cjihrig ... any further thoughts?

cjihrig:
I think we can close this, based on the resolution of #8475

PR #8579, opened 2014-10-19
buffer: implement `iterable` interface

vkurchatkin:
This makes possible to use `for..of` loop with buffers. Also related `keys`, `values` and `entries` methods are added for feature parity with `Uint8Array`.

This is basically the code from v8 array iterator implementation minus spec stuff.


olalonde:
Any plans to merge this?

vkurchatkin:
This is already merged in io.js

Issue #8578, opened 2014-10-18
cipher default key derivation is very weak


calvinmetcalf:
keys and ivs for createCipher using a very weak key derivation formula making brute force password guessing very easy.  Currently node uses EVP_BytesToKey with md5, no salt and one iteration, which means that the formula to derive a 128bit key and iv is:

``` js
key: md5(password)
iv: md5(md5(password) + password)
```

the problem is that is is very easy to do a large number of password guesses (or to precompute them before hand and save them). A random salt would prevent precomputing the list and more rounds and a secure harder would make guesses more costly. 

Currently you can do this yourself using the createCipheriv function but there is no indication in the docs why you might want to do this.

I guess my first question is, am I over blowing this or missing some very good reason for it being like it is?

If not I'd be happy to open a pull updating the docs to explain the default key derivation function and how to generate your own keys from a password as a first step.

PR #8580, opened 2014-10-19
doc: add note about key derivation

calvinmetcalf:
adds a note to the crypto docs passing along
the advice that openssl gives about what
key derivation function they recommend.
#8578


jasnell:
Appears to have already landed in io.js (or at least some derivation of it)

Issue #3770, opened 2012-07-25
Inconsistent url object handling in http.request and url.format


CrabBot:
Related to #1390.

If you `url.parse` a url, alter the pathname, and then pass it to `http.request`, path takes precedence. However, if you first `url.format` the url, pathname takes precedence:

``` javascript
var uri = url.parse('http://www.google.com/search?q=asdf');
uri.pathname += '/naaaaah';

var r1 = http.get(uri, function(res) {
    if (r2.path && r1.path !== r2.path) throw new Error('Mismatch');
});

var r2 = http.get(url.parse(url.format(uri)), function(res) {
    if (r1.path && r1.path !== r2.path) throw new Error('Mismatch');
});
```

Obviously, this is because http.request does not consider pathname. However, this is inconsistent then with the `url.parse`'s partial compliance regarding `hostname`:

> hostname: To support url.parse() hostname is preferred over host

and the options string:

> options can be an object or a string. If options is a string, it is automatically parsed with url.parse().

See [documentation](http://nodejs.org/docs/latest/api/all.html#all_http_request_options_callback).

jasnell:
Closing the issue here in favor of the ongoing work on the url module in io.js (https://github.com/nodejs/io.js/pull/1650)

PR #8635, opened 2014-10-28
Allow url.format to take the 'path' parameter

lxe:
I know there are already a few tickets relating to the url parse/format inconsistency (https://github.com/joyent/node/issues/3770), but I'm wondering if adding support for 'path', (pathname + search) makes sense. I don't believe this will create any stability issues, as 'path' is not a part of currently allowed url.fromat parameters. What do you think?

### DIFF
--- a/lib/url.js
+++ b/lib/url.js
@@ -420,7 +420,9 @@ Url.prototype.format = function() {
   });
   search = search.replace('#', '%23');
 
-  return protocol + host + pathname + search + hash;
+  var path = this.path || pathname + search;
+
+  return protocol + host + path + hash;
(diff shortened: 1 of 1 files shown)


jasnell:
@lxe ... I know it's been a while on this one. If this is still something you'd like to pursue, the likely best course of action would be to close this here and open a new PR or issue against the active development stream at http://github.com/nodejs/node. This is not a change that would land here.

jasnell:
Closing this as it cannot land here. @lxe, as I mentioned, please open a new PR in nodejs/node if you'd like to pursue.

Issue #7676, opened 2014-05-23
Intl: Build "English-only" binaries for the distribution.


srl295:
A new issue to continue the discussion in #6371 and #4689  - especially to implement https://github.com/joyent/node/issues/6371#issuecomment-42491824 

* provide a way to build against any ICU, not just Chromium's
* default build has very little data (just English) for a small file size
* default disk location at build time for additional data `--with-default-icu-path=...`
* runtime option `--icu-data-dir=`path.. or NODE_ICU_DATA env var can override the default 
* ICU won't be in github, but can be enabled by default in builds.

/cc @tjfontaine @trevnorris

srl295:
- [ ] TODO:  [Document how to update node's timezone data](srl295/node/#9)

misterdjules:
I find that having this "general discussion" issue makes it difficult to track the current state of this work. 

If I understand correctly, here's what we want regarding the distribution of binaries that include ICU support:
* First, release binaries with "small" ICU support, the target for that is the 0.11.15 milestone. This is already supported, only Jenkins jobs must be updated to use `--with-intl=small-icu` and `vcbuild.bat small-icu`.
* When https://github.com/joyent/node/issues/8983 is implemented, include "full" ICU data in binary distributions (Windows and OS X installers, as well as binary tarball). This can be done in a later milestone, like 0.12.1, because it doesn't change node's API.

If we agree on that, I suggest closing this issue and creating two separate issues that correspond to the two points mentioned above.

@tjfontaine @srl295 @trevnorris What do you think?

srl295:
> First, release binaries with "small" ICU support, the target for that is
the 0.11.15 milestone. This is already supported, only Jenkins jobs must be
updated to use --with-intl=small-icu and vcbuild.bat small-icu.

But, that's the only and entire point of #7676 - if that's true,, it should
be considered fixed.

misterdjules:
@srl295 Ok, I was just confused by some comments [that mention several sub-issues/sub-tasks](https://github.com/joyent/node/issues/7676#issuecomment-64704230) that we may want to address separately. This issue also contains [comments that were not updated](https://github.com/joyent/node/issues/7676#issuecomment-63399139) and I think that just adds to the confusion.

Like I said I would rather have each issue/PR address only "atomic" issues so that we can track them more easily.

[...]

PR #8681, opened 2014-11-06
build: i18n: fix issue with icu toolset dependencies

srl295:
So what I did here is to make the `icutools` library
actually contain the entire ICU dependencies needed
for host-side tools. Sadly, this will build ICU twice,
but avoids conflicts between host and target side.
This all seems like a gyp bug of some sort, but without
docs for `toolsets`, who’s to say?

I removed the `icuio` library, because it was only
used by the host-side tools.

ref: https://github.com/joyent/node/issues/7676

### DIFF
--- a/tools/icu/icu-generic.gyp
+++ b/tools/icu/icu-generic.gyp
@@ -90,7 +90,7 @@
     {
       'target_name': 'icui18n',
       'type': '<(library)',
-      'toolsets': [ 'host', 'target' ],
+      'toolsets': [ 'target' ],
       'sources': [
         '<@(icu_src_i18n)'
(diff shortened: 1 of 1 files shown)


jelmd:
Just tried node-v5.0.0 with --with-intl=system-icu and I get the following error:

```
gyp: Dependency '/export/scratch/build/node-v5.0.0/tools/icu/icu-system.gyp:icui18n#host' not found while trying to load target /export/scratch/build/node-v5.0.0/deps/v8/tools/gyp/v8.gyp:v8_base#host
Error running GYP
```

Not sure, what host is all about. Any hints?

PR #8682, opened 2014-11-06
Add a note about readable in flowing mode

### LINKED ISSUE #8568: streams: data/pipe and readable can't be used together
`data/pipe` events and `readable` can't be used together, the latter is ignored.

E.g. in this example, there are many `data` events, but no `readable` events (except the last one), so only 1 dot is printed:
```
fileStream.on('data', function(data) { console.log(data); });
fileStream.on('readable', function() { console.log('.');  });
```

The same with pipe - only 1 dot even if there are many data chunks:
```
fileStream.pipe(outputStream);
fileStream.on('readable', function() { console.log('.');  });
```

It is quite visible in the node code why that happens, but I couldn't find any mentions in the docs that "data/pipe" and "readable" are antagonists.

Also, it is not obvious from the common sense standpoint why the examples above don't work "as intended".

Is there a reason for such decision? Anything in the docs about that?

P.S. Node 0.11.14, the stream is `fs.ReadStream`.

iliakan:
Closes #8568

### DIFF
--- a/doc/api/stream.markdown
+++ b/doc/api/stream.markdown
@@ -164,6 +164,8 @@ readable.on('readable', function() {
 Once the internal buffer is drained, a `readable` event will fire
 again when more data is available.
 
+The `readable` event is not emitted in the "flowing" mode with the sole exception of the last one, on end-of-stream.
+
 #### Event: 'data'
 
 * `chunk` {Buffer | String} The chunk of data.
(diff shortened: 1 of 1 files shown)


jasnell:
LGTM

jasnell:
@trevnorris ... unless you object, I can add this one to my batch doc update in https://github.com/joyent/node/pull/25565

trevnorris:
I'd suggest making sure it's word wrapped properly. But go right ahead. Good change.

jasnell:
Landed in b436e59

PR #8684, opened 2014-11-06
src: don't handle UTF16 external string

zcbenz:
The external string of V8 is in UTF16 encoding, force converting it to `char*` would always get wrong results.

### DIFF
--- a/src/string_bytes.cc
+++ b/src/string_bytes.cc
@@ -277,13 +277,6 @@ bool StringBytes::GetExternalParts(Isolate* isolate,
     *data = ext->data();
     *len = ext->length();
     return true;
-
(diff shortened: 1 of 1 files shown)


jasnell:
@trevnorris ... any further thoughts on this one?

zcbenz:
@jasnell I need to write some test cases for this PR, while I don't have time for it now.

jasnell:
Ok. Sounds good. No rush, just wanted to make sure it was still something
you wanted to pursue

trevnorris:
Should look at the impl in nodejs/node. Some fixes to `StringBytes` have made it in that, i don't believe ,have been backported.

PR #8693, opened 2014-11-07
Configure script supports WDK on Windows 8.1 64bit

WaffleSouffle:
Removed: unused import re.
Altered: "configure_winsdk()" handling case difference in
environment variables, and checking in both
Windows SDK bin and bin\x86 directories.

### DIFF
--- a/configure
+++ b/configure
@@ -2,7 +2,6 @@
 import optparse
 import os
 import pprint
-import re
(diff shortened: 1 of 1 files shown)


jasnell:
Looks like this is a duplicate. Closing. Can reopen if necessary

PR #8703, opened 2014-11-10
Prevent memory leak at getPeerCertificate when only the top certificate is requested

kevinmartin:


jasnell:
@indutny ... any reason to keep this one open?

indutny:
@jasnell nope

PR #8705, opened 2014-11-10
Gaurd against infinite memory leak in dgram

Raynos:
There is a unbounded array in the dgram library. We've seen this array have over 250 thousands items in production heap dumps.

We put about a huge amount of traffic through our statsd UDP socket. When we get DNS failures this unbounded queue buffers
all statsd data which causes a huge memory leak.

cc @tjfontaine @indutny

### DIFF
--- a/lib/dgram.js
+++ b/lib/dgram.js
@@ -311,7 +311,9 @@ Socket.prototype.send = function(buffer,
     return;
   }
 
-  self._handle.lookup(address, function(ex, ip) {
+  self._handle.lookup(address, dgramSocketSendLookupCallback);
+
+  function dgramSocketSendLookupCallback(ex, ip) {
     if (ex) {
       if (callback) callback(ex);
       self.emit('error', ex);
@@ -336,7 +338,7 @@ Socket.prototype.send = function(buffer,
         });
       }
     }
-  });
+  }
 };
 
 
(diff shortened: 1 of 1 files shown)


jasnell:
@Raynos ... still an issue for you? This would likely need to be updated and retargeted at master http://github.com/nodejs/node if you'd like to pursue.

Fishrock123:
@mcollina could you check to see if this might still be an issue?

cjihrig:
@Fishrock123 I believe I can artificially reproduce the problem on master:

```javascript
'use strict';
const dns = require('dns');

dns.lookup = function(address, family, callback) {
  callback(new Error('foo'));
};

const dgram = require('dgram');
const socket = dgram.createSocket('udp4');

socket.on('error', (err) => {});

setInterval(() => {
  for (let i = 0; i < 100000; ++i) {
    socket.send('foobar', 80, 'google.com');
  }

  console.log(`queue length: ${socket._queue ? socket._queue.length : 0}`);
}, 1000);
```

The queue grows on every `send()` until DNS lookup succeeds.

Raynos:
Sounds like it's still an issue.

At uber we only have one UDP use case and it has a manual gaurd ( https://github.com/uber/node-statsd-client/blob/master/lib/ephemeral-socket.js#L136-L150 ) in it work around this bug.

So, funny story. Because of the "nodejs dns bug" we've literally stopped using DNS in our uber data centers and only use IP addresses.

cjihrig:
Closed in https://github.com/nodejs/node/pull/11036.

PR #8711, opened 2014-11-11
crypto: allow creation of GCM ciphers with createCipher

calvinmetcalf:
Sets the authenticated encryption specific methods ([set|get]AuthTag and setAAD)
on the Cipher prototype not just the Cipheriv prototype.


jasnell:
Closing. Appears to have already been landed.

PR #8714, opened 2014-11-12
fs: add access() and accessSync()

cjihrig:
`fs.exists()` and `fs.existsSync()` do not follow typical node conventions. `access()` and `accessSync()` are added as alternatives in this commit.

@trevnorris I opened this as a separate PR because it was decided in #8418 that `exists()` and `existsSync()` would not be deprecated yet. One thought that I had is that these functions might seem less C-like if they returned `true` on success instead of the 0 returned by libuv. Thoughts?


timoxley:
bump.

trevnorris:
Thanks for the bump. Landed in 29449349da.

PR #8729, opened 2014-11-15
doc: warn of instant truncation of fs.writeFile()

### LINKED ISSUE #7807: fs.writeFile() causes previously created ReadStreams to fail.
Assume you have a file `foo` containing `1234`.

Here is a snippet of code:

``` js
var fs = require('fs');
var filename = './foo';

fs.createReadStream(filename).pipe(process.stdout);
fs.writeFile(filename, '1234\n');
fs.createReadStream(filename).pipe(process.stdout);
```

It outputs the following:

```
1234
```

Commenting the `fs.writeFile` line would make the program output the following:

```
1234
1234
```

… which is what we would expect.

Besides, moving the `fs.writeFile` line at the end of the program causes nothing to be output when the program is run, which makes me believe that `fs.writeFile()` corrupts previously created ReadStreams.

We would also get what we'd expect by replacing `fs.writeFile` with `fs.writeFileSync`.

Using `process.nextTick()` around the otherwise corrupted ReadStream also gives us what we'd expect:

``` js
var fs = require('fs');
var filename = './foo';

process.nextTick(function() {
  fs.createReadStream(filename).pipe(process.stdout);
});
fs.writeFile(filename, '1234\n');
fs.createReadStream(filename).pipe(process.stdout);
```

I am using stable v0.10.29.

espadrine:
Fixes #7807

### DIFF
--- a/doc/api/fs.markdown
+++ b/doc/api/fs.markdown
@@ -353,6 +353,10 @@ On Linux, positional writes don't work when the file is opened in append mode.
 The kernel ignores the position argument and always appends the data to
 the end of the file.
 
+Warning: when using the flag `'w'`, the file is truncated instantly, but
+it may only be written to many event loop ticks later. Any attempt
+to read from the file until then will return no data and no error.
+
 ## fs.openSync(path, flags, [mode])
 
 Synchronous version of `fs.open()`.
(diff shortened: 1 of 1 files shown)


jasnell:
@joyent/node-tsc ... closing this one due to lack of any further action. Can reopen if anyone feels it is necessary.

PR #8733, opened 2014-11-16
Add option to systematically activate HTTP 1.1 requests

valir:
When trying to get documents from an Alfresco server, I got 505
errors. I investigated this only to discover that node only emit
HTTP 1.1 requests under some particular conditions.

This commit add support for a new option named useHTTP11.
When this option is found, the HTTP client will specify HTTP 1.1
protocol version. This has fixed my alfresco connectivity problem.

Example:

var opts = {
    useHTTP11: true
};

httpClient.get("http://server/url", opts);

### DIFF
--- a/lib/_http_client.js
+++ b/lib/_http_client.js
@@ -128,7 +128,7 @@ function ClientRequest(options, cb) {
   if (util.isArray(options.headers)) {
     self._storeHeader(self.method + ' ' + self.path + ' HTTP/1.1\r\n',
                       options.headers);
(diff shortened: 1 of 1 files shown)


jasnell:
At this point, I'm inclined to close this here. It's not something that we would land in v0.12 or v0.10. @valir if this is something you'd like to pursue, closing this PR and opening a new one against the active development branch in http://github.com/nodejs/node would be the best approach.

jasnell:
As indicate, this is not a change that would land here. I encourage you to open a new PR if this is something you'd like to pursue.

PR #8734, opened 2014-11-17
assert: use util.inspect() to create error messages

### LINKED ISSUE #8696: assert.equal (and friends) throw incorrect error on circular refs
When `assert.equal` fails, I would expect to get an `AssertionError`. Instead, when one of the first two arguments has a circular reference in it, I get a `TypeError`. Here is a contrived example from a node#v0.10.33 repl:

```
> x = {}
{}
> x.x = x
{ x: [Circular] }
> assert.equal(x, {})
TypeError: Converting circular structure to JSON
    at Object.stringify (native)
    at getMessage (assert.js:75:24)
    at new AssertionError (assert.js:45:37)
    at fail (assert.js:92:9)
    at Function.equal (assert.js:121:27)
    at repl:1:9
    at REPLServer.self.eval (repl.js:110:21)
    at Interface.<anonymous> (repl.js:239:12)
    at Interface.emit (events.js:95:17)
    at Interface._onLine (readline.js:202:10)
```

Clearly this is happening because `JSON.stringify` fails on `x` with its circular references. Maybe something like `util.inspect` can be used instead, in order to replace circular references with `[Circular]`?

I've tested the following on both 0.10.33 and 0.11.14 and they all behave similarly for the `x` defined above.

```
assert.equal(x, {})
assert.equal({}, x)
assert.deepEqual(x, {})
assert.deepEqual({}, x)
assert.strictEqual(x, {})
assert.strictEqual({}, x)
```

cjihrig:
Currently, `JSON.stringify()` is used to create error messages on failed assertions. This causes an error when stringifying objects with circular references. This commit switches out `JSON.stringify()` for `util.inspect()`, which can handle circular references.

Closes #8696

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -56,19 +56,6 @@ assert.AssertionError = function AssertionError(options) {
 // assert.AssertionError instanceof Error
 util.inherits(assert.AssertionError, Error);
 
-function replacer(key, value) {
-  if (util.isUndefined(value)) {
-    return '' + value;
-  }
-  if (util.isNumber(value) && !isFinite(value)) {
-    return value.toString();
-  }
-  if (util.isFunction(value) || util.isRegExp(value)) {
-    return value.toString();
-  }
-  return value;
-}
-
 function truncate(s, n) {
   if (util.isString(s)) {
     return s.length < n ? s : s.slice(0, n);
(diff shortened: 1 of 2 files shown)


piscisaureus:
This is a good idea. +1

misterdjules:
LGTM, thanks @cjihrig!

misterdjules:
All tests pass (unsurprisingly), landed in bcff90e0c299ce472d3e00b8f886dac8e99478bd.

piscisaureus:
@cjihrig Would be good to land this in io.js as well.

cjihrig:
@piscisaureus this could be considered a breaking change. Thoughts on that?

TooTallNate:
I would agree

piscisaureus:
@cjihrig Semver-minor? The merge window for 1.1 is open.

cjihrig:
@piscisaureus I would say semver-major because perfectly valid programs could stop working if they were doing any string matching on the output. I'll open a PR over there and see what others say.

PR #8738, opened 2014-11-17
http: allow createServer() to return a HTTPS server

### LINKED ISSUE #8670: A single http.createServer([options], [listener]) interface for http/s
Here's a silly question: why do we have to require different submodules and call the same method with slightly different arguments when we can easily merge the two into a single call? Every place where http/https is an option, we repeat the same silliness of checking for TLS options and based on that requiring a different agent, etc.

Can we just enhance http.createServer() to accept options, and if those options include `tls`, to return the right server? This will be fully backwards compatible, but will make for much cleaner code elsewhere.

cjihrig:
This commit allows the `http` module to create a HTTPS server if TLS options are passed to `createServer()` or the `Server()` constructor.

As pointed out in #8670, this is a convenience that is backwards compatible. People generally seemed to be in favor of this change (myself included), so I decided to PR it. However, I can also see an argument for this as a userland module. I'll leave it up to the maintainers to decide.

Closes #8670.

### DIFF
--- a/doc/api/http.markdown
+++ b/doc/api/http.markdown
@@ -57,9 +57,11 @@ A collection of all the standard HTTP response status codes, and the
 short description of each.  For example, `http.STATUS_CODES[404] === 'Not
 Found'`.
 
-## http.createServer([requestListener])
+## http.createServer([options][, requestListener])
 
-Returns a new web server object.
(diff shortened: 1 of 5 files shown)


jasnell:
@cjihrig ... any reason to keep this open here?

cjihrig:
Nope.

PR #8758, opened 2014-11-21
util: Proposed fix for #8696, replacing circular references with "[Circular]"

nathanmacinnes:
JSON is used to create error messages when assertions fail. This causes a TypeError when there are circular references. This commit replaces circular references with the string "[Circular]"


jasnell:
@joyent/node-tsc ... see reason to keep this one open?

cjihrig:
I don't think so. This should be fixed by bcff90e0c299ce472d3e00b8f886dac8e99478bd

Issue #8759, opened 2014-11-21
child_process stderr.end() never invokes callback


timbertson:
The child_process docs[0](http://nodejs.org/api/child_process.html#child_process_child_stdin) say that child_process.spawn will return an object with a writable stream when using a pipe for stdin. But its `end()` method ignores its third argument (the callback)[1](http://nodejs.org/api/stream.html#stream_writable_end_chunk_encoding_callback). This means code expecting a new style stream will hang indefinitely waiting for the stream to actually end.

```
var child_process = require('child_process');

var proc = child_process.spawn('cat', [], {stdio:['pipe', 1, 2]});

proc.stdin.write('Hello!\n');
proc.stdin.end('(eof)', 'utf-8', function(rv) {
    console.log("child proc finished with:", rv);
});

setTimeout(function() {
    console.log("time out...");
}, 1000);
```

outputs:

```
Hello!
(eof)
time out...
```

(tested on nodejs 0.10.33)

I also tried killing the process after a second, to see if end() was waiting for the process death or something. But it didn't make any difference.

PR #8768, opened 2014-11-22
net: forward Socket#end cb argument to Duplex#end

chrisdickinson:
The callback parameter was being dropped when Socket.prototype.end
called Duplex.prototype.end. This remedies that situation.

This problem exists in both the v0.10 and v0.12 branches.

Fixes: https://github.com/joyent/node/issues/8759

R=@indutny

### DIFF
--- a/doc/api/net.markdown
+++ b/doc/api/net.markdown
@@ -354,13 +354,16 @@ buffer. Returns `false` if all or part of the data was queued in user memory.
 The optional `callback` parameter will be executed when the data is finally
 written out - this may not be immediately.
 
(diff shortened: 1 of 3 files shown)


jasnell:
@cjihrig @chrisdickinson ... any updates on this?

cjihrig:
This didn't go into 0.10 because it would have been a breaking change. Now that 0.12 is out, the same applies there. I'd say open it on nodejs/node.

jasnell:
@chrisdickinson ... going to close this here. If this is one you'd still like to see, let's get a new PR opened against nodejs/node master.

PR #8769, opened 2014-11-22
doc: supplement docs for Writable and Transform streams

vovacodes:
Slightly fix and supplement the documentation for Writable#_write and
Transform#_transform methods. Make it more consistent. Mention
passing 'buffer' as a encoding param in case chunk is a Buffer.


jasnell:
This landed over in nodejs/node already. While it could land in v0.10, there's no pressing issue now that v0.10 is moving into maintenance

PR #8784, opened 2014-11-26
debugger: use requireRepl() to load debugger repl

### LINKED ISSUE #8359: Debugger crashes when attempting autocomplete on yet-to-be-required object
Scenario: While debugging attempted to use util.inspect(). But instead of requiring into a var and then calling the inspect method, tried performing a `require('util').inspect(obj)`. Crash with below stack trace.

``` Javascript
> require('util').inspThere was an internal error in Node's debugger. Please report this bug.
Object.keys called on non-object
TypeError: Object.keys called on non-object
    at Function.keys (native)
    at REPLServer.complete (repl.js:449:23)
    at Interface.complete [as completer] (repl.js:152:10)
    at Interface._tabComplete (readline.js:359:8)
    at Interface._ttyWrite (readline.js:779:14)
    at ReadStream.onkeypress (readline.js:99:10)
    at ReadStream.emit (events.js:98:17)
    at emitKey (readline.js:1095:12)
    at ReadStream.onData (readline.js:840:14)
    at ReadStream.emit (events.js:95:17)
```

I'm on Ubuntu 12.04 

```
$ uname -srvmpio
Linux 3.8.0-34-generic #49~precise1-Ubuntu SMP Wed Nov 13 18:05:00 UTC 2013 x86_64 x86_64 x86_64 GNU/Linux

$ node
> process.versions
{ http_parser: '1.0',
  node: '0.10.31',
  v8: '3.14.5.9',
  ares: '1.9.0-DEV',
  uv: '0.10.28',
  zlib: '1.2.3',
  modules: '11',
  openssl: '1.0.1i' }
```

cjihrig:
Currently, the debugger uses `require('repl')` to setup the repl. However, `require.extensions` is not available yet, causing a crash on tab completion of `require('`. This commit uses the `module.requireRepl()` method to bootstrap the repl.

Also added a test for `require('` tab completion in the repl module. Creating a test for tab completion with the debugger running is proving to be a bit painful.

Closes #8359

### DIFF
--- a/test/simple/test-repl-tab-complete.js
+++ b/test/simple/test-repl-tab-complete.js
@@ -211,3 +211,18 @@ testMe.complete(' ', function(error, data) {
 testMe.complete('toSt', function(error, data) {
   assert.deepEqual(data, [['toString'], 'toSt']);
 });
+
+// Tab complete provides built in libs for require()
+putIn.run(['.clear']);
+
+testMe.complete('require(\'', function(error, data) {
+  assert.strictEqual(error, null);
+  repl._builtinLibs.forEach(function(lib) {
+    assert.notStrictEqual(data[0].indexOf(lib), -1, lib + ' not found');
+  });
+});
+
+testMe.complete('require(\'n', function(error, data) {
+  assert.strictEqual(error, null);
+  assert.deepEqual(data, [['net'], 'n']);
+});
--- a/lib/_debugger.js
+++ b/lib/_debugger.js
@@ -23,7 +23,8 @@ var util = require('util'),
     path = require('path'),
     net = require('net'),
     vm = require('vm'),
-    repl = require('repl'),
+    module = require('module'),
+    repl = module.requireRepl(),
     inherits = util.inherits,
     assert = require('assert'),
     spawn = require('child_process').spawn;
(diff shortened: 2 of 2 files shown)


test/simple/test-repl-tab-complete.js
```diff
@@ -211,3 +211,18 @@ testMe.complete(' ', function(error, data) {
 testMe.complete('toSt', function(error, data) {
   assert.deepEqual(data, [['toString'], 'toSt']);
 });
+
```
misterdjules: Do these newly added tests actually test the bug that this PR fixes? It seems these tests don’t involve the debugger, so they seem not to be regression tests for this issue, or am I missing something?

misterdjules: Ah ok, I hadn't seen the second part of the commit message, sorry.

misterdjules:
Before landing this change, I'd actually like to see if we can create a test that can run automatically within the context of the debugger, and that could thus be part of the tests suite.

cjihrig:
@misterdjules here is the problem I'm running into. The tab autocomplete only appears to be activated when stdout is a tty. When I spawn the child process with a tty stream as stdout, I can't capture the output to validate it (although I can see it, and it is correct). Any suggestions?

jasnell:
@cjihrig ... what would you like to do with this one?

cjihrig:
This also landed in the converged repo.

PR #8786, opened 2014-11-27
Remove the option of building against a shared cares

jbergstroem:
..since it won't build anyway. Should be reverted if/when upstream adapts a similar featureset.

Ping @indutny


jasnell:
This is not something that would land at this point. If it's something that's worth exploring further, a new PR against nodejs/node would be required.

PR #8789, opened 2014-11-27
v8: don't busy loop in cpu profiler thread

bnoordhuis:
Reduce the overhead of the CPU profiler by replacing sched_yield() with
nanosleep() in V8's tick event processor thread.  The former only yields
the CPU when there is another process scheduled on the same CPU.

Before this commit, the thread would effectively busy loop and consume
100% CPU time.  By forcing a one nanosecond sleep period rounded up to
the task scheduler's granularity (about 50 us on Linux), CPU usage for
the processor thread now hovers around 10-20% for a busy application.

Refs strongloop/strong-agent#3 and strongloop-internal/scrum-cs#37.

R=@trevnorris?

### DIFF
--- a/deps/v8/src/platform-freebsd.cc
+++ b/deps/v8/src/platform-freebsd.cc
@@ -539,11 +539,6 @@ void Thread::SetThreadLocal(LocalStorageKey key, void* value) {
 }
 
 
-void Thread::YieldCPU() {
-  sched_yield();
-}
-
-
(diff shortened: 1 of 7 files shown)


bnoordhuis:
@trevnorris Ping.

trevnorris:
@bnoordhuis Thanks for the bump. Landed in 6ebd85e1053.

PR #8791, opened 2014-11-27
src: reduce cpu profiler overhead

bnoordhuis:
Reduce the overhead of the CPU profiler by suppressing SIGPROF signals
when sleeping / polling for events.  Avoids unnecessary wakeups when
the CPU profiler is active.  Depends on libuv/libuv#15.

Refs strongloop/strong-agent#3 and strongloop-internal/scrum-cs#37.

See also #8789.

R=@trevnorris?

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -2962,6 +2962,7 @@ char** Init(int argc, char *argv[]) {
   V8::SetFlagsFromCommandLine(&v8argc, v8argv, false);
 
 #ifdef __POSIX__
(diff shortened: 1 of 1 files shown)


jasnell:
@trevnorris @bnoordhuis ... is there reason to keep this one open still?

bnoordhuis:
I'm going to close this.  I still think this is a good change but it needs an update to libuv v0.10.37 and the PR as-is breaks v8-profiler (see https://github.com/nodejs/node/pull/2324).

The fix would be trivial if v0.10 had the command line parsing infrastructure that v4 has but it doesn't and porting that over is too much work and risk.

PR #8805, opened 2014-12-01
QueryString optimization: moved try-catch UNTESTED

PlNG:
After reading and becoming aware of V8 Optimization Killers I was
reviewing the code to extract the URL module for my own uses when
I saw this performance killer - the try-catch in a loop.  The
statement seemed simple enough to move to its own function, so
I did just that and declared it outside of the loop.

Also, at worst case scenario, this module can be **two** try-catches
deep. This may be another area of performance improvement.

I have not tested this code. If someone more fluent in using node
could do a benchmark and verify my tweak, that would be great.

### DIFF
--- a/lib/querystring.js
+++ b/lib/querystring.js
@@ -187,6 +187,18 @@ QueryString.parse = QueryString.decode = function(qs, sep, eq, options) {
   if (options && typeof options.decodeURIComponent === 'function') {
     decode = options.decodeURIComponent;
   }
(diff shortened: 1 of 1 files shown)


jasnell:
@PING ... Unfortunately it looks like this stalled. If this is something that you would like to pursue, please go through @indutny's comments and open a new PR against master on http://github.com/nodejs/node.

PR #8813, opened 2014-12-03
Relative junction should behave like dangling symlink

vweevers:
A junction path must be absolute, which is why fs.symlink resolves paths on Windows if the type is `junction`. E.g. `fs.symlink("../foo", "my/link", "junction")`. From the [docs](http://nodejs.org/api/fs.html#fs_fs_symlink_srcpath_dstpath_type_callback):

> `fs.symlink(srcpath, dstpath, [type], callback)`
> 
> Note that Windows junction points require the destination path to be absolute. When using 'junction', the destination argument will automatically be **normalized to absolute path**.

_Sidenote_: Reading that, you'd think the `dstpath` argument is normalized. What the docs mean to say is that the `srcpath` (the _target_) is normalized. The function signature in the source is `fs.symlink(destination, path)`; these inconsistencies are confusing.

However, the target seems to be normalized relative to the working directory - not relative to the link's parent directory like it should:

> `man ln`
> 
> Symbolic links can hold arbitrary text; if later resolved, a relative link is interpreted in relation to its parent directory.

Of course, a junction is not the same as a symlink, but the behavior should be consistent, given that node on unix ignores the `junction` argument and will create a dangling symlink. It should do the same on Windows, or at least simulate it. This PR normalizes the target, relative to the link's parent directory. Meaning:

`fs.symlink("..\foo", "C:\my\link", "junction")`

becomes:

`fs.symlink("C:\foo", "C:\my\link", "junction")`

The included test is copied from `test-fs-symlink-dir-junction.js`. These tests could probably be merged, I didn't (yet) for the sake of clarity - and because the other (regular symlink) tests fail anyway on Windows XP.

### DIFF
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -707,12 +707,14 @@ fs.readlinkSync = function(path) {
   return binding.readlink(pathModule._makeLong(path));
 };
 
-function preprocessSymlinkDestination(path, type) {
+function preprocessSymlinkDestination(path, type, linkPath) {
   if (!isWindows) {
     // No preprocessing is needed on Unix.
     return path;
   } else if (type === 'junction') {
     // Junctions paths need to be absolute and \\?\-prefixed.
+    // A relative target is relative to the link's parent directory.
+    path = pathModule.resolve(linkPath, '..', path);
     return pathModule._makeLong(path);
   } else {
     // Windows symlinks don't tolerate forward slashes.
@@ -727,7 +729,7 @@ fs.symlink = function(destination, path, type_, callback) {
   if (!nullCheck(destination, callback)) return;
   if (!nullCheck(path, callback)) return;
 
-  binding.symlink(preprocessSymlinkDestination(destination, type),
+  binding.symlink(preprocessSymlinkDestination(destination, type, path),
                   pathModule._makeLong(path),
                   type,
                   callback);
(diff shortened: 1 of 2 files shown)


jasnell:
@piscisaureus ... looks like this never landed... any further thoughts on it?

piscisaureus:
@jasnell - it landed in io.js: https://github.com/nodejs/node/commit/764c5c7fd176c90513406797182cde1d0f24e1cf
I consider it a proper bug so from my perspective it could land in node 0.10/0.12 as well, but at the time I wasn't landing patches in joyent/node.

jasnell:
@vweevers @piscisaureus ... I know it's been a while on this one, but after apply this fix to the current `v0.10`, the included test case fails with...

```
=== release test-fs-symlink-dir-junction-relative ===                          
Path: simple/test-fs-symlink-dir-junction-relative
linkData: /Users/james/Node/main/node-joyent/test/fixtures/cycles/
linkPath: /Users/james/Node/main/node-joyent/test/tmp/cycles_link
relative target: ../fixtures/cycles
assert.js:93
        ^
AssertionError: "../fixtures/cycles" == "/Users/james/Node/main/node-joyent/test/fixtures/cycles/"
    at /Users/james/Node/main/node-joyent/test/simple/test-fs-symlink-dir-junction-relative.js:54:14
    at Object.oncomplete (fs.js:108:15)
Command: out/Release/node /Users/james/Node/main/node-joyent/test/simple/test-fs-symlink-dir-junction-relative.js
```

vweevers:
The tests were later improved, see [this](https://github.com/piscisaureus/io.js/commits/68ba608dc2b8809a4f3d94560658fada15468809/test/simple/test-fs-symlink-dir-junction-relative.js) and [this](https://github.com/nodejs/node/commits/master/test/parallel/test-fs-symlink-dir-junction-relative.js).

PR #8824, opened 2014-12-04
Fix for path.resolve across drives in Windows

### LINKED ISSUE #8818: path.resolve across drives in Windows
The path does not resolve correctly when resolving from one drive to another drive in windows using `path.resolve`

ankurp:
Fixes #8818

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -59,7 +59,7 @@ if (isWindows) {
   // Regex to split a windows path into three parts: [*, device, slash,
   // tail] windows-only
   var splitDeviceRe =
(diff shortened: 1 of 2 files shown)


ankurp:
Good to merge? @misterdjules

cjihrig:
Closing this for the same reason @piscisaureus closed https://github.com/iojs/io.js/pull/57

> `..\..\c:\etc\bla` is always in invalid path.
If a path operation turns an invalid path into a valid path, that's a disaster or security issue waiting to happen.
I think "garbage in, garbage out" is wildly more preferrable.

PR #8828, opened 2014-12-05
Feature: Improve add-on init dispatch

agnat:
Rebased #8806 on v0.12 branch.

> Handle different add-on `init()` signatures in a generic and type safe way. An adapter is selected at compile-time based on the signature of the init function. The adapter is stored in the module struct. At run-time the adapter unpacks the users init function and provides the requested arguments.
> 
> This is related to #8802.


agnat:
Bitrot. Taking it over to https://github.com/nodejs/node/pull/78

PR #8838, opened 2014-12-08
windows: Create npm folder in AppData directory

### LINKED ISSUE #8141: Windows installer doesn't ensure that $USERNAME\AppData\Roaming exists
npm uses `$APPDATA\npm` so that non-Administrator roles have a writable place to stash global packages. Since creating that directory if it doesn't exist _does_ require Administrator privileges, `npm -g install <whatever>` won't / can't work if the base directory doesn't exist (with the correct ACL) before npm is run. The installer should ensure that the whole path is set up at boot time.

Also, either the _documentation_ needs to change, or maybe the _installer_, because right now, people trying to upgrade npm using npm after they've used the .msi frequently end up confused because `npm -g install npm@latest` doesn't actually upgrade them to the newest version of npm, because the first npm on their path is the one that was originally installed alongside node in the Administrator directory. Either the positions of `AppData\roaming\npm` and the Node binary directory need to be swapped in `PATH`, npm should be installed to `AppData\roaming\npm` in the first place, or the documentation should be clarified to tell Windows users exactly how to upgrade npm with the .msi-installed Node.

rockarts:
Create the empty npm folder in Roaming\Appdata so that
non-Administrator users have a place to store global packages.
This fixes the error Error: ENOENT, stat error that occurs
when a user tries to run the npm install <package> command.

Fixes #8141

### DIFF
--- a/tools/msvs/msi/product.wxs
+++ b/tools/msvs/msi/product.wxs
@@ -70,6 +70,7 @@
       <ComponentRef Id="NpmCmdScript"/>
       <ComponentRef Id="NpmBashScript"/>
       <ComponentRef Id="NpmConfigurationFile"/>
+      <ComponentRef Id="AppData" />
       <ComponentGroupRef Id="NpmSourceFiles"/>
     </Feature>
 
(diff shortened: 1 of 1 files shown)


piscisaureus:
lgtm code-wise.
Fixed indentation errors and squashed; https://github.com/iojs/io.js/commit/e431cae7e70069cd1631081f9dca09990b948feb

jfroffice:
+1 still not working till node v0.10.32 :8ball:

piscisaureus:
Landed in 18d413d (v0.10) and 4ffbce4 (v0.12).

PR #8840, opened 2014-12-09
docs: add clarity to exit code 2

netpoetica:
Provide more detailed information to exit code 2, bash error. This is 
in regards to issue 7677, and mimics bash documentation available at 
http://tldp.org/LDP/abs/html/exitcodes.html


jasnell:
agree with @sam-github on this one. Closing this without further action. If there's any reason to pursue this further, opening a new PR or issue against http://github.com/nodejs/node would be the appropriate action.

PR #8848, opened 2014-12-09
docs: document input encoding in hmac

DanielSidhion:
Document the input_encoding parameter in the update function of the
Hmac class. This parameter is similar to the update function of the
Hash class, so I simply replicated the explanation.


jasnell:
@indutny ... just wanted to double check with you on this one. Any issues with it?

PR #8861, opened 2014-12-11
build: change `graphviz.py` to support "all"

srl295:
With this change, the following:

```
python tools/gyp_node.py -f dump_dependency_json
python tools/gyp/tools/graphviz.py > node_deps.dot
```

.. will produce node_deps.dot with all deps.  Previously, `graphviz.dot` needed exact targets, including pathnames.

I know this is a change to upstream, just putting it here for reference.

Sample output [here](https://drive.google.com/file/d/0B1J55jsbiIIpdE5lUFVzbTdHRXc/view?usp=sharing)

### DIFF
--- a/tools/gyp/tools/graphviz.py
+++ b/tools/gyp/tools/graphviz.py
@@ -27,6 +27,10 @@ def LoadEdges(filename, targets):
   edges = json.load(file)
   file.close()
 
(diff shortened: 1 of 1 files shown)


jasnell:
@srl295 ... what do you want to do with this one? Still needed?

PR #8868, opened 2014-12-13
doc: fix Folders as Modules omission of index.json

es128:
This section was inconsistent with the actual behavior of `require` and
with other parts of this same document, which do describe the ability
to treat index.json as a module’s default file.

PR inspired by the discussion at https://github.com/michaelficarra/us-states/pull/3


jasnell:
I'm rather surprised this was never looked at. I notice the doc problem still exists in v0.12 and nodejs/node master too. LGTM

jasnell:
Landed in 62c8948

Issue #8730, opened 2014-11-15
function becomes non-function when optimized by v8


briancavalier:
This issue was originally reported to v8 here: https://code.google.com/p/v8/issues/detail?id=3692

I'm filing it here so that it's findable via the node issue tracker as well.  I realize that it's unlikely to be fixed in the 0.10.x line, and please feel free to close if that's the case.  However, since this has already affected production systems (see References below) I think it's important for users to know that is the case, and to be able to find this information.

---

This test shows what we believe is node 0.10.33 + v8 (seems to be 3.14 based on the [changelog](https://raw.githubusercontent.com/joyent/node/master/ChangeLog)) generating invalid optimized code, causing a function to suddenly become an object.  This appears to affect when.js versions 3.4.2 and higher, although the failure appears to happen under different conditions depending on the version of when.js and/or the version of Node and v8.

Here's an executable showing the problem, and a more complete description of the issue, expected and actual results, etc.:

https://github.com/briancavalier/when-v8-invalid-optimization
### Environment

The test in the repo above is design to show the failure on:
- Node 0.10.33 (v8 3.14?)
- when.js 3.6.0
### Expected result

Program completes without error.
### Actual result

Program fails at a seemingly random, but reproducible iteration.

The stack trace below points to this line in the `_beget` function: https://github.com/cujojs/when/blob/3.6.0/lib/makePromise.js#L166).

```
<lots more output above>
#57 reading file with promise: `dist/lodash.compat.js`
#57 parsing file: `dist/lodash.compat.js`
**
 * @license
 * Lo-Dash 2.4.1 (Custom Build) <http://lodash.com/>
 * Build: `lodash -o ./dist/loda Program
registering catch # 393
Potentially unhandled rejection [1] TypeError: object is not a function
    at Promise._beget (/private/tmp/345/node_modules/when/lib/makePromise.js:166:16)
    at Promise.then (/private/tmp/345/node_modules/when/lib/makePromise.js:140:17)
    at Promise.catch (/private/tmp/345/node_modules/when/lib/makePromise.js:156:16)
    at Promise.catch.Promise.otherwise (/private/tmp/345/node_modules/when/lib/decorators/flow.js:37:22)
    at /private/tmp/345/test.js:87:24
    at next (/private/tmp/345/node_modules/when/lib/decorators/iterate.js:57:20)
    at /private/tmp/345/node_modules/when/lib/decorators/array.js:39:24
[...]

PR #8871, opened 2014-12-15
deps: fix deoptimization for CallNew/arguments

indutny:
Combination of CallNew and ArgumentElements instructions may lead to
deoptimizer OSR loading incorrect stack slots for pushed arguments.

Latest v8 seems to mitigate this by pushing all arguments late in one
instructions. Do the same thing but with multiple instructions placed
closer to the CallNew.

Fix joyent/node#8730

cc @tjfontaine @misterdjules


jasnell:
@indutny ... at this point do you see reason to land this in v0.10?

indutny:
I'd rather not do it anymore.

PR #8884, opened 2014-12-16
net: throw on invalid socket timeouts

cjihrig:
This commit restricts socket timeouts non-negative, finite numbers. Any other value throws a `TypeError` or `RangeError`. This prevents subtle bugs that can happen due to type coercion.

@misterdjules @tjfontaine


tjfontaine:
LGTM as well

cjihrig:
Landed in f34757398fcc393685b4dfbcbdc692fb38332d6c

PR #8925, opened 2014-12-22
Add support for 'prefer' and 'preference-applied' (RFC 7240)

jasnell:
RFC 7240 defines the Prefer and Preference-Applied header fields. Each can appear multiple times within an HTTP request/response (similar to Link). This adds basic support for Prefer/Preference-Applied and adds Link and Prefer to the relevant test case.


jasnell:
Closing to rebase against v0.10.35-release

PR #8927, opened 2014-12-22
http.OutgoingMessage.prototype.setHeaders

jasnell:
Small suggested api addition to http.OutgoingMessage ... add a
setHeaders to OutgoingMessage that allows multiple headers to be set at
once, similar to addTrailers.

Example:

``` javascript
response.setHeaders({a:'b', c:'c'});
```

Also, add documentation to addTrailers that illustrates how setting
headers using an array is supposed to work.

### DIFF
--- a/doc/api/http.markdown
+++ b/doc/api/http.markdown
@@ -324,6 +324,28 @@ or
 
 Boolean (read-only). True if headers were sent, false otherwise.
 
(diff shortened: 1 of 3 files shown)


jasnell:
Not going to land here. Will reopen on nodejs/node if necessary

PR #8933, opened 2014-12-23
test: use executable in spawn() test

### LINKED ISSUE #8930: Backport fix in test/simple/test-child-process-spawn-typeerror.js from v0.12 to v0.10
#8928 fixes an issue that was exposed by a merge from v0.10 to v0.12. However, [the same issue is present in v0.10 too](https://github.com/joyent/node/pull/8928#issuecomment-67902951).

Once [the merge from v0.10 to v0.12](https://github.com/joyent/node/pull/8886) is done, we should backport #8928 to v0.10.

The Right Way to do it would be to fix it in v0.10 first and then merge it into v0.12, but in this case a merge from v0.10 to v0.12 is already in progress, and we're close to be done with it, so we don't want to add more noise and complication.

cjihrig:
Currently, the `test-child-process-spawn-typeerror.js` is calling `execFile()` on a JavaScript source file, which is causing failures on Windows. This commit switches to calling `spawn()` on an actual executable.

Closes #8930
cc: @misterdjules

### DIFF
--- a/test/simple/test-child-process-spawn-typeerror.js
+++ b/test/simple/test-child-process-spawn-typeerror.js
@@ -41,10 +41,10 @@ assert.doesNotThrow(function() { spawn(cmd, a, o); });
 assert.doesNotThrow(function() { spawn(cmd, o); });
 
 // Variants of undefined as explicit 'no argument' at a position
-assert.doesNotThrow(function() { execFile(empty, u, o); });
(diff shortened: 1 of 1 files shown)


misterdjules:
Landed in 0a430b9aecf6a82fba5d76a41dba9b5279e8ce89, thanks!

PR #8934, opened 2014-12-23
Don't touch drive letter case on windows

piscisaureus:
See the two commit messages for motivation.

This has already landed in io.js, but I'd really like to see it landed before node v0.12 drops too.
Because we can have nice things.

@tjfontaine @orangemocha

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -153,12 +153,6 @@ win32.resolve = function() {
   resolvedTail = normalizeArray(resolvedTail.split(/[\\\/]+/),
                                 !resolvedAbsolute).join('\\');
 
-  // If device is a drive letter, we'll normalize to lower case.
-  if (resolvedDevice && resolvedDevice.charAt(1) === ':') {
-    resolvedDevice = resolvedDevice[0].toLowerCase() +
-        resolvedDevice.substr(1);
-  }
-
   return (resolvedDevice + (resolvedAbsolute ? '\\' : '') + resolvedTail) ||
          '.';
 };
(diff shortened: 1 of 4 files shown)


misterdjules:
Adding it to the 0.11.15 milestone because even though it's not an API breaking change, landing it post 0.12 would in my opinion change such a basic behavior on Windows that it would break backward compatibility.

misterdjules:
https://github.com/piscisaureus/node/commit/f09e321871d4c858c44f55e4964fe0060f07f5ac LGTM except for the missing newline in `test/simple/test-module-nodemodulepaths.js`.

However, as @orangemocha mentioned, I'm not sure I understand in which case https://github.com/piscisaureus/node/commit/9877cb4412934101c50bc5cbbbc2c763443919ae would be needed.

From my understanding of `require.resolve` and how `__dirname` is set, none of them alter the case of the "main" script passed on the command line. Moreover, the test passes for me on Windows regardless of the case used. So if I understand correctly how they work, I think https://github.com/piscisaureus/node/commit/9877cb4412934101c50bc5cbbbc2c763443919ae would actually make the test hide some potential issues, because it would always lower case both outputs, which could actually be different with regard to their case.

Thus, I suggest we land piscisaureus@f09e321 with the newline fix, and wait for more information on why https://github.com/piscisaureus/node/commit/9877cb4412934101c50bc5cbbbc2c763443919ae would be needed before maybe landing it.

misterdjules:
Landed piscisaureus@f09e321 in 016e08458cc38d7324c907c6594f2b56a5a19719 with the newline fixed as discussed in https://github.com/joyent/node/pull/8934#issuecomment-70178456. 

I did not land piscisaureus@9877cb4, but I'll be happy to do so when I understand it better. Closing though as in my opinion it fixes the original issue as it landed.

@piscisaureus @orangemocha  Thank you very much!

orangemocha:
@misterdjules I second your analysis on the test changes being unnecessary and potentially hiding other issues. Thanks for all the help!

Issue #8920, opened 2014-12-22
fs.symlink casues strange errors - EISDIR, EPERM, EEXIST


crackcomm:
* `README.md` exists
* `tmp` is empty directory
* at the end no symlink was created
* `fs.symlink` and `fs.symlinkSync` are affected
* node is running with Administrator rights
* `Windows 7`
* `node v0.10.x`, `node v0.11.14`
* works fine with MinGW

```JavaScript
> fs.symlinkSync('README.md', 'tmp/README.md', 'file')
Error: EISDIR, illegal operation on a directory 'D:\cc\ui-7\tmp\README.md'
    at Error (native)
    at Object.fs.symlinkSync (fs.js:767:18)
    at repl:1:4
    at REPLServer.defaultEval (repl.js:133:27)
    at bound (domain.js:254:14)
    at REPLServer.runBound [as eval] (domain.js:267:12)
    at REPLServer.<anonymous> (repl.js:280:12)
    at REPLServer.emit (events.js:107:17)
    at REPLServer.Interface._onLine (readline.js:206:10)
    at REPLServer.Interface._line (readline.js:535:8)
> fs.symlinkSync('README.md', 'tmp/README.md', 'dir')
Error: EISDIR, illegal operation on a directory 'D:\cc\ui-7\tmp\README.md'
    at Error (native)
    at Object.fs.symlinkSync (fs.js:767:18)
    at repl:1:4
    at REPLServer.defaultEval (repl.js:133:27)
    at bound (domain.js:254:14)
    at REPLServer.runBound [as eval] (domain.js:267:12)
    at REPLServer.<anonymous> (repl.js:280:12)
    at REPLServer.emit (events.js:107:17)
    at REPLServer.Interface._onLine (readline.js:206:10)
    at REPLServer.Interface._line (readline.js:535:8)
> fs.symlinkSync('README.md', 'tmp/', 'file')
Error: EPERM, operation not permitted 'README.md'
    at Error (native)
    at Object.fs.symlinkSync (fs.js:767:18)
    at repl:1:4
    at REPLServer.defaultEval (repl.js:133:27)
    at bound (domain.js:254:14)
    at REPLServer.runBound [as eval] (domain.js:267:12)
    at REPLServer.<anonymous> (repl.js:280:12)
    at REPLServer.emit (events.js:107:17)
    at REPLServer.Interface._onLine (readline.js:206:10)
    at REPLServer.Interface._line (readline.js:535:8)
[...]

PR #8936, opened 2014-12-23
Revert "doc: clarify fs.symlink and fs.symlinkSync parameters"

piscisaureus:
The offending doc change clarified nothing, but put the arguments to
fs.symlink() and fs.symlinkSync() in the wrong order.

This patch should be applied to v0.10 too.

See also #8920

### DIFF
--- a/doc/api/fs.markdown
+++ b/doc/api/fs.markdown
@@ -208,7 +208,7 @@ the completion callback.
 
 Synchronous link(2).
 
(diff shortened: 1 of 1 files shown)


schnittstabil:
I think this won't help, `destination` and `path` aren't better.

In addition this PR would introduce a new inconsistency:
> fs.link(srcpath, dstpath, callback)
> fs.symlink(destination, path[, type], callback)

PR #8941, opened 2014-12-25
url: remove duplicate keys

mscdex:


ronkorving:
Are they really duplicate? I see colons in one set, not in the other.

cjihrig:
Good eye

ronkorving:
And I wasn't even wearing my glasses ;)

PR #8942, opened 2014-12-25
dns: allow v8 to optimize lookup()

mscdex:
Reassigning the values of parameter variables causes
v8 to disable function optimization.

### DIFF
--- a/lib/dns.js
+++ b/lib/dns.js
@@ -21,7 +21,7 @@
 
 var cares = process.binding('cares_wrap'),
     net = require('net'),
-    isIp = net.isIP;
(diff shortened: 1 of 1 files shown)


jasnell:
@mscdex ... is this still something you think needs to be done in v0.10?

mscdex:
@jasnell Well, it's an optimization. If anything it might make lookups a tad faster. If it doesn't apply cleanly to v0.10 anymore I guess I wouldn't worry about it.

jasnell:
So long as CI passes, LGTM

jasnell:
Landed in 6502160

Issue #7077, opened 2014-02-08
net: unref() is not persistent


bnoordhuis:
```javascript
var net = require('net');
var server = net.createServer();
server.unref();
server.listen();
net.connect(server.address().port).unref();
```

It hangs because of the unref() that comes before listen() and the one after connect().  In that vein:

```javascript
require('net').connect(42).unref();
```

It fails with ECONNREFUSED because the unref() is a no-op.  The expected behavior for both scripts is that they simply quit.  Affects v0.10 and master.

The documentation doesn't mention when it's okay to call unref().  For the sake of user friendliness, net.Socket and net.Server should remember the ref state and retroactively apply it when the handle is created.  Probably applies to other handle types as well.

cjihrig:
https://github.com/nodejs/io.js/commit/85d99830096a48b7d50cc3b0e5c3fba4172c2d02 just landed. It can either be ported, or we can close this and pick up the changes in the converged branch.

PR #8966, opened 2015-01-03
net: make Server.prototype.unref() persistent

cjihrig:
Currently, the `unref()` method does not remember any state if called before the server's handle has been created. This commit adds state to track calls to `ref()` and `unref()`.

Related to #7077. Note - a similar fix is possible for `Socket.prototype.unref()` (which I have code for). However, `Socket` appears to have additional `unref()` issues other than persistence.

### DIFF
--- a/lib/net.js
+++ b/lib/net.js
@@ -1041,6 +1041,7 @@ function Server(/* [ options, ] listener */) {
   this._handle = null;
   this._usingSlaves = false;
   this._slaves = [];
+  this._unref = false;
 
   this.allowHalfOpen = options.allowHalfOpen || false;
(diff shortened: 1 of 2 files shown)


tjfontaine:
This looks like it should be targeted at v0.10 as a bug fix, are there other places this is true -- for instance in dgram?

cjihrig:
Closing in favor of #9115.

@tjfontaine Regarding dgram - the following code prints 'bound' and then exits. Removing the `unref()` causes the program to not exit.

```javascript
var dgram = require('dgram');
var socket = dgram.createSocket('udp4');

socket.unref();
socket.bind(8000, function() {
  console.log('bound');
});
```

Issue #8948, opened 2014-12-27
Debugger in 0.10.35 is corrupted


3y3:
After 45f1330425b671905a02fe30ee7a1dd9544c2709 there is some problem in `debug-debugger.js`
[L1964](https://github.com/joyent/node/blob/v0.10.35-release/deps/v8/src/debug-debugger.js#L1964), [L2224](https://github.com/joyent/node/blob/v0.10.35-release/deps/v8/src/debug-debugger.js#L2224)

I'm not sure that this is all errors.

[Related issue in Node Inspector](https://github.com/node-inspector/node-inspector/issues/461#issuecomment-68174466)

pflannery:
+1

sbrinkmann:
+1

justsml:
Ahhhhhhh, I wasted way too much time on this bug.... +1

ariellyrycs:
+1

bajtos:
Pull request that should fix the problem: #8973

wlepinski:
@dotnetCarpenter You had to install it from source? The only available source for me on nodesource is 0.10.35 and 0.10.25 on archive.ubuntu :+1:

dotnetCarpenter:
@wlepinski I'm on Mac OSX at the moment. But you can always find the linux binaries on *nodejs.org/dist/*.
Linux 64: http://nodejs.org/dist/v0.10.33/node-v0.10.33-linux-x64.tar.gz
Linux 32: http://nodejs.org/dist/v0.10.33/node-v0.10.33-linux-x32.tar.gz

ekulabuhov:
@dotnetCarpenter thank you! Downgrading to 0.10.34 helped.
Packages for all OSs here: http://blog.nodejs.org/2014/10/23/node-v0-10-33-stable/

2color:
+1

misterdjules:
We're most likely going to revert the breaking change with #8999. If you guys had some time to give it a try and let us know if it fixes your issues, that would be very much appreciated.

If you need help with testing it, please let me know and I'll be more than happy to assist you.

Thank you!

misterdjules:
Landed in 31051e5c7b253b6a0006b536188fc0494bd6e082 and forward ported to v0.12 in merge commit fbfe562d71ae8d8f8bbf627808c755e513e4e905.

bajtos:
That's great news, thank you. Is there any estimate when the next v0.10 release will be published?

misterdjules:
[...]

PR #8973, opened 2015-01-05
deps: fix v8 after 45f1330

bajtos:
45f1330 enabled "use strict" in "deps/v8/src/debug-debugger.js", which surfaced usage of undefined variables in this file.

```
ReferenceError: frame_index is not defined
ReferenceError: includeSource is not defined
```

This patch fixes the problem in Node until it is fixed upstream.

Node issue: #8948
V8 issue: https://code.google.com/p/v8/issues/detail?id=3790

@misterdjules @tjfontaine

### DIFF
--- a/deps/v8/src/debug-debugger.js
+++ b/deps/v8/src/debug-debugger.js
@@ -1961,7 +1961,7 @@ DebugCommandProcessor.prototype.resolveFrameFromScopeDescription_ =
   // Get the frame for which the scope or scopes are requested.
   // With no frameNumber argument use the currently selected frame.
   if (scope_description && !IS_UNDEFINED(scope_description.frameNumber)) {
-    frame_index = scope_description.frameNumber;
+    var frame_index = scope_description.frameNumber;
     if (frame_index < 0 || this.exec_state_.frameCount() <= frame_index) {
       throw new Error('Invalid frame number');
     }
@@ -2221,7 +2221,7 @@ DebugCommandProcessor.prototype.lookupRequest_ = function(request, response) {
 
   // Set 'includeSource' option for script lookup.
   if (!IS_UNDEFINED(request.arguments.includeSource)) {
-    includeSource = %ToBoolean(request.arguments.includeSource);
+    var includeSource = %ToBoolean(request.arguments.includeSource);
     response.setOption('includeSource', includeSource);
   }
 


bajtos:
@chrisdickinson perhaps you can review this patch instead of @tjfontaine and/or @misterdjules? It is fixing a regression in V8 debugger introduced in v0.10.35.

> 45f1330 enabled "use strict" in "deps/v8/src/debug-debugger.js", which surfaced usage of undefined variables in this file.

> ```
ReferenceError: frame_index is not defined
ReferenceError: includeSource is not defined
```

bajtos:
FWIW, there are at least 14 people that have run into this problem in the last week or so: node-inspector/node-inspector/issues/461

Can somebody from the Node core team look at this pull request and/or the root problem please?

@trevnorris @indutny

ChrisGciso:
I reverted to 10.33 to avoid the issue for now, but would like to be on the latest stable version (10.35). Need the debugger to work.

misterdjules:
@bajtos @ChrisGciso I just got back from vacation, I'm looking at it this morning. I'll keep you posted. Thank you!

misterdjules:
The change looks good to me. I ran node-inspector's tests suite and the results are back to what they were before the regression was introduced by 45f1330.

In my mind, we have two options:

1) Land this change

https://github.com/joyent/node/issues/8685 would still be fixed and the regression for node-inspector would be fixed too.

Cons:
* This adds another floating patch for V8 (it's difficult to know when and how https://code.google.com/p/v8/issues/detail?id=3790 will land).
* It's difficult for me to be 100% sure that this would fix all regressions introduced by 45f1330. However, as mentioned above, node-inspector's tests suite has the same results with this change, so my confidence is fairly high.

2) Revert 45f1330 and land it later in a development branch after https://code.google.com/p/v8/issues/detail?id=3790 lands upstream

https://github.com/joyent/node/issues/8685 would not be fixed anymore in upcoming 0.10.x releases, but we would know for sure that there's no regression in node-inspector. 

I like 1) as it moves things forward, but I would like to consider 2) for two reasons:
* node-inspector is in my opinion a more important use case than being able to use `--use-strict` with Node.js' debugger.
* It would prevent us from having to float another V8 patch in the v0.10 branch.

@bajtos @tjfontaine @trevnorris @chrisdickinson @orangemocha @cjihrig @indutny What do you guys think?

bajtos:
I am personally preferring 1). As you @misterdjules said, it is moving things forward, and the risks are IMO low, since the debugger is not a production-critical part of Node. Even if this patch did not fix all regressions, we could always fix the remaining ones later, when they are eventually discovered.

cjihrig:
@misterdjules I don't have a preference either way, but I do agree that getting node-inspector working is a much higher priority than making `--use-strict` work with the debugger.

trevnorris:
+1 to @cjihrig

2color:
+1 @cjihrig

glasser:
Once this decision is made, it would be great to see a speedy 0.10.36 with it.

misterdjules:
V8 landed [a similar patch](https://codereview.chromium.org/813873007) that would probably fix #8948. However, we couldn't really test this in the wild and make sure that we wouldn't be breaking things again, so I think we should:
* Revert the breaking change, which is what #8999 does.
* Land https://codereview.chromium.org/736583007 and https://codereview.chromium.org/813873007 in a development branch (either 0.11.15 or more likely master) so that it can be tested more thoroughly.

Closing this PR anyway as even if we decided to not revert the breaking change, we would backport https://codereview.chromium.org/813873007 instead of merging this PR. However, feel free to comment if you disagree with my suggestions.

Thank you all again, and I apologize of breaking it in the first place.

timothyjlaurent:
OMG... how can we get this fix now? I need to debug... very inconvenient...

misterdjules:
@timothyjlaurent #8999 fixes the same issue, and we think it's less risky to land in a stable version. We're working on getting it in the next stable release.

chuyik:
I updated node to 0.10.36, and the `ReferenceError: frame_index is not defined` error was gone. :+1:

misterdjules:
@chuyik Thank you very much for letting us know!

Issue #8947, opened 2014-12-26
Newlines in HTTP request method results in the injection of HTTP headers in requests


aegarbutt:
The http request method does not have any validation applied to it before constructing the first line of an HTTP request (https://github.com/joyent/node/blob/master/lib/_http_client.js#L129). A carefully constructed method, such as
`GET / HTTP/1.1\r\nX-Foobar: Bazbang\r\nX-Discard:`
if passed to `http.request(method, '/intendedpath')` would result in an HTTP request like:

```
GET / HTTP/1.1
X-FOOBAR: BAZBANG
X-DISCARD: /intendedpath HTTP/1.1
```

The forced uppercasing that occurs at https://github.com/joyent/node/blob/master/lib/_http_client.js#L89 can be bypassed by URL encoding the method.

Security impact occurs when a server constructs and submits HTTP requests from client provided data.

HTTP request methods are limited in the HTTP/1.0 and 1.1 RFCs to `token`, which is defined as:

```
token          = 1*<any CHAR except CTLs or tspecials>
CTL            = <any US-ASCII control character
                    (octets 0 - 31) and DEL (127)>
tspecials      = "(" | ")" | "<" | ">" | "@"
               | "," | ";" | ":" | "\" | <">
               | "/" | "[" | "]" | "?" | "="
               | "{" | "}" |  SP |  HT
```

Header injection within headers themselves is accounted for at:
https://github.com/joyent/node/blob/master/lib/_http_outgoing.js#L296-L297

jasnell:
I can confirm this but the syntax appears to be a bit off above... Verified by using `nc -l 9999` then doing `http.request({host:'localhost', port:9999,method:'GET /foo\r\nX-Boo: ', path:'/bar'}, function(e,d) {}).end();`. The `nc` output shows `GET /FOO` with `X-Boo: /bar`. 

There would be a slight performance hit to consider, but checking the options.method against `var token = /^[^\x00-\x31\x127\(\)<>@,;:\\"/\[\]\?=\{}]+$/` would likely be a good idea.

brendanashworth:
This has been fixed in master as of https://github.com/nodejs/node/commit/6192c9892f3eddf165a4826ba917ea08ea627ea2. @jasnell is this something you'd like to see backported to 0.10 or 0.12?

jasnell:
Yes. To v0.12 at least.

PR #8976, opened 2015-01-05
V0.10.35 http method check

jasnell:
Per #8947, the http.request constructor is not properly checking the validity
of the HTTP method input by the user. The problem allowed HTTP request
headers to be injected via the method option as reported in #8947. This fixes
the problem by introducing a quick check against a the HTTP token rule to 
ensure that the method is valid.

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -30,7 +30,7 @@ var FreeList = require('freelist').FreeList;
 var HTTPParser = process.binding('http_parser').HTTPParser;
 var assert = require('assert').ok;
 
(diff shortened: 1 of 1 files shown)


jasnell:
Yep, another approach is to scan through each char and throw on the first non-token char (or not throw and fallback to the default, which seems bad too).

(I thought I had squashed those commits together... how odd. Will retarget this to v0.10 tomorrow and resubmit.)

misterdjules:
Another thing I failed to mention previously is that we'd need to have tests for this change.

jasnell:
Replacing with .. https://github.com/joyent/node/pull/9017

Issue #8829, opened 2014-12-05
Proposed changes to querystring parse() and stringify()


cjihrig:
There are a few things I would like to change about the querystring methods.

- `stringify()` has an undocumented `options` argument. This is likely just a documentation error, as `parse()` has the same signature, with `options` documented.
- Both methods take optional `sep` and `eq` arguments. Supporting optional arguments like this is a pain, especially when both are of the same data type. It also doesn't make a ton of sense given that an `options` object can also be passed (perhaps this was done for backwards compatibility).
- There have been a number of issues related to how querystring deals with `null`, `undefined`, empty string, etc. There is not currently a good way to tweak querystring to your individual needs. A few relevant issues are #7971 and #8825.

I recommend making the following changes. Unfortunately, some are breaking changes:

- The signature for `stringify()` should be `stringify(obj[, options])`, with the existing `sep` and `eq` arguments moved into `options`. Same deal with `parse()`.
- Add an optional function to let the user define how they would like to parse and stringify values. If this is not passed, Node's current implementation is used.

jasnell:
Looking it over, this seems like a good change. The optional function would also be passed within `options`, correct?

cjihrig:
@jasnell yes, under my proposal, a single options object would be passed along with the stringify/parse value.

jasnell:
@cjihrig ... any updates on this? Was thinking this could be deferred to the converged stream.

cjihrig:
I'm more inclined to just close this. It's unlikely to ever get through.

PR #8977, opened 2015-01-05
Improve querystring parse and stringify

jasnell:
Per https://github.com/joyent/node/issues/8829

Improve querystring.stringify and querystring.parse so that
undefined and null are optionally omitted. Allow sep, eq,
and name to be passed in options instead of as separate
parameters. Add new omitnull option. See documentation
changes for example.


jasnell:
Will reopen against nodejs/node if necessary

PR #8990, opened 2015-01-07
doc: clarify http.request language

skeggse:
I can't even count the number of times I've looked at the http docs to see whether `host` or `hostname` lets you include a port for `http.request`.

### DIFF
--- a/doc/api/http.markdown
+++ b/doc/api/http.markdown
@@ -454,9 +454,10 @@ automatically parsed with [url.parse()][].
 
 Options:
 
-- `host`: A domain name or IP address of the server to issue the request to.
-  Defaults to `'localhost'`.
-- `hostname`: To support `url.parse()` `hostname` is preferred over `host`
+- `host`: A domain name or IP address of the server to issue the request to,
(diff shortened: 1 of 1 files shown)


jasnell:
@cjihrig @chrisdickinson ... not convinced this one is necessary. Inclined to close but wanted to check with you first.

skeggse:
Just to expand on this, the current language in the docs says that the `host` parameter is "A domain name or IP address of the server to issue the request to. Defaults to 'localhost'," and described the `hostname` parameter with "To support url.parse() hostname is preferred over host." The `host` description actually fits the `hostname` parameter better, and it seems like the `host` description needs to say something about allowing a port.

jasnell:
@skeggse ... yeah, the description in general should be improved. Tell you what... can you close this here and open this as a new PR against http://github.com/nodejs/node master? Active development is moving there and I'd like to get others to weigh in on the change.

cjihrig:
I'm in favor of improving the documentation here. I also agree that we should move this over to nodejs/node. Going to close here.

PR #8991, opened 2015-01-07
docs: delete unused/duplicate css files

robertkowalski:
- `sh.css` already exists in `api_assets`
- `sh_vim-dark.css` is unused, but used in the repo `node-website`
      now

I think they are leftovers from the code thats now in `joyent/node-website`, as they are still used there


trevnorris:
LGTM. Will land on 0.12 once the merge is done.

misterdjules:
@trevnorris This seems to apply to v0.10 too, so I would suggest landing in v0.10 first and then merging into v0.12.

misterdjules:
@robertkowalski Landed in 0c50195 and 307daed, thank you!

PR #8993, opened 2015-01-08
Docs: Fix HTTP Request example

ruben1:
Changes: 
1.- On the "data" event in the response, wait for all the data to arrive and append that data to an empty string
2.- On the "end" event in the response, console log the body response after it has been appended to an empty string in the "data" event

### DIFF
--- a/doc/api/http.markdown
+++ b/doc/api/http.markdown
@@ -500,8 +500,12 @@ Example:
       console.log('STATUS: ' + res.statusCode);
       console.log('HEADERS: ' + JSON.stringify(res.headers));
       res.setEncoding('utf8');
+      var str = '';
       res.on('data', function (chunk) {
-        console.log('BODY: ' + chunk);
+        str += chunk;
+      });
+      res.on('end', function() {
+        console.log('BODY: ' + str);
       });
     });
 


doc/api/http.markdown
```diff
@@ -500,9 +500,13 @@ Example:
       console.log('STATUS: ' + res.statusCode);
       console.log('HEADERS: ' + JSON.stringify(res.headers));
       res.setEncoding('utf8');
+      var str = "";
```
cjihrig: This should use single quotes like the rest of the example.

doc/api/http.markdown
```diff
@@ -500,9 +500,13 @@ Example:
...
       res.on('data', function (chunk) {
-        console.log('BODY: ' + chunk);
+        str += chunk;
       });
+      res.on("end", function() {
```
cjihrig: Single quotes please.

doc/api/http.markdown
```diff
@@ -500,9 +500,13 @@ Example:
...
+        console.log('BODY: ' + str);
+      })
```
cjihrig: Missing a semicolon

cjihrig:
Please update the commit message to adhere to the guidelines in [CONTRIBUTING.md](https://github.com/joyent/node/blob/master/CONTRIBUTING.md).

ruben1:
Thanks for the feedback. I amended the last commit

jasnell:
@cjihrig .. it looks like a somewhat similar change was already made in nodejs/node that would make this a bit superfluous. I'm inclined to close without further action but wanted to check with you first.

cjihrig:
If the same change already exists on nodejs/node, then I agree that we don't need this. If the change hasn't been made over there, feel free to open this there :-)

PR #8994, opened 2015-01-08
build: add small-icu support for binary packages

### LINKED ISSUE #7676: Intl: Build "English-only" binaries for the distribution.

misterdjules:
Invokes the configure script used to build binary packages
(OSX pkg, binary tarballs, pkgsrc, MSI) with --download=all
--with-intl=small-icu.

Also makes PACKAGEMAKER customizable, because PackageMaker is not
necessarily installed in /Developer on OSX anymore.

Tested all binary packages on Windows, OSX, Linux and SmartOS.

Fixes #7676.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -258,7 +258,7 @@ TARBALL=$(TARNAME).tar.gz
 BINARYNAME=$(TARNAME)-$(PLATFORM)-$(ARCH)
 BINARYTAR=$(BINARYNAME).tar.gz
 PKG=out/$(TARNAME).pkg
-packagemaker=/Developer/Applications/Utilities/PackageMaker.app/Contents/MacOS/PackageMaker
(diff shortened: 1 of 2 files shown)


srl295:
intl part LGTM

trevnorris:
LGTM

misterdjules:
@srl295 @trevnorris Thanks for the review, landed in 67f87a743dee4951374dd240c25e31cd507e6c5f.

oncletom:
That's great :-)

PR #8998, opened 2015-01-08
test: fix test-process-active-wraps.js

### LINKED ISSUE #8986: test-process-active-wraps.js fails on windows
It seems that commit b636ba8186d191c52ee36f2f2b1aebbbb4d95575 introduced the following regression on Windows:
```
>.\Release\node.exe test\simple\test-process-active-
wraps.js

assert.js:98
        ^
AssertionError: 1 == 0
    at Immediate._onImmediate (c:\Users\jgilli\dev\node-bis\test\simple\test-pro
cess-active-wraps.js:69:16)
    at processImmediate [as _immediateCallback] (timers.js:340:17)

>
```

@cjihrig Would you mind taking a look?

cjihrig:
b636ba8186d191c52ee36f2f2b1aebbbb4d95575 caused a regression on Windows due to the way server handles are cleaned up. This commit fixes the test by allowing the handle to be cleaned up. Closes #8986.

cc: @piscisaureus @misterdjules

### DIFF
--- a/test/simple/test-process-active-wraps.js
+++ b/test/simple/test-process-active-wraps.js
@@ -49,7 +49,7 @@ var handles = [];
 
   expect(1, 0);
   var conn = net.createConnection(common.PORT);
-  conn.on('lookup', onlookup); 
+  conn.on('lookup', onlookup);
   conn.on('error', function() { assert(false); });
   expect(2, 1);
   conn.destroy();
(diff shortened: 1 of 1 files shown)


piscisaureus:
LGTM, landed in b5c9dcb6d73d97eb2763e368c693da4a62b2f362

test/simple/test-process-active-wraps.js
```diff
@@ -65,9 +65,12 @@ var handles = [];
   });
   function onclose() {
     if (++n === handles.length) {
-      setImmediate(function() {
-        assert.equal(process._getActiveHandles().length, 0);
-      });
+        // give server handle a chance to close - see GH #8986
```
misterdjules: Do you think we could add to this comment (or at the beginning of this file, as a general comment) that this test depends a lot on the implementation of how handles are closed, and as such can and will likely break in the future? 

We could add a note saying that if the test breaks, it may be because of a correct change in the way handles are closed, and it doesn't necessarily means that node is broken.

It may save us some time next time this test breaks.

misterdjules:
@cjihrig  Except for my inline comment, LGTM.

cjihrig:
Thanks! @misterdjules I updated the comment as you asked. Landed in 1fad3730c2d6130c13b46ad78ded808fe9bb4c10

misterdjules:
@cjihrig @piscisaureus Thanks you!

PR #9001, opened 2015-01-09
readline: Fix output of completions when # of columns is unknown

### LINKED ISSUE #8644: repl autocomplete broken when output is socket
```
repl.start({
  prompt: 'remote> ',
  input: sock,
  output: sock,  // (replace this with process.stdout and autocomplete works fine)
  terminal: true
});
```

Problem: Tab-autocomplete will only display empty lines. (Tested on 0.10.32 win, 0.10.29 linux).
Workaround:  Set  `sock.columns=80;`

Probable cause:
https://github.com/joyent/node/blob/master/lib/readline.js#L414
  results in  minRows = 0, because maxColumns=Infinity  (via #L382, from #L158)

ackalker:
When output.columns is undefined, tab-completion outputs blank
lines instead of completions. Fix this by outputting one column
of completions.

Fixes #8644

### DIFF
--- a/lib/readline.js
+++ b/lib/readline.js
@@ -379,7 +379,10 @@ Interface.prototype._tabComplete = function() {
         var width = completions.reduce(function(a, b) {
           return a.length > b.length ? a : b;
         }).length + 2;  // 2 space padding
-        var maxColumns = Math.floor(self.columns / width) || 1;
+        var maxColumns = Math.floor(self.columns / width);
+        if (maxColumns === Infinity) {
+          maxColumns = 1;
+        }
         var group = [], c;
         for (var i = 0, compLen = completions.length; i < compLen; i++) {
           c = completions[i];
--- a/test/simple/test-repl-tab-complete-output.js
+++ b/test/simple/test-repl-tab-complete-output.js
@@ -1,0 +1,79 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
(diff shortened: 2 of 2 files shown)


TooTallNate:
Needs tests. And a before/after of the output now in a comment would be helpful as well.

ackalker:
@TooTallNate 
Needs a comment, where? In the code? Wouldn't that bloat the size of lib/readline.js embedded in Node?
Or do you mean a comment in the commit message?

TooTallNate:
> Needs a comment, where?

On this PR :)

ackalker:
Ah, OK, I see.

Note that testing tab-completion *output* isn't currently possible with the test fixture in `test/simple/test-repl-tab-complete.js` because it has no provisions for capturing real repl output stream output, not just results from `complete()`.

I'm afraid that implementing that is a 'little' bit more involved, could be relevant to testing many more issue fixes than just this one (pause/resume behaviour of repl/readline input/output streams, etc.) and, honestly, goes a little bit above my pay grade, i.e. $0. :-)

ackalker:
Hm, perhaps my comment above was a bit blunt. I see that I can steal some code from `test/simple/test-repl.js`, I'll see what I can do with that.

ackalker:
Here's my first stab at a new test to be added to `test/simple/test-repl-tab-complete.js` (naturally, the `require` should eventually go at the top of the file):

```js
// tab completion should output completion candidates even if output.columns
// is undefined.
var spawn = require('child_process').spawn;
var args = ['-e', 'require("repl").start({input: process.stdin, output: process.stdout, terminal: true});'];
var child = spawn(process.execPath, args);

var input = 'process.std\t';
var expectOut = /process.stderr\r\nprocess.stdin\r\nprocess.stdout\r\n/;

child.stderr.setEncoding('utf8');
child.stderr.on('data', function(c) {
  throw new Error('child.stderr be silent');
});

child.stdout.setEncoding('utf8');
var out = '';
child.stdout.on('data', function(c) {
  out += c;
});
child.stdout.on('end', function() {
  assert(expectOut.test(out));
  console.log('ok');
});

child.stdin.end(input);
```

This test fails on unpatched `lib/readline.js`, passes with my commit applied.
It can do with another test with `process.stdout.columns=80` or something, after which things will need some refacturing, but first I would like to know your opinion.

TooTallNate:
Seems like a good approach to me!

ackalker:
Thanks, I've pushed the tests.

ackalker:
oops, that extra semicolon is a free gift thanks to flycheck and jshint ;-)

ackalker:
I've edited-in the before/after comments after your request for them, to try to keep this thread tidy, hope you don't mind :-)

ackalker:
Because of issue #9006 , I've put my tests in a separate file, sorry for the forced update.

lib/readline.js
```diff
@@ -379,7 +379,10 @@ Interface.prototype._tabComplete = function() {
         var width = completions.reduce(function(a, b) {
           return a.length > b.length ? a : b;
         }).length + 2;  // 2 space padding
-        var maxColumns = Math.floor(self.columns / width) || 1;
+        var maxColumns = Math.floor(self.columns / width);
```
smilingthax: Leaving out the ` || 1`, i.e. allowing `maxColumns==0` (e.g. set `self.columns==1` while `width>1`) is probably not a good idea, because it will send handleGroup into an infinite loop (via `minRows==Infinity`)...

jasnell:
@ackalker ... how would you like to proceed with this? If it's still something you'd like to pursue, opening a new updated PR against master on http://github.com/nodejs/node would likely be the best approach. Otherwise, the PR could possibly be retargeted at the v0.12 branch. Either way, I don't think this can land in it's current state.

jasnell:
Opened a new issue in nodejs/node to track. Will need a new PR against nodejs/node master or v0.12. Closing this here.

PR #9003, opened 2015-01-09
fix assert() spec

bsnote:
"message" argument is optional for both assert() and assert.ok().

### DIFF
--- a/doc/api/assert.markdown
+++ b/doc/api/assert.markdown
@@ -9,7 +9,7 @@ access it with `require('assert')`.
 
 Throws an exception that displays the values for `actual` and `expected` separated by the provided operator.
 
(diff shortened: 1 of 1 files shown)


cjihrig:
They are the same function, so it makes sense. LGTM

trevnorris:
LGTM. @cjihrig feel free to merge.

cjihrig:
Thanks! Landed in 0.10 in e45c52c573cc2799749050c60228a769231674f8. Landed in 0.12 in 430096a90f69fc98b1786dfa0009c02b0d011143. There was a documentation update in 0.12 to use correct notation for optional arguments. I amended this in 0.12 to use that notation.

Issue #8992, opened 2015-01-07
url resolve(href, '.') sometimes leaves off trailing slash


blatx:
The following script:

    var url = require('url');
    console.log(url.resolve('http://www.example.com/path/to/edit', 'view'));
    console.log(url.resolve('/path/to/edit', 'view'));
    console.log(url.resolve('http://www.example.com/path/to/edit', '.'));
    console.log(url.resolve('/path/to/edit', '.'));

produces the following output:

    http://www.example.com/path/to/view
    /path/to/view
    http://www.example.com/path/to/
    /path/to

which is: resolving '.' produces a result that contains a trailing slash if the base url is absolute including host, but leaves it off if there is no host name (AKA "relative to the domain root").

These are very different urls both for a webserver and as a base url for a browser.

On every browser I have tested, using the following html

    <html>
    <head>
    <base href="/path/to/edit" />
    </head>
    <body>
    <a href="." onclick="alert(this.href); return false">test</a>
    </body>
    </html>

the href of the test link also always includes the slash.

Tested with v0.10.35

amir-s:
I'm not sure if this is a bug or not.

In the source code (https://github.com/joyent/node/blob/master/lib/url.js#L473) it is mentioned: "urlParse appends trailing / to urls like http://www.example.com"

blatx:
"http://www.example.com" and "http://www.example.com/" represent the exact same thing: the root in the domain. This is something else.

I found the following comment and code [in the same file you pointed to](https://github.com/joyent/node/blob/master/lib/url.js#L627):

      // if a url ENDs in . or .., then it must get a trailing slash.
      // however, if it ends in anything else non-slashy,
      // then it must NOT get a trailing slash.
      var last = srcPath.slice(-1)[0];
      var hasTrailingSlash = (
          (result.host || relative.host) && (last === '.' || last === '..') ||
          last === '');

I'm sure the "bug" is in that test for the host in `(result.host || relative.host)`. I see *no good reason at all* to let that behaviour depend on whether a host as provided, or not. There is no well-known browser in the world that behaves that way, and compatibility with browser behaviour is supposed to be a very high priority for this module.

"." and ".." are directory specs, "/path/to/" is a directory spec, but "/path/to" is not. 

Hence, the trailing slash should *always* be there.

amir-s:
You are right.

Actually `url.resolve('/blah', '.')` produces an empty string too, which should produce `/`.

I managed to solve the problem on my machine. I'll send a pull request.

[...]

PR #9005, opened 2015-01-10
resolving urls with '.' and '..' #8992

amir-s:
The array `srcPath` contains url segments. If the path doesn't contain a hostname, the first element will be an empty string. In that case, if the last element is `.` or `..`, the final value must end with a trailing slash.


amir-s:
Tests are passed!

PR #9010, opened 2015-01-11
url: reslove urls with . and ..

amir-s:
`.` and `..` are directory specs and resolving urls -with or without the hostname- with `.` and `..` should add a trailing slash to the end of the url. (#8992)

The array `srcPath` contains url segments. If the path doesn't contain a hostname, the first element will be an empty string. In that case, if the last element is `.` or `..`, the final value must end with a trailing slash.


trevnorris:
LGTM. I'll land it after the 0.10 to 0.12 merge.

misterdjules:
That also affects v0.10:
```
➜  node git:(v0.10) ✗ ./node --version
v0.10.36-pre
➜  node git:(v0.10) ✗ ./node -p 'require("url").resolve("/foo/", ".");'        
/foo
➜  node git:(v0.10) ✗
```
Thus, I suggest landing this in v0.10 first. Removing from the 0.11.15 milestone as a consequence. I'm not sure when we want to land it in v0.10, so I'm not adding it to any v0.10 milestone for now, but feel free to do so.

trevnorris:
@misterdjules Sounds like a good idea. Might as well land this on the v0.10 branch now. Can't think of a reason to hold it back. Though, does it apply cleanly?

misterdjules:
@trevnorris It applies cleanly to v0.10.

trevnorris:
@misterdjules Feel free to throw it on.

amir-s:
When this is going to be merged?

trevnorris:
Thanks. Landed in ad0684807c474.

PR #9011, opened 2015-01-12
Change an ungraceful repetition of 'that'

kxprim:
The sentence felt clunky so rephrased the second part of the sentence to:
- "if there's nothing in the write buffer, then this means nothing more will ever be provided".
  Instead of:
- "if there's nothing in the write buffer, then that means that nothing more will ever be provided".


jasnell:
Going to close this as we're not landing in master here any more. A new PR can be opened against nodejs/node master if necessary

PR #9017, opened 2015-01-12
http: Check for malformed HTTP method for request

jasnell:
Add a check to make sure that the specified HTTP method is a valid
token per the spec. To do so, we look at the method string and check
that each character is valid per the token rule. The first violation
aborts the check and throws.

This check is necessary to avoid malicious http request header
injection. per https://github.com/joyent/node/issues/8947

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -1344,6 +1344,22 @@ Agent.prototype.removeSocket = function(s, name, host, port, localAddress) {
 var globalAgent = new Agent();
 exports.globalAgent = globalAgent;
 
+function check_is_token(val) {
+  if (!val || typeof val !== 'string')
+    return false;
+  for (var n = 0, l = val.length; n < l; n++) {
+    var c = val.charCodeAt(n);
+    if (c <= 32 || c == 34 ||
+        c == 44 || c == 47 ||
+        c == 123 || c == 125 ||
+        c == 40 || c == 41 ||
+        (c >= 58 && c <= 64) ||
+        (c >= 91 && c <= 93) || 
+        c >= 128)
+      return false;
+  }
+  return true;
+}
 
 function ClientRequest(options, cb) {
   var self = this;
@@ -1365,6 +1381,9 @@ function ClientRequest(options, cb) {
   self.socketPath = options.socketPath;
 
   var method = self.method = (options.method || 'GET').toUpperCase();
+  if (!check_is_token(method)) {
+    throw new Error('Malformed HTTP request method');
+  }
   self.path = options.path || '/';
   if (cb) {
     self.once('response', cb);
@@ -2139,6 +2158,9 @@ Client.prototype.request = function(method, path, headers) {
     headers = path;
     path = method;
     method = 'GET';
+  }
+  if (!check_is_token(method)) {
+    throw new Error('Malformed HTTP request method');
   }
   options.method = method;
   options.path = path;
--- a/test/simple/test-http-malformed-method.js
+++ b/test/simple/test-http-malformed-method.js
@@ -1,0 +1,34 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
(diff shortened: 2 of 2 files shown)


trevnorris:
@misterdjules @cjihrig Thoughts?

cjihrig:
I'm +1 on fixing the problem. Does anyone know how much of a security risk do we face by ensuring that the request method matches the token description precisely? In other words, could we safely get away with cheating and just detecting newlines?

lib/http.js
```diff
@@ -1344,6 +1344,22 @@ Agent.prototype.removeSocket = function(s, name, host, port, localAddress) {
 var globalAgent = new Agent();
 exports.globalAgent = globalAgent;
 
+function check_is_token(val) {
```
misterdjules: `check_is_token` sounds a bit generic, even though this is in `lib/http.js`. Maybe `check_is_http_token` is more explicit?

misterdjules: Also, I think it would help to mention that this code is written to conform to the HTTP/1.1 RFC, and from which section comes the set of forbidden characters. Otherwise, it might look like it was chosen arbitrarily.

test/simple/test-http-malformed-method.js
```diff
@@ -0,0 +1,34 @@
...
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+var common = require('../common');
+var assert = require('assert');
+var http = require('http');
+
+// this should throw
+var c = 0;
+try {
```
misterdjules: We could maybe use `assert.doesNotThrow` here?

misterdjules:
Commented inline, I have a few additional comments:
* The code does not pass `make jslint`.
* I think the test provided could be more thorough. Among other things, it could:
 * Test that valid `method` options do not throw (using `assert.doesNotThrow`).
 * Test more invalid characters. If we're making this part of the API more RFC compliant and not just fixing the header injection issue, I think we should test what the RFC describes more thoroughly.

What do you think?

test/simple/test-http-malformed-method.js
```diff
@@ -0,0 +1,34 @@
...
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+var common = require('../common');
+var assert = require('assert');
+var http = require('http');
+
+// this should throw
```
misterdjules: Generally, I like to mention using comments what the goal of a given test is. We could mention the link to the given issue, and the rationale for this test (not only fixing the "header injection" issue, but also making the API more compliant with the HTTP RFCs). It helps when revisiting tests much later and trying to understand the intent of the author(s).

lib/http.js
tjfontaine: I agree we want to change the name, and I think general we're more `camelCase` style -- perhaps `checkValidToken`

lib/http.js
```diff
@@ -1344,6 +1344,22 @@ Agent.prototype.removeSocket = function(s, name, host, port, localAddress) {
...
+    if (c <= 32 || c == 34 ||
+        c == 44 || c == 47 ||
+        c == 123 || c == 125 ||
+        c == 40 || c == 41 ||
+        (c >= 58 && c <= 64) ||
+        (c >= 91 && c <= 93) || 
+        c >= 128)
+      return false;
+  }
+  return true;
```
tjfontaine: Since all callers are just going to take false and `throw` I would advocate this method should just `throw`, then all our state is preserved.

Also as far as the error message, I would like to improve it -- ideally the message would be something like: `"Invalid character in HTTP method at %d for %s", n, val`

Then on the actual error object that's constructed attach the same information.

If we're against throwing in this method, I would advocate returning null or falsey in the happy case, and returning the as described error object in the case where we should throw.

jasnell:
Ok, will refactor this one a bit based on the discussion and get it reworked. Target v0.12?

test/simple/test-http-malformed-method.js
jasnell: Agreed that a more thorough review of the RFC requirements is going to be necessary overall. I've marked that on my short list to do for v0.14. If folks think it's important enough to do sooner, let me know.

misterdjules: My opinion is that without more thorough tests, we won't be able to be confident that this set of changes does what it intends to do, so I would advocate for including them from the start.

misterdjules:
Targeting v0.10 still makes sense to me, as the preferred way to land changes is to do that in the oldest supported & affected branch.

test/simple/test-http-malformed-method.js
jasnell: I'm expanding the test case to check a wider range of valid and invalid characters in the method.

jasnell:
Replaced by #9076

PR #9021, opened 2015-01-13
Fix author attribution.

thughes:
Tom Hughes and Tom Hughes-Croucher are different people.

@sh1mmer


jasnell:
Closing this here. The fix already landed in nodejs/node. Can cherry pick to v0.12 if necessary.

PR #9027, opened 2015-01-14
test: fix ssl/tls options matrix test

### LINKED ISSUE #9020: test/external/ssl-options fails
The [most recent merge from v0.10 to v0.12](https://github.com/joyent/node/pull/9018) introduced a regression: test/external/ssl-options now fails on all supported platforms.

This test is not run as part of the standard tests suite used on our continuous integration platform. It is run manually and is used to make sure that the compatibility of client/server SSL/TLS runtime settings for the default build options is what we expect.

It is a somewhat critical test in the sense that it allows us to make sure that, for instance, a client that didn't enable the SSLv3 protocol cannot connect to a server that enabled it.

misterdjules:
The tests suite available in test/external/ssl-options was originally
written for security fixes made in the v0.10 branch. In this branch, the
client's default ciphers list is compatible with SSLv2.

After merging this change from v0.10 to v0.12, this tests suite was
broken because commits 5d2aef17ee56fbbf415ca1e3034cdb02cd97117c and
f4c8020d1068ffba57458b327c62b61b1f29ec63 make SSL/TLS clients use a
default ciphers list that is not compatible with the SSLv2 protocol.

This change fixes two issues:
1) The cipher list that was setup for a given test was not passed
properly to the client.
2) When either or both of clients/servers were using SSLv2, tests were
expected to succeed when at least the server end was using SSLv2
compatible ciphers. Now, tests are expected to succeed only if
SSLv2 compatible ciphers are used on both ends.

Fixes #9020.

### DIFF
--- a/test/external/ssl-options/test.js
+++ b/test/external/ssl-options/test.js
@@ -169,26 +169,17 @@ function testSetupsCompatible(serverSetup, clientSetup) {
     return false;
   }
 
-  if (isSsl2Protocol(serverSetup.secureProtocol) ||
-      isSsl2Protocol(clientSetup.secureProtocol)) {
-
-      /*
-       * It seems that in order to be able to use SSLv2, at least the server
-       * *needs* to advertise at least one cipher compatible with it.
-       */
-      if (serverSetup.ciphers !== SSL2_COMPATIBLE_CIPHERS) {
-        return false;
-      }
-
-      /*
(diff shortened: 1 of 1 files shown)


trevnorris:
LGTM

misterdjules:
@trevnorris Thanks! Landed in 408bffe.

PR #9037, opened 2015-01-15
cluster: avoid race enabling debugger in worker

### LINKED ISSUE #6440: test-debug-signal-cluster.js failing sporadically
Example: http://jenkins.nodejs.org/job/node-pullrequest/1471/DESTCPU=ia32,label=linux/tapTestReport/test.tap-104/

Running the test a bunch of times in a loop usually makes it fail within 10 or so iterations for me.

tjfontaine:
Previously if a worker's state machine had already transitioned into the
'listening' state when it received the message enabling the debugger,
the worker would never enable its debugger.

Change the logic to allow the 'listening' as a valid state for enabling
the debugger.

Fixes #6440

Also improve the test that covers this mechanism.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -285,7 +285,7 @@ function masterInit() {
       var key;
       for (key in cluster.workers) {
         var worker = cluster.workers[key];
-        if (worker.state === 'online') {
+        if (worker.state === 'online' || worker.state === 'listening') {
           process._debugProcess(worker.process.pid);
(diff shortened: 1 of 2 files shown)


misterdjules:
@tjfontaine Thank you, landed in e9df9a0216550abab4923bcfc6dea4f4b32368d2 and 58fcc6598561b5b99a8c6756b3b1cca55b635b0f!

PR #9038, opened 2015-01-16
doc: i18n: add license content from ICU

srl295:
Node, when built with ICU, should present ICU's license.

Added license info from:
 http://source.icu-project.org/repos/icu/icu/trunk/license.html
All text pasted. Long lines wrapped. (original is HTML.)

Note wording that ICU's license may or may not apply, if node isn't
built with it.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -781,3 +781,388 @@ maintained libraries. The externally maintained libraries used by Node are:
 
        END OF TERMS AND CONDITIONS
   """
+
+- ICU's license follows:
+  From http://source.icu-project.org/repos/icu/icu/trunk/license.html
+  """
+  ICU License - ICU 1.8.1 and later
+
+    COPYRIGHT AND PERMISSION NOTICE
+
+    Copyright (c) 1995-2014 International Business Machines Corporation and others
+
+    All rights reserved.
+
+    Permission is hereby granted, free of charge, to any person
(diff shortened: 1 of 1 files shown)


srl295:
Maybe the build process could add some `tools/icu/ICU-LICENSE`  to the LICENSE file. (Or at that point, it could reformat `deps/icu/license.html` and be super up to date)

misterdjules:
Adding to the 0.11.15 milestone as installers will start shipping with ICU enabled from that milestone.

misterdjules:
Thank you! Two comments:
* The wording about the fact that the ICU license may or may not apply should be removed (as well as the corresponding explanation in the commit message). The `LICENSE` file is included and displayed as is in the package installers, and in that case it applies all the time.
* The newline at the end of the file was removed, could you please put it back?

Other than that, LGTM!

srl295:
@misterdjules fixed. I left the URL in place but can remove that also.

tjfontaine:
Longterm for improving the installers in general it would be nice if we broke out each license, but this LGTM

misterdjules:
@srl295 Thanks! Landed in 70d04e7f35667381aa6e9d76f56ea962c019d5bc.

srl295:
@misterdjules can you add the i18n tag here also please? thanks

PR #9039, opened 2015-01-16
test: fix test-fs-access.js

### LINKED ISSUE #9033: test/simple/test-fs-access.js fails most of the time
For instance: http://jenkins.nodejs.org/job/nodejs-v0.12/DESTCPU=ia32,label=ubuntu-12.04/lastBuild/tapTestReport/simple.tap-196/.

It seems it passes all the time on Windows, and sometimes passes on other platforms though.

misterdjules:
On non-windows supported platforms, fs.access(readOnlyFile, W_OK, ...)
is expected to fail, but always succeeds if node runs as the super user,
which is often the case for tests running on our continuous integration
platform.

This change makes the test try to change its process user id to nobody
on non-windows platforms so that the above mentioned test can pass and
still perform the actual desired test. If changing the process user id
to a nobody is not possible, then the test checks that
fs.access(readOnlyFile, W_OK, ...) actually succeeds.

Fixes #9033.

### DIFF
--- a/test/simple/test-fs-access.js
+++ b/test/simple/test-fs-access.js
@@ -25,8 +25,9 @@ var fs = require('fs');
 var path = require('path');
 var doesNotExist = __filename + '__this_should_not_exist';
 var readOnlyFile = path.join(common.tmpDir, 'read_only_file');
+var readWriteFile = path.join(common.tmpDir, 'read_write_file');
 
-var removeFile = function(file) {
+function removeFile(file) {
   try {
     fs.unlinkSync(file);
   } catch (err) {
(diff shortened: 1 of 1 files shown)


misterdjules:
Please note that I would have liked to be able to change the process' user id only for the failing test. I started to do it and it , but it turned out that would need a major refactoring of `test-fs-access.js`. I think it may not be worth the effort, and this change is in my opinion at least good enough for now.

misterdjules:
@cjihrig If you could take a look at it, that would be great!

cjihrig:
@misterdjules LGTM, although I didn't run the test. If it's passing now, that's great. I think if it starts encountering the same problem again, we should just opt out of the read only file check when running as root.

tjfontaine:
This also looks good to me

misterdjules:
@cjihrig @tjfontaine Thanks! Landed in 28d0cbb.

PR #9048, opened 2015-01-16
http: rename flush to flushHeaders

tjfontaine:
In order to preserve the potential for a flush method being added to the
streams API, rename flush to flushHeaders which is much more clear about
the behavior of this method.


misterdjules:
LGTM.

tjfontaine:
thanks, landed in 89f3c90

PR #9049, opened 2015-01-16
test: fix test-debug-port-from-cmdline.js

misterdjules:
Make this test less prone to race conditions by using synchronous
interprocess communication instead of a timer to determine when the
child process is ready to receive messages from its parent.

Also, remove a superfluous timer since the tests suite already makes
tests time out after a while.


tjfontaine:
lgtm

misterdjules:
@tjfontaine Thanks! Landed in 7a8ea15d6fbbf14ee1aef50714ae254f2130b9b5.

Issue #9052, opened 2015-01-17
deps: update openssl to 1.0.1l


jasnell:
The openssl dependency needs to be updated. 
This landed previously is io.js here: 
https://github.com/iojs/io.js/commit/eebdf7ac85b36223833330ff7f1057591dcbe651
https://github.com/iojs/io.js/commit/b910613792dac946b295855963869933a9089044
https://github.com/iojs/io.js/commit/ced41b07dd6ebeaceb2a9be1d2bb6376eb7d0db4

I am working on applying this to v0.12. Will submit a PR for review once I'm sure it at least builds and passes the initial sniff-test

There's also this:

https://github.com/iojs/io.js/commit/a76811c811f8e1602181f1c4368370c192adff0c

But it touches a test that currently does not exist in v0.12 (/test-crypto-stream.js)

jasnell:
Ok, located /test-crypto-stream.js ... in io.js it's been moved out of simple.

jasnell:
See: https://github.com/jasnell/node/tree/v0.12-openssl-k

misterdjules:
Thanks @jasnell! 

Usually changes are landed in the most stable or oldest active branch where they apply. In this case it would be v0.10. Moreover, that would be a good candidate to include for the next 0.10 release (0.10.36).

Also, instead of floating the build fix for windows by landing https://github.com/iojs/io.js/commit/ced41b07dd6ebeaceb2a9be1d2bb6376eb7d0db4, we could first land the changes from io.js you mentioned except this one, and then upgrade from OpenSSL 1.0.1k to OpenSSL 1.0.1l which is supposed to include the build fix for Windows.

How does that sound?

jasnell:
Ok... so just so I'm clear: apply the patch to branch v0.10 (without ced41b0) and resubmit?

steakknife:
Latest OpenSSL is 1.0.1l

misterdjules:
@jasnell Yes, and also apply the upgrade from 1.0.1k to 1.0.1l in a separate commit if it actually fixes the Windows build for Node.js and more generally doesn't bring anything in that would potentially break. I haven't looked at the diff, it's possible that the code changes are the same as https://github.com/iojs/io.js/commit/ced41b07dd6ebeaceb2a9be1d2bb6376eb7d0db4.

steakknife:
[...]

PR #9054, opened 2015-01-17
deps: update openssl to 1.0.1k

jasnell:
re: #9052

Applied @indutny's openssl update (see https://github.com/iojs/io.js/commit/eebdf7ac85b36223833330ff7f1057591dcbe651) to v0.12.

There was one difference in the location of the test-crypto-stream.js test but everything
else merged in with no obvious errors. Builds clean with zero test failures on macos 10.10.
I have not yet tested on the other platforms.


indutny:
LGTM

misterdjules:
@jasnell Thanks! See [my comments](https://github.com/joyent/node/issues/9052#issuecomment-70349222) in #9052.

Issue #9000, opened 2015-01-09
[Security] OpenSSL 1.0.1j -> 1.0.1k


steakknife:
http://packetstormsecurity.com/files/129867/secadv_20150108.txt

steakknife:
1.0.1l was released yesterday, so node is now 2 revs behind.

steakknife:
Will close after it's merged into 0.10.x branch/es at least

misterdjules:
Thank you @steakknife and sorry for not getting back to you earlier. I will close this as a duplicate of #9052 if you don't mind, because #9052 has a bit more info on how the upgrade will be done, and not having duplicates helps us track things and avoids confusion.

steakknife:
@misterdjules Over a week with no response on a security issue.  Is security@joyent.com valid for node?

misterdjules:
@steakknife The actual email is security@nodejs.org. I'm sorry this information was not easily available when you submitted this issue. It should be after https://github.com/joyent/node-website/pull/60 is merged.

steakknife:
So a safe bet would be security@nodejs.org cc: report@nodesecurity.io

GPG .asc URL?

misterdjules:
That seems good. What do you mean by "GPG .asc URL?"? @tjfontaine Could you please confirm.

steakknife:
Probably simpler to negotiate with a named individual email person to send encrypted confidential info w/ GPG.  Having a shared GPG key can be a chore, initially.  A reminder to not send confidential info in the clear might be wise.

steakknife:
@misterdjules Thanks for clarifications.  Happy Friday or weekend as the case may be. :)

PR #9056, opened 2015-01-17
deps: update openssl to 1.0.1l

jasnell:
Updates v0.10 openssl to 1.0.1l, walking through @indutny's 1.0.1k patch.

See: #9052 , #9000


jasnell:
Replaces #9054

steakknife:
:metal:  Hopefully there won't be yet another Heartbleed 0d this month.

jasnell:
Let's hope there are no more Heartbleed 0d's any month.

misterdjules:
@jasnell Thanks! The fix for the key press requirement on Windows is present twice as two different commits,  can we have it just once?

misterdjules:
@jasnell I would also prefer to have the fix to `test/simple/test-crypto-stream.js` in a separate commit if you don't mind. It doesn't change much when we consider only this branch (v0.10), but because node's crypto implementation is much different in v0.12 (and possibly in other branches in the future), being able to land these commits separately in other branches can be handy.

misterdjules:
@jasnell The x64 build fails on Windows with the following error message:
```
MASM : fatal error A1000: cannot open file : d:\jenkins\workspace\node-review-windows-julien\1d92f228\deps\openssl\openssl\crypto\bn\asm\x86_64-win32-masm.asm [d:\jenkins\workspace\node-review-windows-julien\1d92f228\deps\openssl\openssl.vcxproj]
d:\jenkins\workspace\node-review-windows-julien\1d92f228\deps\openssl\openssl.targets(28,5): error MSB3721: The command "call ml64.exe "/Zi" "/Fo" "Release\obj\openssl\x86_64-win32-masm.obj" "/c" "d:\jenkins\workspace\node-review-windows-julien\1d92f228\deps\openssl\openssl\crypto\bn\asm\x86_64-win32-masm.asm"" exited with code 1. [d:\jenkins\workspace\node-review-windows-julien\1d92f228\deps\openssl\openssl.vcxproj]
```

More details here: http://jenkins.nodejs.org/job/node-review-windows-julien/DESTCPU=x64,label=windows/lastBuild/console.

Do you have some bandwidth to take a look into it? I have a Windows x64 build setup, so if you don't have one I can look into it too. Just let me know what works best for you.

jasnell:
@misterdjules Unfortunately, I do not have ready access to a Windows machine this week.

jasnell:
The keypress fix was an odd one. Because of the order in which commits were made in io.js, I had to apply the fix once, apply their 1.0.1k patch, then apply it again. We could squash the first one possibly.

misterdjules:
@jasnell No problem, I'll look into the Windows x64 build issue, I'll let you know how it goes.

misterdjules:
@jasnell Regarding the Windows build issue, it is fixed by https://github.com/iojs/io.js/commit/7c4a50dd2fc8ca7325292ef0cca99295a39cc6da.

As for applying https://github.com/jasnell/node/commit/b07594feb1a84cb77f2749e530e5b915f1d78b3a, I don't think we actually need to apply it even once, since the fix is already present in the v0.10 branch. 

What I did locally is that I used `git --git-dir=../io.js/.git format-patch -k -1 --stdout eebdf7ac85b36223833330ff7f1057591dcbe651 | git am  -k -` to apply eebdf7ac85b36223833330ff7f1057591dcbe651. 

It couldn't apply as is and I add to apply it by using `git apply --reject /Users/JulienGilli/dev/node/.git/rebase-apply/patch` and I left the only rejected change out which would have basically reverted https://github.com/jasnell/node/commit/b07594feb1a84cb77f2749e530e5b915f1d78b3a. Finally I added all changed files (plus a newly added one) with `git add -u deps/openssl` and then committed the change with `git apply --continue`.

The resulting set of commits is available in PR #9063. Please let me know what you think!

misterdjules:
@jasnell Closing this one in favor of #9063, please let me know if that's not what you had in mind.

PR #9060, opened 2015-01-19
fs: disallow overwrite F_OK/R_OK/W_OK/X_OK

JacksonTian:


trevnorris:
Thanks. Landed w/ a test in 4823afcbe2

Issue #8589, opened 2014-10-21
V8 SIGILL on v0.12 on Raspberry Pi


sxa:
I believe this is different from the issue mentioned in #8549 so I'm raising this separately.
There are two issues with the 0.12 build on ARM at the moment that prevent it from running correctly (although it builds without issues)

Firstly, there is a crash that is introduced when vfpv3 is used as arm_fpu (this is the default in the 0.12 configure because apparently V8 doesn't support vfpv2 which is the default in 0.10). However if I compile with it set to something else like "fpu" or "vfpv2" it does compile and start, although that's likely not the best fix ... The comment about V8 not supporting v2 is here in configure:

  o['variables']['arm_fpu'] = 'vfpv3'  # V8 3.18 no longer supports VFP2.

Second issue is that while it does compile or start with the above change on older kernel versions, for more recent ones (anything "current") it fails because V8 is doing CPU detection based on /proc/cpuinfo which changed as a result of this delta in the Linux kernel:
https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/commit/arch/arm/kernel/setup.c?id=b4b8f770eb10a1bccaf8aa0ec1956e2dd7ed1e0a

The change modified the line on which (v6l) shows up from "Processor" to "model name". Here is the output on my "old" kernel 3.2.27+:

sxa@raspberrypi ~ $ head -2 /proc/cpuinfo
Processor   : ARMv6-compatible processor rev 7 (v6l)
BogoMIPS    : 697.95

Here it is on a newer one (3.12.22+)

sxa@pi ~ $ head -2 /proc/cpuinfo
processor   : 0
model name  : ARMv6-compatible processor rev 7 (v6l)

Line 370 of https://github.com/joyent/node/blob/master/deps/v8/src/cpu.cc is scanning for the line prefixed "Processor" to be able to set the CPU flags correctly. Modifying the code to check for both options allows Node to execute on both old and new kernels:

```
if (architecture_ == 7) {
  char* processor = cpu_info.ExtractField("Processor");
  if (HasListItem(processor, "(v6l)")) {
    architecture_ = 6;
  } else {
    delete[] processor;
    processor = cpu_info.ExtractField("model name");
    if (HasListItem(processor, "(v6l)")) {
      architecture_ = 6;
    }
  }  
  delete[] processor;
}
```

[...]

Issue #7222, opened 2014-03-02
Build breakage on ARMv6 devices with v0.11 branch


adammw:
I've just been [pinged on Twitter](https://twitter.com/sageahn/status/439056429010272256) that the [v0.11.9 build here](https://gist.github.com/adammw/3245130) throws an illegal instruction on ARMv6 devices (basically the Raspberry Pi, I haven't tested any other devices). Seems to do the same on all of the ["offical" builds](http://nodejs.org/dist/) past v0.11.8.

Running `node --v8-options` presents:

```
target arm v6 vfp2 hard
ARMv7=1 VFP3=0 VFP32DREGS=0 NEON=0 SUDIV=0 UNALIGNED_ACCESSES=1 MOVW_MOVT_IMMEDIATE_LOADS=0 USE_EABI_HARDFLOAT=1
```

On v0.11.7:

```
target arm v6 vfp2 hard
ARMv7=0 VFP3=0 VFP32DREGS=0 NEON=0 SUDIV=0 UNALIGNED_ACCESSES=0 MOVW_MOVT_IMMEDIATE_LOADS=0 USE_EABI_HARDFLOAT=1
```

Obviously, an armv6 shouldn't be setting the ARMv7 flag, and that's causing illegal instructions to be generated.

I've been able to bisect it to something being changed between v0.11.7 and v0.11.8, but I haven't yet found the commit or change that needs reverting.

wilson0x4d:
Cross-posted from another thread:

I also had previously edited `configure` to force the use of "vfpv2" instead of "vfpv3" by changing:

`o['variables']['arm_fpu'] = 'vfpv3'  # V8 3.18 no longer supports VFP2.`
to
`o['variables']['arm_fpu'] = 'vfpv2'  # V8 3.18 no longer supports VFP2.`

In addition to the armv6 change to `cpu.cc`. 

Since "vfpv2" isn't considered supported in "v8" I have not generated a patch (I'm really not sure what the correct action is wrt the configure script tbh.) My limited understanding is that 'vfp' are floating point extensions, and my assumption here is that by specifying `vfpv2` the node build is omitting said extensions (as the `VFP3=0` from --v8-options suggests.) Thus, someone will probably want to generate an appropriate patch to `configure`, or perhaps give me some insight into what scenarios it is acceptable (or unacceptable) to opt out of VFPv3.

kevinhikaruevans:
Was this patched in master?

[...]

PR #9068, opened 2015-01-21
v8: workaround to arm v6l detection in newer kernel versions as described in Issue 8589

wilson0x4d:
v8: workaround to arm v6l detection in newer kernel versions as described in Issue 8589

See: https://github.com/iojs/io.js/issues/283
See: https://github.com/joyent/node/issues/7222
See: https://github.com/joyent/node/issues/8589
See: https://code.google.com/p/v8/issues/detail?id=3112

Also reflected in v8 pull: https://github.com/v8/v8/pull/18

### DIFF
--- a/deps/v8/src/base/cpu.cc
+++ b/deps/v8/src/base/cpu.cc
@@ -374,7 +374,20 @@ CPU::CPU() : stepping_(0),
       char* processor = cpu_info.ExtractField("Processor");
       if (HasListItem(processor, "(v6l)")) {
         architecture_ = 6;
-      }
+      } else {
+        // Due to changes in Linux kernel, there is a scenario
+        // where the "Processor" field no longer contains the
+        // processor architecture. The workaround for this issue
+        // is to inspect the "model name" field.
+        //
+        // See: https://github.com/joyent/node/issues/8589
+        // See: https://code.google.com/p/v8/issues/detail?id=3112
+        delete[] processor;
+        processor = cpu_info.ExtractField("model name");
+        if (HasListItem(processor, "(v6l)")) {
+          architecture_ = 6;
+        }
+      }  
       delete[] processor;
     }
   }
--- a/deps/v8/src/base/cpu.cc
+++ b/deps/v8/src/base/cpu.cc
@@ -365,25 +365,23 @@ CPU::CPU() : stepping_(0),
     //
     // See http://code.google.com/p/android/issues/detail?id=10812
     //
-    // We try to correct this by looking at the 'elf_format'
+    // We try to correct this by looking at the 'elf_platform'
     // field reported by the 'Processor' field, which is of the
     // form of "(v7l)" for an ARMv7-based CPU, and "(v6l)" for
     // an ARMv6-one. For example, the Raspberry Pi is one popular
     // ARMv6 device that reports architecture 7.
+	// NOTE: the 'elf_platform' moved to the model name field in 
+	// Linux v3.8, so we check that first
     if (architecture_ == 7) {
-      char* processor = cpu_info.ExtractField("Processor");
+      char* processor = cpu_info.ExtractField("model name");
       if (HasListItem(processor, "(v6l)")) {
         architecture_ = 6;
       } else {
-        // Due to changes in Linux kernel, there is a scenario
-        // where the "Processor" field no longer contains the
-        // processor architecture. The workaround for this issue
-        // is to inspect the "model name" field.
-        //
-        // See: https://github.com/joyent/node/issues/8589
-        // See: https://code.google.com/p/v8/issues/detail?id=3112
+        // prior to Linux v3.8 'elf_platform' was found in the
+        // "Processor" field, we check this second for backward 
+        //compatibility
         delete[] processor;
-        processor = cpu_info.ExtractField("model name");
+        processor = cpu_info.ExtractField("Processor");
         if (HasListItem(processor, "(v6l)")) {
           architecture_ = 6;
         }
--- a/configure
+++ b/configure
@@ -490,20 +490,19 @@ def configure_arm(o):
   else:
     arm_float_abi = 'default'
 
-  if is_arch_armv7():
-    o['variables']['arm_version'] = '7'
-    o['variables']['arm_fpu'] = 'vfpv3'
-    o['variables']['arm_version'] = '7'  
-  elif is_arch_armv6():
-    o['variables']['arm_version'] = '6'
-    o['variables']['arm_fpu'] = 'vfpv2'
-  else:
-    o['variables']['arm_version'] = 'default'
-    o['variables']['arm_fpu'] = 'vfp'
-    
+  # configure arm defaults which would be most widely supported
   o['variables']['arm_neon'] = int(is_arm_neon())
   o['variables']['arm_thumb'] = 0      # -marm
   o['variables']['arm_float_abi'] = arm_float_abi
+  o['variables']['arm_version'] = 'default'
+  o['variables']['arm_fpu'] = 'vfpv2' # according to 'armcc', ARMv1 is syn. with ARMv2
+
+  # implement arch-specific overrides
+  if is_arch_armv7():
+    o['variables']['arm_version'] = '7'
+    o['variables']['arm_fpu'] = 'vfpv3'
+  elif is_arch_armv6():
+    o['variables']['arm_version'] = '6'
 
 
 def configure_node(o):
--- a/configure
+++ b/configure
@@ -492,12 +492,15 @@ def configure_arm(o):
 
   if is_arch_armv7():
     o['variables']['arm_version'] = '7'
+    o['variables']['arm_fpu'] = 'vfpv3'
+    o['variables']['arm_version'] = '7'  
   elif is_arch_armv6():
     o['variables']['arm_version'] = '6'
+    o['variables']['arm_fpu'] = 'vfpv2'
   else:
     o['variables']['arm_version'] = 'default'
-
-  o['variables']['arm_fpu'] = 'vfpv3'  # V8 3.18 no longer supports VFP2.
+    o['variables']['arm_fpu'] = 'vfp'
+    
   o['variables']['arm_neon'] = int(is_arm_neon())
   o['variables']['arm_thumb'] = 0      # -marm
   o['variables']['arm_float_abi'] = arm_float_abi


silverwind:
@wilson0x4d did the compiled binary work for you? I wasn't having any luck with a very similar patch so far, with either node or iojs (they SIGILL on movw/movt instructions). Also I'd be interested on which kernel and model you tested.

wilson0x4d:
yes, and no.

I also had previously edited `configure` to force the use of `vfpv2` instead of `vfpv3' by changing:

`o['variables']['arm_fpu'] = 'vfpv3'  # V8 3.18 no longer supports VFP2.`
to
`o['variables']['arm_fpu'] = 'vfpv2'  # V8 3.18 no longer supports VFP2.`

In addition to the armv6 change to cpu.cc.. Since "vfpv2" isn't considered supported in "v8" I have not generated a patch (I'm really not sure what the correct action is wrt the configure script tbh, but my limited understanding is that 'vfp' are floating point extensions, and my assumption here is that by specifying `vfpv2` the node build is omitting said extensions (or, rather, simply NOT including vfpv3, which is incompatible with ARMv6.) Thus, someone will probably want to generate an appropriate patch to `configure`, or perhaps give me some insight into what scenarios it is acceptable (or unacceptable) to opt out of VFPv3.

I have seen a similar patch floating around that checked for v7 before interrogating 'model name', this patch is taken from @sxa555's snippet which simply determines a failure to read `v6l` from "Processor" field and then blindly attempt to pull 'v6l' from the "model name" field. it avoids problems where a version OTHER than v7 is being detected (for whatever reason.)

After updating local source to detect armv6 I am able to run node successfully again, here is my build command-line (for completeness):

`sudo make clean; sudo ./configure --without-snapshot; sudo make; sudo make install`

and also of interest:

`me@mypi ~/node-build $ uname -a`
`Linux mypi 3.12.35+ #730 PREEMPT Fri Dec 19 18:31:24 GMT 2014 armv6l GNU/Linux`
`me@mypi ~/node-build $ node --v8-options | more`
`target arm v6 vfp2 hard
ARMv7=0 VFP3=0 VFP32DREGS=0 NEON=0 SUDIV=0 UNALIGNED_ACCESSES=0 MOVW_MOVT_IMMEDIATE_LOADS=0 USE_EABI_HARDFLOAT
=1`

and lastly (within node):

`> this.process.version`
`'v0.13.0-pre'`

Hope this helps someone else, I probably burned a total of 4 hours getting node up and running again after upgrading my pi last night :)

wilson0x4d:
after some thought, I believe it would make sense to add a "--without-vfp3" option to `configure`

silverwind:
Great job @wilson0x4d, I've just compiled sucessfully with both patches from https://github.com/iojs/io.js/issues/283

wilson0x4d:
the only problem I have with the chromium fix applied to iojs.io is that it's still hardcoding a version check, it will necessarily break again with ARMv8.

that aside, would be nice to get similar patches pushed into joyent/node so we can get updated ARMv6 packages :) and I don't deserve all the credit, this was solved by multiple people, I simply gathered all the info available. I'm surprised nobody else fixed it sooner...

wilson0x4d:
additional commits based on those submitted to bnoordhuis/io.js

1. do not assume ARMv7 is the version we incorrectly detect in all cases
2. do not assume non-ARMv7 has VFPv2 support
3. assume ARMpre6 supports VFPv1

Maintainers should expect ARMv8 will require an additional conditional to `configure` in the future, and ideally no change to `cpu.cc`, based on these commits.

HTH someone else ;) I'll keep my fork live until there's a merge or equivalent fix put into place.

bnoordhuis:
@wilson0x4d I'll be happy to take a pull request.  I wonder though, does V8 still work on ARMv5 / with VFP < 2?  V8 generally only supports what AOSP and Chrome support.

wilson0x4d:
I'll be honest and note that I'm not sure (ARMv5 on v8), I found conflicting info on the net with respect to ARM cores, and do not have an ARMv5 based device to test. 

I based the conditional on info from wikipedia, which lists 2 of 3 official ARMv5 based cores as supporting VFPv2, and the other supporting VFP (no specific version that I could find). 

The VFPv2-compatible listing included boards/devices such as Atmel's Make Controller, Lego's Mindstorms NXT, "most" older Nokia devices, an unlisted number of automobiles, etc. The VFPv1 based ARMv5 chip appeared to only be in use as a microcontroller of sorts (as opposed to the main CPU for anything) for example video decoding on the Wii. I'm only really interested in Pi support, and perhaps a few older phones (which are not ARMv5), but it seemed appropriate to have an explicit conditional in the off chance an ARMv5 device was detected so that an ARMv6 architecture wasn't specified to any build tools. (for example, there's at least one guy attempting to build nodejs for the OLinuXino.)

As an aside, ARM's tool chain appears to treat vfpv1 and vfpv2 as synonymous, so it likely makes sense that armv5 detection use vfpv2 instead. However, **if Chromium doesn't support ARMv5 based architectures then perhaps the configure script should warn the user upon detecting ARMpre6** (but still allow the build to move forward in the off chance the end user knows something we don't, or is willing to make things work, Fx. using node/io on an older programmable board or robotics controller.)

I'll make a few more edits.

wilson0x4d:
i think this version of configure better conveys intent, where arch specific overrides can be added/modified as needed to ensure a proper build. if someone were to come back to this and edit it in the future they would be coerced into writing code that conveys their intent more clearly (e.g. for version X the tool chain requires Y.)

trevnorris:
@tjfontaine @misterdjules Thoughts on this?

deps/v8/src/base/cpu.cc
```diff
@@ -365,16 +365,27 @@ CPU::CPU() : stepping_(0),
...
-    // We try to correct this by looking at the 'elf_format'
+    // We try to correct this by looking at the 'elf_platform'
     // field reported by the 'Processor' field, which is of the
     // form of "(v7l)" for an ARMv7-based CPU, and "(v6l)" for
     // an ARMv6-one. For example, the Raspberry Pi is one popular
     // ARMv6 device that reports architecture 7.
+	// NOTE: the 'elf_platform' moved to the model name field in 
+	// Linux v3.8, so we check that first
     if (architecture_ == 7) {
-      char* processor = cpu_info.ExtractField("Processor");
```
misterdjules: Why is this change different than [the one landed upstream](https://codereview.chromium.org/867713003/diff/1/src/base/cpu.cc)?

misterdjules:
@trevnorris Besides my inline comment, the change seems to make sense. Thank you @wilson0x4d!

That would make us float another patch on top of V8 though, which makes me wonder how @jasnell's work on having V8 built from vanilla source code + patches queue is going :)

misterdjules:
@trevnorris @wilson0x4d Also, shouldn't we land that in v0.12 instead of master?

trevnorris:
@misterdjules It's possible that's just how the PR was created, instead of actually intended. I'm fine landing this on v0.12.

WarheadsSE:
There are very very few ARMv5 generation processors with VFP units, any they were all essetentially custom. You're far better assuming _no_ VFP then _any_ VFP below ARMv6. Even in ARMv6, it was an options, albeit a standard one. The only reason for is proliferation of code with vfpv2 is because of the RPi. I have several ARMv6 boards with no VFP at all. I do entirely agree that hard-coding is a Bad Thing &trade; but a proper functional solution does need found.

If you ever need test platforms, let me know. I have access to the full range of low-high ARMv5-ARMv7 boards and should soon have an ARMv8.

misterdjules:
@WarheadsSE Thanks for your input! It seems that this PR doesn't fix the vfp detection on all ARM platforms, but it seems it fixes a lot of issues with common platforms, and thus would bring some improvements while not causing any regression. I have the impression that we could land this PR, and then work on a more robust and broader ARM support later. What do you think?

@trevnorris I'm also fine on landing this, but I would rather split it into two commits as done in this branch: https://github.com/misterdjules/node/commits/fix-armv6-detection (see the top 2 commits). This preserves all the information on the V8 fix upstream and land the configure changes in a separate commit.

@wilson0x4d What do you think?

WarheadsSE:
@misterdjules Spot on with my meaning regarding detection. I know that a chunk of this is faults within v8, but I would say that it is safe to call ARMv6 vfpv2, since 90% of all ARMv6 boards running targeted toolchains (read not a generic set from random distro) are /sigh/ RPi's. 

Please keep in mind also, that ARMv7 does not mean neon, and often when there is neon, there might be vpfpv4! (This is how Cortex-A15/A7 pairs in big.LITTLE work)

I also agree with splitting the commits as you suggested.

misterdjules:
@WarheadsSE Excellent.

@wilson0x4d @WarheadsSE  Do you have some time to test that https://github.com/misterdjules/node/commits/fix-armv6-detection fixes the build on the ARM platforms it's supposed to fix (Raspberry PI B+, but probably others)?

Thank you!

WarheadsSE:
@silverwind can confirm that this does work. I know this, because he directed me to the chain of issues that leads here.

misterdjules:
@WarheadsSE Yes, I just want to make sure that the changes in my branch, which are slightly different than the changes in this PR, also work as expected. They _should_ work as they seem to be semantically equivalent, but I would like to be certain :)

silverwind:
@misterdjules: I can start off a build on my Pi for it. Does your branch include the v8 fix? (These builds usually take ~5 hours)

silverwind:
Nevermind, I see it's included. ETA on build 5 hours.

misterdjules:
@silverwind Thank you!

silverwind:
Got an error : `object of type 'filter' has no len()`, details here:

https://gist.github.com/silverwind/8fea5a87983b411b1d9e

Note that this is just `./configure && make`, no `--without-snapshot` which I'll try now.

silverwind:
Nevermind that error, I was on the wrong branch it seems.

silverwind:
@misterdjules build done, and it works!

misterdjules:
@silverwind That's great, thank you so much for your help! We'll land https://github.com/misterdjules/node/commits/fix-armv6-detection as soon as possible.

misterdjules:
Actually, we'll probably want to just integrate https://github.com/iojs/io.js/pull/559 since it's the exact same changes as my branch, plus the removal of `arm_neon`, which is effectively not used by the current V8 code in v0.12.

meshpoint:
Y U NO USE [UNAME](http://pubs.opengroup.org/onlinepubs/9699919799/functions/uname.html)?
```c
#include <stdio.h>
#include <sys/utsname.h>
int main () {
        struct utsname u;
        if (uname (&u)<0) {
                perror ("uname: ");
                return 1;
        }
        printf ("machine: %s\n", u.machine);
        return 0;
}
```
> machine: armv6l

misterdjules:
Closing in favor of https://github.com/joyent/node/pull/14193 as discussed previously. Hopefully we'll be able to release that soon. Maybe for the upcoming v0.12.2, probably for v0.12.3 at the latest.

wilson0x4d:
finally catching up on this thread, /agree /agree /agree.  code changes look great! thanks for the focus on this issue by everyone, hopefully this means we'll be able to deploy a 'vanilla' node to our devices soon. :)

misterdjules:
@wilson0x4d Thank you for all your help :+1:

magic890:
Here is the official patch from V8 team: https://github.com/v8/v8-git-mirror/blob/master/src/base/cpu.cc#L483-L499

I have created a pull request to solve this issue: #25625

Issue #9070, opened 2015-01-21
`url.format` strips query in 0.11.15


pierre-elie:
In `0.11.15`, `url.format(urlObj)` seems to remove `query` from `urlObj` and does not include it in the formatted url.

![capture d ecran 2015-01-21 a 12 08 42](https://cloud.githubusercontent.com/assets/5923751/5840873/a2b28d98-a166-11e4-8eec-55aab55788b4.png)

`0.11.14` does not have this issue.

misterdjules:
@pierre-elie Thank you for reporting this issue!

Just because it's a bit easier to copy/paste, here's a repro of this issue as text:
```
var url = require('url');
var assert = require('assert');

var parsed = url.parse('http://example.com', false);
parsed.query = {key: 'value'};

url.format(parsed, true);

assert.ok(parsed.query, 'query property should exist');
assert.equal(parsed.query.key, 'value', "query,key should equal 'value'");
```

The breaking change is d312b6d15c69cf4c438ed7d884e6396c481a57f6. @chrisdickinson Do you have some time to take a look at it?

chrisdickinson:
`path` overrides `query`, per the following diagram:

```
+-----------------------------------------------------------------------------------+
| href                                                                              |
+----------+--+-----------+--------------------+-----------------------------+------+
| protocol |* | auth      | host               | path                        | hash |
|          |  |           +-------------+------+----------+------------------+      |
|          |  |           | hostname    | port | pathname | search           |      |
|          |  |           |             |      |          +-+----------------+      |
|          |  |           |             |      |          | | query          |      |
" http:     //   user:pass@some.host.com:65535 /path/name ? search=1&continues#hash "
|          |  |           |             |      |          | |                |      |
+----------+--+-----------+-------------+------+----------+-+----------------+------+
(all spaces in the "" line should be ignored -- they're purely for formatting)
*: given by "slashes" key
```

`path` is to `query`, `search`, and `pathname` as `host` is to `hostname` and `port`.

chrisdickinson:
Closing this issue as "works as expected."

misterdjules:
I find it very confusing that [the code mentioned above](https://github.com/joyent/node/issues/9070#issuecomment-70918084) _silently_ swallows the `query` property.

[...]

PR #9071, opened 2015-01-21
Fix issue 9070

ds2co:
A !query was inserted into an if statement that was breaking the
url.format function and removing the query variables.


chrisdickinson:
The corresponding issue, #9070, is exhibiting expected behavior -- `path` overrides `query` and `pathname` when available (similar to `host` vs. `hostname` and `port`). Thanks for the quick patch, though!

PR #9076, opened 2015-01-22
http: verify http method is a valid http token

jasnell:
RFC 2616/7230 requires that http method conform to
the token rule. This adds a check to verify that there are
no invalid characters in the method. If there are, then
we throw, otherwise fall through and continue.

Replacement for https://github.com/joyent/node/pull/9017 based on
feedback.
Targeted at v0.10

(I have a version targeted at v0.12 just in case).

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -1344,6 +1344,39 @@ Agent.prototype.removeSocket = function(s, name, host, port, localAddress) {
 var globalAgent = new Agent();
 exports.globalAgent = globalAgent;
 
+// http token is defined in RFC2616 as...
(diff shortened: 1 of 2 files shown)


jasnell:
cc @joyent/node-coreteam ... can we get a review on this one? does it still make sense to target this to v0.10 or should I rebase it on v0.12?

trevnorris:
@jasnell v0.10 is basically in maintenance, so unless this is a bug fix I'd just rebase on v0.12.

jasnell:
+1... I suspected as much. Will rebase and resubmit.

PR #9081, opened 2015-01-22
url: throw error on mismatching keys

chrisdickinson:
Throw an error if path and any of pathname, query, or search conflict.

This is mostly for discussion at this point. I'm not entirely convinced it's a great idea to go this route.

Fixes: https://github.com/joyent/node/issues/9070

### DIFF
--- a/doc/api/url.markdown
+++ b/doc/api/url.markdown
@@ -9,11 +9,27 @@ Parsed URL objects have some or all of the following fields, depending on
 whether or not they exist in the URL string. Any parts that are not in the URL
 string will not be in the parsed object. Examples are shown for the URL
 
-`'http://user:pass@host.com:8080/p/a/t/h?query=string#hash'`
+`'http://user:pass@host.com:65535/p/a/t/h?query=string#hash'`:
+
+```
++-----------------------------------------------------------------------------------+
+| href                                                                              |
++----------+--+-----------+--------------------+-----------------------------+------+
+| protocol |* | auth      | host               | path                        | hash |
+|          |  |           +-------------+------+----------+------------------+      |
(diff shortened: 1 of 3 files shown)


misterdjules:
Adding it to the 0.11.16 milestone to make it easier to track.

misterdjules:
I am also not convinced that throwing errors, likely out of context of the actual error, is a good idea.

My general feeling is that https://github.com/joyent/node/pull/8755 fixes some minor API usability issues that node users learned and know how to deal with, but introduce significant and new ones.

I agree that the APIs for `url.format`, `url.parse` and the interoperability with `http(s).request` should be much better, but it seems that it's not a trivial task and that it needs more thoughts.

Because we are trying to close the latest remaining issues in node v0.12's release candidates, I would advocate for reverting the original breaking change, and working on a better API in node v0.13 or later.

What do you think?

jasnell:
This change isn't sitting well with me either. I'd rather find a better way of handling the mutation (and related errors) independently of the actual format. I can think of a few ways of doing it but they'd diverge fairly significantly from the current API

tjfontaine:
After lots of conversation around this concept -- I think I can summarize our where we landed as:

If we don't like the inconsistency of only having half this particular api work this way, and want to solve it in a better (potentially backward incompatible way in the future) then in the mean time we should revert the original path precedence patch -- and fix it for the world post v0.12.

misterdjules:
I would like to close this PR and fix #9070 with #9109 instead. Please let me know what you think!

misterdjules:
@chrisdickinson Closing in favor of https://github.com/joyent/node/issues/9117 as discussed recently. Please feel free to reopen if that doesn't summarize our discussions. Thank you!

PR #9088, opened 2015-01-24
src: improve tooling around AUTHORS file

misterdjules:
This is a first draft to replace tools/updateAuthors.awk with tools/update-authors.js. The new
tool generates an AUTHORS file that is stable-ordered alphanumerically
(or more precisely "according to unicode code points").

The new tool also generates a .mailmap file that is essentially the same
content as the AUTHORS file, but with all email addresses used by each
author when committing changes to the repo.

As it is currently, the following section:

```
# These people didn't contribute patches to node directly,
-# but we've landed their v8 patches in the node repository:
-Daniel Clifford <danno@chromium.org>
-Erik Corry <erik.corry@gmail.com>
-Jakob Kummerow <jkummerow@chromium.org>
-Kevin Millikin <kmillikin@chromium.org>
-Lasse R.H. Nielsen <lrn@chromium.org>
-Michael Starzinger <mstarzinger@chromium.org>
-Toon Verwaest <verwaest@chromium.org>
-Vyacheslav Egorov <vegorov@chromium.org>
-Yang Guo <yangguo@chromium.org>
```

is removed by this tool, do we want to keep it?

Also, here's the output that shows that the newly generated `.mailmap` and `AUTHORS` include all the contributors:

```
➜  v0.12 git:(fix-issue-9077) ✗ git log --format='%aN' | sort | less | uniq | wc -l
     691
➜  v0.12 git:(fix-issue-9077) ✗ wc -l AUTHORS .mailmap                             
     691 AUTHORS
     691 .mailmap
    1382 total
➜  v0.12 git:(fix-issue-9077) ✗
```

@tjfontaine I'm interested in getting initial feedback to know if that corresponds to what you had in mind. If so, then I'll implement the stable-sort part and I'll update the PR.


tools/update-authors.js
```diff
@@ -0,0 +1,146 @@
...
+/*
+ * This script updates the AUTHORS and .mailmap files so that:
+ * - The AUTHORS file contains a list of contributors' names ordered
+ *   alphabetically.
+ * - The .mailmap file contains the same list of contributors' names along with
+ *   all the email addresses they've used
+ *   in their commits.
+ */
+
+"use strict";
```
cjihrig: +1 for using strict mode. Small nit - single quotes are used throughout the rest of the file.

tools/update-authors.js
```diff
@@ -0,0 +1,146 @@
...
+
+var child_process = require('child_process');
+var fs = require('fs');
+var assert = require('assert');
+var util = require('util');
+
+var debug = require('util').debuglog('update-authors');
```
cjihrig: Since `util` was already imported, why not just use it here?

tools/update-authors.js
```diff
@@ -0,0 +1,146 @@
...
+      } else {
+        cmdLineOpts.authorsFile = cmdLineArgs[i];
+      }
+    } else if (cmdLineArgs[i] === '-m') {
+      if (!cmdLineArgs[++i]) {
+        usage();
+      } else {
+        cmdLineOpts.mailmapFile = cmdLineArgs[i];
+      }
+    } else if (cmdLineArgs[i]) {
```
cjihrig: Could this just be `else`?

rvagg:
feel free to use https://github.com/iojs/io.js/blob/v1.x/tools/update-authors.sh

we're still ordering by first-commit though (I don't see a good reason to order any other way)

misterdjules:
@rvagg Thanks for the heads-up!

One thing that I like with writing our tools as Node.js programs is that:
* They run on all platforms supported by node, including Windows, out of the box.
* We're doing more dogfooding, which we should do more in my opinion.

As for ordering alphanumerically instead of by first commit, the reason if I remember correctly is that it's even more "stable" across branches (with first-commit order, you could have the same set of authors but different AUTHORS files for two different branches), and in the end is simpler to reason about.

misterdjules:
@cjihrig Thank you for the preliminary review! I made changes according to your comments, and pushed a more robust and complete version.

tools/update-authors.js
misterdjules: Updated, thanks!

tools/update-authors.js
misterdjules: Indeed, thanks!

tools/update-authors.js
misterdjules: Yes, thank you!

misterdjules:
I ended up not including the automatic generation of a `.mailmap` file with this tool. Indeed, `.mailmap` entries are of different forms with different meaning.

When generating them automatically, it is easy to add bogus entries that further confuse `git log` when using formats that use the `.mailmap` file (`%aN` and `%aE`). I think that fixing automatically generated entries in the AUTHORS file by hand when they are not valid (and manually adding the entries in the `.mailmap` if necessary) is safer, because they never taint the output of git log.

tjfontaine:
I agree with generating the mailmap by hand, there are still some places where humans have better/more context. We can potentially improve the tooling like `git-apply-pr` to have an interactive mode or verification mode to manage the authors while committing as opposed to releasing.

All in all this is a great step in the maintainable direction -- thanks @misterdjules LGTM

misterdjules:
@tjfontaine Thanks! Landed in 31610d0db42b4f61f85841a0d58f5014c98d0fc5.

PR #9090, opened 2015-01-24
Adding wrappers for missing OpenSSL SSL methods.

cmuellner:
This patch adds support for the following selectable SSL/TLS
methods, which are provided by OpenSSL:
- TLSv1.1
- TLSv1.2
- DTLSv1.0


indutny:
This is a new feature and cannot be landed in 0.10, I'm afraid :(

trevnorris:
Ah. Completely missed it was for v0.10. Thanks.

@cmuellner Sorry. Thanks for taking the time to submit a PR.

Issue #8996, opened 2015-01-08
Publish tarball that contains only full ICU data


misterdjules:
Eventually, binary installers will ship with full ICU data (see #8995). However, other binaries (binary tarballs, custom builds from source) will still not ship and load the full ICU data by default.

To allow users of these other binaries to use the full ICU data, we should generate and publish a separate tarball that contains only the full ICU data as a separate `.dat` file.

To use the full ICU data, users could put this file in `$PREFIX/share/node/icu/` or could pass its path on the command line using `--icu-data-dir=`. 

The tarball would be automatically generated and published by a Jenkins job.

This depends on #8983.

cc @srl295 @tjfontaine

srl295:
You can pull the full data file out of `deps/icu/source/data/in/*.dat` after a successful build. However, it won't necessatily have the right endianness.

srl295:
Note: this is blocked by #9046 - we can't generate the big endian tarball for the same reason (conversion).  Probably the solution proposed there needs to go in first, IF we want to publish BE data.

srl295:
@misterdjules ^^ thoughts on #9091 ?

srl295:
FYI @sxa555 - how does this look?

srl295:
- [ ] add a `README` to the tarball with some instructions. maube lodge it in `tools/icu/README-data-tarballs.md`

srl295:
- [ ] add a command line to `test-intl.js` to verify full
- [ ] add another test to upload the tarball, download it and test it.

srl295:
@misterdjules where should the extended tests ^^ live?

sxa555:
Seems good - subject to an assumption that Windows users will be ok with a .tgz instead of a ZIP.

srl295:
@sxa555 yeah wasn't sure about that. Make a .zip also? make them all zips?

misterdjules:
@srl295 Thank you very much for your work, and sorry for taking so long to do a first review.

It's not clear to me yet how to add tests for that, since I believe we don't currently have any similar "integration" test.

I would put them in a separate directory in `test`, maybe in `test/integration` (I don't like this name, but `test/i18n` might be too specific). 

The test doesn't need to upload the tarball, it could just build node with `small-icu` support, generate the data tarballs, and then run node such that it loads the data in this tarball and run a small script that tests for full localization support.

We would also want to have that work on all supported platforms (OS X, Linux, SmartOS and Windows).

What do you think?

sxa555:
[...]

PR #9091, opened 2015-01-24
create full-icu locale tarballs

srl295:
Work in progress on joyent/node#8996

Generates:
- 9.8M    node-v0.11.16-locales-be.tar.gz
- 9.6M    node-v0.11.16-locales-le.tar.gz

```
$ tar tvfpz node-v0.11.16-locales-be.tar.gz 
-rw-r--r--  0 srl    staff 25335456 Jan 24 10:45 node-v0.11.16/icudt54b.dat
$ tar tvfpz node-v0.11.16-locales-le.tar.gz 
-rw-r--r--  0 srl    staff 25335456 Jan 24 10:45 node-v0.11.16/icudt54l.dat
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -268,6 +268,16 @@ BINARYNAME=$(TARNAME)-$(PLATFORM)-$(ARCH)
 BINARYTAR=$(BINARYNAME).tar.gz
 PKG=out/$(TARNAME).pkg
 PACKAGEMAKER ?= /Developer/Applications/Utilities/PackageMaker.app/Contents/MacOS/PackageMaker
+LEDATATAR=$(TARNAME)-locales-le.tar
+BEDATATAR=$(TARNAME)-locales-be.tar
+LEDATATGZ=$(LEDATATAR).gz
+BEDATATGZ=$(BEDATATAR).gz
+#ICU_DATA_IN=$(firstword $(wildcard deps/icu/source/data/in/icudt*l.dat))
+#ICU_DATA_SRC=out/Release/gen
+ICU_DATA_SRC=deps/icu/source/data/in
+ICU_DATA_IN=$(firstword $(wildcard $(ICU_DATA_SRC)/icudt*[bl].dat))
+ICU_DATA_BE=$(ICU_DATA_IN:$(ICU_DATA_SRC)/icudt%l.dat=icudt%b.dat)
+ICU_DATA_LE=$(ICU_DATA_IN:$(ICU_DATA_SRC)/icudt%l.dat=icudt%l.dat)
 
 PKGSRC=nodejs-$(DESTCPU)-$(RAWVER).tgz
 ifdef NIGHTLY
@@ -441,3 +451,19 @@ cpplint:
 lint: jslint cpplint
 
 .PHONY: lint cpplint jslint bench clean docopen docclean doc dist distclean check uninstall install install-includes install-bin all staticlib dynamiclib test test-all test-addons build-addons website-upload pkg blog blogclean tar binary release-only bench-http-simple bench-idle bench-all bench bench-misc bench-array bench-buffer bench-net bench-http bench-fs bench-tls
+
+## Build data tarballs
+data-tarballs: $(LEDATATGZ) $(BEDATATGZ)
+
+$(LEDATATGZ) $(BEDATATGZ): node
+#	rm -rf $(BINARYNAME) out/deps out/Release
+#	$(PYTHON) ./configure --download=all --with-intl=full-icu
+	rm -rf $(TARNAME)
+	mkdir $(TARNAME)
+	out/Release/icupkg -tb $(ICU_DATA_IN) $(TARNAME)/$(ICU_DATA_BE)
+	out/Release/icupkg -tl $(ICU_DATA_IN) $(TARNAME)/$(ICU_DATA_LE)
+	tar -cf $(BEDATATAR) $(TARNAME)/$(ICU_DATA_BE)
+	gzip -f -9 $(BEDATATAR)
+	tar -cf $(LEDATATAR) $(TARNAME)/$(ICU_DATA_LE)
+	gzip -f -9 $(LEDATATAR)
+	rm -rf $(TARNAME) $(LEDATATAR) $(BEDATATAR)
--- a/tools/icu/icu-generic.gyp
+++ b/tools/icu/icu-generic.gyp
@@ -19,6 +19,10 @@
       'direct_dependent_settings': {
         'defines': [
           'UCONFIG_NO_CONVERSION=1',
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -441,3 +451,19 @@ cpplint:
 lint: jslint cpplint
 
 .PHONY: lint cpplint jslint bench clean docopen docclean doc dist distclean check uninstall install install-includes install-bin all staticlib dynamiclib test test-all test-addons build-addons website-upload pkg blog blogclean tar binary release-only bench-http-simple bench-idle bench-all bench bench-misc bench-array bench-buffer bench-net bench-http bench-fs bench-tls
+
+## Build data tarballs
+data-tarballs: $(LEDATATGZ) $(BEDATATGZ)
+
+$(LEDATATGZ) $(BEDATATGZ): node
```
misterdjules: Depending on how node has been built, `icupkg` may not have been built. Instead of depending on `node`, maybe we could make these two targets depend on `icupkg`? In the current state of the build system, an `icupkg` target  would depend on calling `./configure --with-intl=small-icu` first.

Makefile
```diff
@@ -441,3 +451,19 @@ cpplint:
...
+#	rm -rf $(BINARYNAME) out/deps out/Release
+#	$(PYTHON) ./configure --download=all --with-intl=full-icu
```
misterdjules: This is probably related to my comment above.

Makefile
```diff
@@ -441,3 +451,19 @@ cpplint:
...
+	rm -rf $(TARNAME)
```
misterdjules: I would not generate the localization data files in the same directory as where the source tarball's content is stored, just to avoid some interdependency between these two unrelated targets.

Makefile
```diff
@@ -268,6 +268,16 @@ BINARYNAME=$(TARNAME)-$(PLATFORM)-$(ARCH)
 BINARYTAR=$(BINARYNAME).tar.gz
 PKG=out/$(TARNAME).pkg
 PACKAGEMAKER ?= /Developer/Applications/Utilities/PackageMaker.app/Contents/MacOS/PackageMaker
+LEDATATAR=$(TARNAME)-locales-le.tar
+BEDATATAR=$(TARNAME)-locales-be.tar
+LEDATATGZ=$(LEDATATAR).gz
+BEDATATGZ=$(BEDATATAR).gz
+#ICU_DATA_IN=$(firstword $(wildcard deps/icu/source/data/in/icudt*l.dat))
+#ICU_DATA_SRC=out/Release/gen
+ICU_DATA_SRC=deps/icu/source/data/in
```
misterdjules: Do you think we could add a `ICU_DATA_IN` in the `config.mk` file that is generated from the `configure` script to avoid having it hardcoded in several different places?

Makefile
```diff
@@ -441,3 +451,19 @@ cpplint:
...
+
+## Build data tarballs
+data-tarballs: $(LEDATATGZ) $(BEDATATGZ)
+
+$(LEDATATGZ) $(BEDATATGZ): node
+#	rm -rf $(BINARYNAME) out/deps out/Release
+#	$(PYTHON) ./configure --download=all --with-intl=full-icu
+	rm -rf $(TARNAME)
+	mkdir $(TARNAME)
+	out/Release/icupkg -tb $(ICU_DATA_IN) $(TARNAME)/$(ICU_DATA_BE)
```
misterdjules: `$(ICU_DATA_IN)` needs to have been generated, which is not necessarily the case depending on how `node` has been built (with or without icu support). What do you think about having these targets depending on `icpukg` and `icudata` instead of on `node`?

Makefile
```diff
@@ -268,6 +268,16 @@ BINARYNAME=$(TARNAME)-$(PLATFORM)-$(ARCH)
 BINARYTAR=$(BINARYNAME).tar.gz
 PKG=out/$(TARNAME).pkg
 PACKAGEMAKER ?= /Developer/Applications/Utilities/PackageMaker.app/Contents/MacOS/PackageMaker
+LEDATATAR=$(TARNAME)-locales-le.tar
+BEDATATAR=$(TARNAME)-locales-be.tar
+LEDATATGZ=$(LEDATATAR).gz
+BEDATATGZ=$(BEDATATAR).gz
+#ICU_DATA_IN=$(firstword $(wildcard deps/icu/source/data/in/icudt*l.dat))
```
misterdjules: I would remove all comments from this PR if possible, otherwise it makes it difficult for reviewers to know if commented out code should be reviewed or not.

tools/icu/icu-generic.gyp
```diff
@@ -30,12 +34,8 @@
       'toolsets': [ 'host', 'target' ],
       'direct_dependent_settings': {
         'defines': [
-          'UCONFIG_NO_LEGACY_CONVERSION=1',
-          'UCONFIG_NO_IDNA=1',
-          'UCONFIG_NO_TRANSLITERATION=1',
-          'UCONFIG_NO_SERVICE=1',
-          'UCONFIG_NO_REGULAR_EXPRESSIONS=1',
           'U_ENABLE_DYLOAD=0',
+          'UCONFIG_NO_SERVICE=1',
```
misterdjules: I don't remember exactly what is the goal of these changes. Is it to save space in the target executable and in memory?

Makefile
```diff
@@ -268,6 +268,16 @@ BINARYNAME=$(TARNAME)-$(PLATFORM)-$(ARCH)
...
+#ICU_DATA_SRC=out/Release/gen
+ICU_DATA_SRC=deps/icu/source/data/in
+ICU_DATA_IN=$(firstword $(wildcard $(ICU_DATA_SRC)/icudt*[bl].dat))
```
misterdjules: @tjfontaine Do we generally allow using GNU make extensions in the `Makefile`?

tjfontaine: I think there are a lot of places already where we assume GNU make, fairly sure gyp will use gmake on the bsds and sunos based systems

Makefile
srl295: icupkg isn't available in some configurations. maybe better to have this code run `configure&&make` directly.

misterdjules: But it is used in the rule that builds these targets, so it must be available or am I missing something?

misterdjules:
@srl295 Did you have some time to take a look at my comments?

tools/icu/icu-generic.gyp
srl295: Yes, space, memory and I think there were build dependency issues also.

srl295:
@misterdjules I thought the npm was the direction we wanted to go with this, as per #8996

jasnell:
@srl295 ... does this need to stay open?

srl295:
No it can close, this pr isn't the way forward.

jasnell:
Awesome. Ok

PR #9096, opened 2015-01-26
Corrected a grammar mistake in Stream documentation

3674823674252:
This morning I was reading through the documentation on Streams and came across this sentence, which seemed not valid:

"If you supply as the second argument to the it will be passed to push method,.."

So I corrected it to:

"If you supply output as the second argument to the callback, it will be passed to push method,.."

That is the only change in this pull request.


jasnell:
Closing this, it was addressed in a different way in nodejs/node

PR #9098, opened 2015-01-26
WIP: src: make build pass with GCC < 4.5

### LINKED ISSUE #9079: 0.11.15 binary linked against GLIBCXX_3.4.15 breaks centos 6.5 default support
It said gcc 4.2 or above is enough to compile the source, however the published binary linked against GLIBCXX_3.4.15 which is not included in gcc 4.4.7 shipped with Centos 6.5. (Might also breaks CentOS 6.6 as it's also shipped with gcc 4.4.7)
Is it intended? Will this also be true in 0.12 binary?
As CentOS 6.5 is still widely used in production environment, this requires compiling & install newer gcc or node by hand, which adds extra complexity in production environments. Is it possible to add some workaround about this?

misterdjules:
Building node with GCC > 4.4 on CentOS makes the node binary depend on a
more recent version of the C/C++ runtime that is not installed by
default on these older CentOS platforms, and probably on other platforms
as well.

Building node with the default gcc and g++ compilers that come with
these older versions of CentOS allows to ship a node binary that runs
out of the box on these setups with older C/C++ runtimes.

This change works around a bug that was fixed in GCC 4.5. Versions of
GCC < 4.5 would not support using the injected-class-name of a
template base class as a type name.

This change also fixes a lot of strict-aliasing warnings due to type
casts in the src/queue.h headers.

Fixes #9079.

### DIFF
--- a/deps/debugger-agent/debugger-agent.gyp
+++ b/deps/debugger-agent/debugger-agent.gyp
@@ -17,6 +17,14 @@
         "include",
       ],
     },
+    'conditions': [
+      [ 'gcc_version<=44', {
+        # GCC versions <= 4.4 do not handle the aliasing in the queue
+        # implementation, so disable aliasing on these platforms
+        # to avoid subtle bugs
+        'cflags': [ '-fno-strict-aliasing' ],
+      }],
+    ],
     "sources": [
       "src/agent.cc",
     ],
--- a/deps/uv/uv.gyp
+++ b/deps/uv/uv.gyp
@@ -84,6 +84,12 @@
         'src/version.c'
       ],
       'conditions': [
+        [ 'gcc_version<=44', {
+          # GCC versions <= 4.4 do not handle the aliasing in the queue
+          # implementation, so disable aliasing on these platforms
+          # to avoid subtle bugs
+          'cflags': [ '-fno-strict-aliasing' ],
+        }],
         [ 'OS=="win"', {
           'defines': [
             '_WIN32_WINNT=0x0600',
--- a/node.gyp
+++ b/node.gyp
@@ -168,6 +168,12 @@
       ],
 
       'conditions': [
+        [ 'gcc_version<=44', {
+          # GCC versions <= 4.4 do not handle the aliasing in the queue
+          # implementation, so disable aliasing on these platforms
+          # to avoid subtle bugs
+          'cflags': [ '-fno-strict-aliasing' ],
+        }],
         [ 'v8_enable_i18n_support==1', {
           'defines': [ 'NODE_HAVE_I18N_SUPPORT=1' ],
           'dependencies': [
--- a/src/cares_wrap.cc
+++ b/src/cares_wrap.cc
@@ -73,7 +73,9 @@ class GetAddrInfoReqWrap : public ReqWrap<uv_getaddrinfo_t> {
 
 GetAddrInfoReqWrap::GetAddrInfoReqWrap(Environment* env,
                                        Local<Object> req_wrap_obj)
-    : ReqWrap(env, req_wrap_obj, AsyncWrap::PROVIDER_GETADDRINFOREQWRAP) {
+    : ReqWrap<uv_getaddrinfo_t>(env,
+                                req_wrap_obj,
+                                AsyncWrap::PROVIDER_GETADDRINFOREQWRAP) {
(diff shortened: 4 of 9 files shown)


misterdjules:
This is a first draft, as there are different ways to solve this problem and I'm looking to gather your feedback.

The fix for the "stric-aliasing" warnings in `src/queue.h` is more or less a revert of joyent/libuv@0520464 that was cherry-picked in joyent/node with d29fe0f2c667085ec403c4856efca8686c52ceef.
I'm not familiar with the inappropriate uses of the QUEUE API mentioned in the original commit message, so I would be happy if @bnoordhuis could give us some context.

Another way to fix the warnings would of course be to pass the `-fno-strict-aliasing` option to gcc and g++, [similar to what had been been done in the past](30b29d8f8745263475d67576c4b81ff164482676) or to [what is done currently on OSX for node](https://github.com/joyent/node/blob/v0.12/common.gypi#L218-L220).

This change also doesn't take care of the queue implementation in libuv, but it would also need to be fixed.

Regarding the changes that work around GCC's bug with injected class name of a template base class as a type name, it makes the code less maintainable as it duplicates the template parameter. There are various ways to mitigate that:

1) Use a typedef to write the template parameter once and use this new type instead at both places of the fully qualified template name twice. This way, if the template parameter needs to change, it only needs to be updated in one place.

2) Get rid of most of the additional sub-classes of `ReqWrap` that were created but which implementation are empty (like `GetAddrInfoReqWrap `) and instead directly create instances of `ReqWrap`. For sub-classes of ReqWrap that don't have an empty implementation, use solution 1.

@tjfontaine @trevnorris @bnoordhuis @chrisdickinson @cjihrig I'm looking forward to reading your comments.

trevnorris:
The fundamentals of the fix LGTM. Though @bnoordhuis would be the best person for this review.

bnoordhuis:
> I'm not familiar with the inappropriate uses of the QUEUE API mentioned in the original commit message, so I would be happy if @bnoordhuis could give us some context.

The original is essentially untyped.  The macros evaluate to a `void*` pointer, which in C is implicitly coercible to almost any other pointer type.  It was too easy to use in a wrong way, so I made it a little stronger typed.

If you want to fix up queue.h, you can do better than this.  Compare QUEUE_DATA() with ContainerOf(): the former is an untyped C macro, the latter is a strongly typed C++ template function.  You can do a similar thing for the QUEUE data structure.

For background, I copied queue.h from libuv because I wanted a quick tailq implementation.  It stuck because most people working on the C++ bits were familiar with it.

misterdjules:
Thank you @bnoordhuis for the info!

I took another approach with https://github.com/misterdjules/node/commit/a541b9257eab1c924a1139f2ac8f761612e22dde. 

Basically, making my changes to `queue.h` type safe would require more thoughts and more time than what I have before the release of v0.11.16, and I don't want to land them if they make the queue implementation less safe.

Instead, I propose to pass `-fno-strict-aliasing` to gcc and g++ when gcc version is <= 4.4 as a temporary solution. This change would also require floating a patch for libuv. Later, we would make the queue implementation compile without `-fno-strict-aliasing` while being as safe as the current implementation.

I ran benchmarks (bench-http, bench-fs and bench-tls, not bench-net as it was hanging and I wanted to get some results quickly) on a CentOS 6.6 VM with 4GB of RAM and 4 CPUs. It seems that [the binary compiled with `-fno-strict-aliasing`](https://gist.github.com/misterdjules/b87e78faa52186fa4e19) doesn't have any major performance regression compared to [the one without](https://gist.github.com/misterdjules/7cccd0747542113d8e23).

The benchmarks' output is not easy to compare with separate files. I just saw how to run benchmarks and output compared results with `benchmark/compare.js`, I'll post the output here as soon as possible.

misterdjules:
The output of `benchmark/compare.js` is now [available as a gist](https://gist.github.com/misterdjules/f0faf024488aa3a7f883).

tjfontaine:
this LGTM, thanks!

misterdjules:
Landed in 5926526, tests pass on all platforms, and [more specifically on the CentOS 5.7 agent _without_ loading a more recent libstdc++](http://jenkins.nodejs.org/job/node-review-unix-julien/36/).

Thank you!

PR #9104, opened 2015-01-27
Documentation update about Buffer initialization

### LINKED ISSUE #7230: new Buffer() can contain any content
When creating a new Buffer using "new Buffer(size)", the buffer can contain any data. I was assuming the buffer to contain just zeroes, but this assumption is not correct.

If this is behaviour is intended, e.g. because of performance concerns, please update the documentation to reflect the fact that the returned buffer may contain random values.

sarathms:
fixes #7230

### DIFF
--- a/doc/api/buffer.markdown
+++ b/doc/api/buffer.markdown
@@ -64,7 +64,7 @@ It can be constructed in a variety of ways.
 
 Allocates a new buffer of `size` octets. Note, `size` must be no more than
 [kMaxLength](smalloc.html#smalloc_smalloc_kmaxlength). Otherwise, a `RangeError`
-will be thrown here.
+will be thrown here. Unlike `ArrayBuffers`, the underlying memory for buffers is not initialized. So the contents of a newly created `Buffer` is unknown. Use `buf.fill(0)`to initialize a buffer to zeroes.
 
 ### new Buffer(array)
 


doc/api/buffer.markdown
```diff
@@ -64,7 +64,7 @@ It can be constructed in a variety of ways.
 
 Allocates a new buffer of `size` octets. Note, `size` must be no more than
 [kMaxLength](smalloc.html#smalloc_smalloc_kmaxlength). Otherwise, a `RangeError`
-will be thrown here.
+will be thrown here. Unlike `ArrayBuffers`, the underlying memory for buffers is not initialized. So the contents of a newly created `Buffer` is unknown. Use `buf.fill(0)`to initialize a buffer to zeroes.
```
cjihrig: I would just say:

The underlying memory of a `Buffer` is not initialized. If this behavior is desired, use `Buffer.prototype.fill()`.

sarathms: I added the 'Unlike ArrayBuffers... ' part as suggested by another core dev. I think it helps. If its the length, I can paraphrase it further. But the `Buffer.protoype.fill()` usage is inconsistent with rest of the documentation. Its gonna take a while to change all references to match it. What do you advice?

cjihrig: I don't see a real need to mention `ArrayBuffer`s at all, but I'm not strongly opposed to it. I guess it makes more sense to stick with the existing style of the docs.

sarathms: How about this?
>Unlike, the underlying memory of a Buffer is not initialized. Use 
`buf.fill(0)` to initialize a buffer to zeroes.
There's a comparison with ArrayBuffer in the explanation for `#slice` too. I'd say let's keep it.

cjihrig: @tjfontaine thoughts?

trevnorris: Please wrap at 80 chars.

trevnorris:
One style nit, but LGTM otherwise.

jasnell:
LGTM

jasnell:
Landed in 70efdf3

PR #9086, opened 2015-01-24
deps: upgrade npm to 2.3.0

othiym23:
As discussed on #9078, here is `npm@2.3.0`. `make test-npm` ran without issue on OS X Yosemite.

r=@misterdjules

PR #9107, opened 2015-01-28
test: make test-npm uses custom node, not global

misterdjules:
When running make test-npm, make would use the node binary available in
the PATH, which would most of the time not be the same binary as the one
built from the source tree from where the make test-npm command in run.

This can be confusing, as it can lead users to think that they tested
npm with the version of node that was built from the current checkout of
the source tree when it would actually run the tests with a completely
different version.

This change modifies the PATH environment variable for all commands that
need to run the node binary to run npm's tests by adding the root of the
local checkout as the first entry, so that the custom built node is
always used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,12 +154,12 @@ test-npm: node
 	cd deps/npm ; npm_config_cache="$(shell pwd)/npm-cache" \
 	     npm_config_prefix="$(shell pwd)/npm-prefix" \
 	     npm_config_tmp="$(shell pwd)/npm-tmp" \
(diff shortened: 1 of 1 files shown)


misterdjules:
This [happened recently](https://github.com/joyent/node/pull/9086#issuecomment-71723055) when testing an upgrade to a newer version of npm.

cjihrig:
LGTM if it works

misterdjules:
Landed in 03b4f44.

misterdjules:
Thank you @cjihrig!

PR #9115, opened 2015-01-29
net: make Server.prototype.unref() persistent

cjihrig:
Currently, the `unref()` method does not remember any state if called before the server's handle has been created. This commit adds state to track calls to `ref()` and `unref()`.

Related to #7077. Replaces #8966


jasnell:
@cjihrig ... at this point do you still want to land this in v0.10?

cjihrig:
Nah, it landed in the converged repo.

PR #9116, opened 2015-01-29
console: allow Object.prototype fields as labels

### LINKED ISSUE #9069: console.timeEnd('hasOwnProperty') fails hilariously
Because the timers are currently stored as named properties of an object, timers with funny labels will result in `console.timeEnd` dutifully logging `NaNms`.

This affects `hasOwnProperty`, `__proto__` and likely anything else that lives on `Object.prototype`.

A solution would be using ES6 symbols when they land in node.

A workaround would be prefixing the property names with some arbitrary string (except `_`, which would re-create the problem for timers labelled `_proto__`) instead of using the label verbatim.

cjihrig:
`Console.prototype.timeEnd()` returns `NaN` if the timer label corresponds to a property on `Object.prototype`. This commit uses `Object.create(null)` to construct the `_times` object. This is not a proper fix for 0.10 because the older version of v8 includes a `__proto__` field. Closes #9069

### DIFF
--- a/lib/console.js
+++ b/lib/console.js
@@ -40,7 +40,7 @@ function Console(stdout, stderr) {
   Object.defineProperty(this, '_stdout', prop);
   prop.value = stderr;
   Object.defineProperty(this, '_stderr', prop);
-  prop.value = {};
+  prop.value = Object.create(null);
   Object.defineProperty(this, '_times', prop);
 
   // bind the prototype functions to this Console instance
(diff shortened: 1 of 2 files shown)


trevnorris:
LGTM

cjihrig:
Landed in 6c3647c38d73f729ce85633a0440cd939d93dea2

PR #9146, opened 2015-02-05
asyncwrap: fix constructor condition for early ret

trevnorris:
AsyncWrap should always properly propagate asynchronous calls to any
child that is created. Regardless whether kCallInitHook is currently
active. The previous logic would always return early if kCallInitHook
wasn't set.

R=@tjfontaine


misterdjules:
Adding to the 0.12.1 milestone as discussed.

trevnorris:
Landed on 605329d7f77.

PR #9153, opened 2015-02-06
Fixes compile error if configured without ssl

dmitry-yudakov:
If node is configured with **--without-ssl** _node_crypto.h_ is not included in _node.cc_, 
but variables declared in it are still used - it leads to compile error.

This change disables **--enable-ssl2** and **--enable-ssl3** options for given configuration.


jasnell:
Closing as duplicate of https://github.com/joyent/node/pull/25862. Can reopen if that is incorrect.

Issue #9156, opened 2015-02-06
Debugger: vm.runInDebugContext change behavior in 0.11.16


3y3:
``` js
var vm = require('vm');
var assert = require('assert');

var proto = vm.runInDebugContext('DebugCommandProcessor.prototype');
proto.option_ = true;
assert(proto.option_ === true);
proto = vm.runInDebugContext('DebugCommandProcessor.prototype');
assert(proto.option_ === true);
proto = vm.runInDebugContext('DebugCommandProcessor.prototype.option_ = true; DebugCommandProcessor.prototype');
assert(proto.option_ === true);
```

This test passes in 0.11.14 and fails in 0.11.16.
I think this is relative to 947408b025c3429b6c80901749e7866e04ec3fe0
@indutny , please confirm this.

Previous behavior is very important for node-inspector. This is my current way to extend v8 debugger protocol. If new behavior is correct, can you propose any way to extend debugger protocol after 0.11.16 and in io.js

indutny:
Looking

indutny:
This is not caused by 947408b for sure, happens on a2a3fd4 too.

indutny:
I think it was broken during the merge 7ca5af8

indutny:
Hm... may be not.

indutny:
Btw, doesn't seem to be working on v0.11.14 for me.

indutny:
And on v0.11.15 too.

indutny:
Same on v0.11.16 ... @3y3 am I missing something? This feature was introduced in v0.11.14 and your test case does not appear to be working with any versions after (and including) it.

3y3:
Hm... Win x64 works on 0.11.14 for me (downloaded from site).

indutny:
@tjfontaine @misterdjules can we confirm this on windows? (I don't have any machines at hand)

3y3:
I also backported this feature to previous versions in `v8-debug` project (small core of node-inspector) [link](https://github.com/node-inspector/v8-debug/blob/master/src/debug.cc#L36-L48) and it works fine excluding to 0.11.15+

indutny:
@3y3 it is failing on:

```
var proto = vm.runInDebugContext('DebugCommandProcessor.prototype');
proto.option_ = true;
```

for me

indutny:
```
~/Code/joyent/node $ ./node -p "process.version"
v0.11.16
~/Code/joyent/node $ cat 1.js
[...]

PR #9157, opened 2015-02-06
contextify: share security token with debug ctx

indutny:
If security token does not match - no changes are allowed to the objects
from different context. We already copy the security token to the newly
created contexts in contextify.cc . Copy the security token to the debug
context too.

Fix: https://github.com/joyent/node/issues/9156


indutny:
cc @tjfontaine @misterdjules

indutny:
Wait... This does not fix the issue. Oh gosh.

jasnell:
@indutny ... what would you like to do with this one?

indutny:
Close it ;)

PR #9161, opened 2015-02-07
fix fs.truncate when first arg is an int fd

bjouhier:
Bug is rather obvious and fix is trivial (see diff below).

### DIFF
--- a/test/simple/test-regress-GH-9161.js
+++ b/test/simple/test-regress-GH-9161.js
@@ -1,0 +1,45 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+var common = require('../common');
+var assert = require('assert');
+var path = require('path');
+var fs = require('fs');
+var tmp = common.tmpDir;
+if (!fs.existsSync(tmp)) 
+  fs.mkdirSync(tmp);
+var filename = path.resolve(tmp, 'truncate-file.txt');
+
+var success = 0;
+
+fs.writeFileSync(filename, 'hello world', 'utf8');
+var fd = fs.openSync(filename, 'r+');
+fs.truncate(fd, 5, function(err) {
+  assert.ok(!err);
+  success++;
+});
+
+process.on('exit', function() {
+  fs.closeSync(fd);
+  fs.unlinkSync(filename);
+  assert.equal(success, 1);
+  console.log('ok');
+});
--- a/doc/api/fs.markdown
+++ b/doc/api/fs.markdown
@@ -107,6 +107,8 @@ Synchronous ftruncate(2).
 
 Asynchronous truncate(2). No arguments other than a possible exception are
 given to the completion callback.
+Currently, fs.truncate also accepts a file descriptor as first argument but 
+you should use fs.ftruncate in this case as this fallback may go away.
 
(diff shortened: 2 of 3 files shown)


cjihrig:
Can you add a regression test please.

bjouhier:
@cjihrig Getting late I'll do it tomorrow. Where should I put it? I guess it is `test/simple/test-regress-GH-9161` but I am not sure.

bjouhier:
@cjihrig I added the test.

test/simple/test-regress-GH-9161.js
```diff
@@ -0,0 +1,53 @@
...
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+var common = require('../common');
+var assert = require('assert');
+var path = require('path');
+var fs = require('fs');
+var tmp = common.tmpDir;
+if (!fs.existsSync(tmp)) fs.mkdirSync(tmp);
```
cjihrig: Can you put the `fs.mkdirSync(tmp);` on a new line.

test/simple/test-regress-GH-9161.js
```diff
@@ -0,0 +1,53 @@
...
+var filename = path.resolve(tmp, 'truncate-file.txt');
+
+var success = 0;
```
cjihrig: I think this and the `process.on('exit', ...)` can be removed.

test/simple/test-regress-GH-9161.js
```diff
@@ -0,0 +1,53 @@
...
+
+fs.writeFile(filename, 'hello world', 'utf8', function(err) {
```
cjihrig: Other than the `fs.truncate()` call, can you change these all to the synchronous equivalents. You won't have to check for errors.

cjihrig:
I notice that this behavior doesn't seem to be documented. Perhaps it should be added there.

test/simple/test-regress-GH-9161.js
bjouhier: done

test/simple/test-regress-GH-9161.js
bjouhier: Exit test is necessary. Otherwise exits without any diagnostics on the broken implementation.

test/simple/test-regress-GH-9161.js
bjouhier: done

bjouhier:
Regarding the doc issue, I have the impression that this is compatibility code. People should use `ftruncate` instead. Do you document compat hooks?

test/simple/test-regress-GH-9161.js
```diff
@@ -0,0 +1,46 @@
...
+if (!fs.existsSync(tmp)) 
+  fs.mkdirSync(tmp);
+var filename = path.resolve(tmp, 'truncate-file.txt');
+
+var success = 0;
+
+fs.writeFileSync(filename, 'hello world', 'utf8');
+var fd = fs.openSync(filename, 'r+');
+fs.truncate(fd, 5, function(err) {
+  if (err) throw err;
```
cjihrig: This can just be `assert.equal(err, null);` (or `undefined` depending what the success case is).

test/simple/test-regress-GH-9161.js
```diff
@@ -0,0 +1,46 @@
...
+  fs.closeSync(fd);
+  var result = fs.readFileSync(filename, 'utf8');
+  assert.equal(result, 'hello');
+  success++;
+});
+
+process.on('exit', function() {
+  assert.equal(success, 1);
```
cjihrig: I still don't think it's necessary to check for success. However, one thing you can add here is cleanup of the temp file used by the test.

cjihrig:
@tjfontaine should this behavior be documented?

test/simple/test-regress-GH-9161.js
bjouhier: The bug is that the callback is replaced by an instance of FSReqWrap. So no callback is called (and no exception is thrown). So I need another way to detect that something went wrong. Verifying the number of assertions at exit time does the job.

bjouhier: The test file that I provided 1) reproduces the bug, 2) produces a clear message that there is a bug and 3) passes when the bug is fixed. The alternative that you are proposing misses 2) because the test completes without any message.

cjihrig: Sorry, my mistake was assuming that `ftruncate()` would throw when its callback isn't a function.

bjouhier: Yep. I was actually surprised to not get an exception.

tjfontaine:
Yes, I think we should document this behavior, for v0.10 and v0.12. We can decide if we want to deprecate/remove this fallback in master at a later date.

I agree that this fix is the the best way to handle the bug for v0.12.

For master though, I think we want to fix `makeCallback`:

 * the first check should be `ifNullOrUndefined` because the caller explicitly didn't pass in an argument
 * then the next check should be `!isFunction` which imo should outright throw, because invalid input was provided.

This would have made there result much more crisp and clear about what went wrong.

bjouhier:
@tjfontaine I agree.  I can handle the doc issue if you want but `makeCallback` should probably be handled as a separate issue.

cjihrig:
@bjouhier can you address the last code nit, and update the docs.

bjouhier:
@cjihrig I simplified the unit test, improved `makeCallback`, added a note to the doc and squashed the commits.

cjihrig:
Thanks! Landed in master in 4c31cda272ec8bb7e6ecef8a3ae01d4589e007a8 with a few tweaks.

cjihrig:
And landed in 0.12 in b3aa876f08563aaed1b51a9dc5c1a254a146af67 without the change to `makeCallback()`.

PR #9171, opened 2015-02-09
timers test: unref does not return this

jscissr:
The unref() call returns undefined, not this.
The test already worked before, because the interval was still unref'd,
and the test also succeeds without clearing the interval.

### DIFF
--- a/test/simple/test-timers-unref.js
+++ b/test/simple/test-timers-unref.js
@@ -42,7 +42,8 @@ setTimeout(function() {
 interval = setInterval(function() {
   unref_interval = true;
   clearInterval(interval);
-}, SHORT_TIME).unref();
+}, SHORT_TIME);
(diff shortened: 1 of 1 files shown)


cjihrig:
LGTM

tjfontaine:
LGTM, though the commit message would be slightly better as just `test: timers-unref ...`

jscissr:
Can/should I do something to fix the commit message?

cjihrig:
@jscissr you can do a `git commit --amend` to change the commit message and then `git push --force` it. Otherwise, it can be changed during landing.

jscissr:
I actually only used the github.com webinterface to change and commit, so I think the first option wouldn't work.

trevnorris:
Thanks. Landed in d2b546187a86c3.

PR #9178, opened 2015-02-10
Update build.mk

sdenel:
Debug flag should not be used in prod.


KenanY:
You are editing the source of one of the deps? Send this [upstream](https://github.com/piscisaureus/cares)!

sdenel:
Thank you KenanY, I just did.

Issue #8988, opened 2015-01-07
node-fastcgi compatibility with node ≥ 0.11.8


ebraminio:
I've filed this bug on [node-fastcgi](https://github.com/fbbdev/node-fastcgi/issues/2) bug tracker also. I think it could be a more general compatibility issue for next release of node so I am reporting it here also:

First have a look at node-fastcgi and the bug I've filed [there](https://github.com/fbbdev/node-fastcgi/issues/2). Here is my configurations:

.lighttpd.conf:
```conf
server.document-root = "."
server.modules += ( "mod_fastcgi" )
server.port = 2020
fastcgi.server += (
  "/myapp" => ((
    "socket" => "/tmp/myapp-fcgi.sock",
    "bin-path" => "/home/user/myapp/server.js",
    "check-local" => "disable",
    "max-procs" => 1,
  ))
)
```

[server.js](https://github.com/fbbdev/node-fastcgi#example):
```JavaScript
#!/home/user/.nvm/v0.11.8/bin/node

var fcgi = require('node-fastcgi');

fcgi.createServer(function(req, res) {
  if (req.method === 'GET') {
    res.writeHead(200, { 'Content-Type': 'text/plain' });
    res.end("It's working");
  } else {
    res.writeHead(501);
    res.end();
  }
}).listen();
```

Do `chmod +x server.js` and `npm install node-fastcgi` then `lighttpd -f .lighttpd.conf -D` and open http://127.0.0.1:2020/myapp on your browser. It works if you install 0.11.7 and change server.js to use it but for 0.11.8 and newer node versions, it just makes browser waiting forever and does nothing.

cjihrig:
Thanks for reporting this. Let's see what comes from the issue you opened on node-fastcgi. If that module needs to be updated, then it should happen there. If you have an example that doesn't contain any external dependencies and demonstrates a problem with Node core, then we can handle it here.

ebraminio:
Interestingly, by more compiling node/bisecting I've found that the bug is coming from libuv upgrade
* 613d76e works
* a35a212 doesn't compile
* 74a664b myapp is broken
So I think it is not just JSish toy code issue and I think this would be better to get more attention, (cc: @tjfontaine ) My naive guess is the issue is on handling unix sockets on newer version of libuv.

ebraminio:
[...]

PR #9179, opened 2015-02-10
deps: uv 1.4.0 update

trevnorris:
Update libuv to 1.4.0 and land a fix that changed default behavior to preserve backwards compatibility.

R=@tjfontaine @misterdjules


test/simple/test-child-process-sync-process-send.js
```diff
@@ -0,0 +1,44 @@
...
+
+if (process.argv[2] === 'child') {
+  process.send(big);
+  process.exit(42);
+}
+
+var proc = fork(__filename, ['child']);
+
+proc.on('message', common.mustCall(function(msg) {
+  assert.equal(typeof msg , 'string');
```
bnoordhuis: Style nit: space before comma. :-)

deps/uv/uv.gyp
```diff
@@ -84,12 +64,6 @@
         'src/version.c'
       ],
       'conditions': [
-        [ 'gcc_version<=44', {
-          # GCC versions <= 4.4 do not handle the aliasing in the queue
-          # implementation, so disable aliasing on these platforms
-          # to avoid subtle bugs
-          'cflags': [ '-fno-strict-aliasing' ],
```
misterdjules: The `-fno-strict-aliasing` flag was added to fix compilation warnings when building Node.js with GCC <= 4.4. See https://github.com/joyent/node/pull/9098 for more information. Although I don't like to float this patch, I think we may want to keep it for now and discuss how we could not have to float it (either by upstreaming, or changing the queue implementation).

misterdjules:
I’m not sure about the impact of [`* win, unix: add pipe_peername implementation`](https://github.com/libuv/libuv/pull/166) on backward compatibility. It's a new API that is not present in v0.12.0. Users could write native modules using this libuv API for node v0.12.1, but these modules would break on platforms where an older libuv 1.x is shipped as a dynamic library, and where Node.js is built to load libuv dynamically.

Are there platforms doing that currently, and if so what version of libuv do they track/ship?

misterdjules:
Just for the record, the build [fails on SmartOS currently](http://jenkins.nodejs.org/job/node-review-unix/DESTCPU=x64,label=smartos/lastBuild/console).

This build issue had been fixed by the fact that the tty related change had been reverted. At some point later, it was put back, but in a version of libuv > 1.02 if I remember correctly. I had sent [a PR that fixed the original breaking change](https://github.com/joyent/libuv/pull/1593), I'm not sure if it still applies though.

misterdjules:
I think we should also track an issue to add dtrace probes support back. In my opinion it's very useful for some users and I would like to at least track it somewhere. Thoughts?

test/simple/test-child-process-sync-process-send.js
trevnorris: heh. thanks. I just did a replace on the () that were there.

misterdjules:
@trevnorris The SmartOS build is fixed by https://github.com/misterdjules/node/commit/8ba154b47a97578ad5cbd4b50abaf227699bcb1d, and I reworded the header of https://github.com/joyent/node/commit/e0b44d688fab7f0fd48c3c7b97ee35409a92b2b2 into https://github.com/misterdjules/node/commit/137d3319f4b94e4b69758c136a88f22cedb294da.

tjfontaine:
### General Dependency Upgrades

Since Node.js doesn't currently define its own API/ABI the only way for users to build and ship binary addons is by building against the libuv and V8 that Node.js releases are built against (and in turn shipped). This also applies to c-ares, zlib, openssl, etc

We can upgrade any of our tightly coupled dependencies in a patch release, so long as those upgrades meet a couple requirements:

 * No public symbols that Node.js is linking against may be removed
 * All public symbols must behave in the way they did in the previous version, less any fixed bugs
  - bugs are generally described as fixed behavior that wasn't working as documented, or that the user couldn't have been explicitly/implicitly depending on
 * No breaking ABI changes for artifacts currently built and linked against the old version

Presuming the upgrade to the library meets those requirements, the only follow up to that is:

 * Node.js must not use any new symbols/APIs

Since we support/allow building and linking against shared versions of our dependencies, consumers may not upgrade them in sync, so the node binary (and also their binary addons) must continue to work with whatever version of the shared library they have installed.

### This Upgrade

From an API/ABI perspective most of this looks fine, save for the change to `uv_tty_set_mode`. I think in practice we're probably fine changing the `int` to `enum` but this is indeed one of those portions where the size of the `enum` is implementation defined, and can be one of those places where ABI is easily broken.

While there is a C++ level API compatibility change, what's not entirely clear is if the `int` to `enum` change is ABI compliant for people using only C.

I have not evaluated all the behavioral changes to know any further potential impacts.

misterdjules:
[Ran the node.js' tests with this upgrade](http://jenkins.nodejs.org/job/node-review-julien/) and [the fix for the SmartOS build](https://github.com/misterdjules/node/commit/8ba154b47a97578ad5cbd4b50abaf227699bcb1d), and so far on UNICEs all tests pass [except for one test on SmartOS](http://jenkins.nodejs.org/job/node-review-unix-julien/DESTCPU=ia32,label=smartos/lastBuild/tapTestReport/internet.tap-1/).

misterdjules:
It also seems to break a lot of tests [on Windows](http://jenkins.nodejs.org/job/node-review-windows-julien/DESTCPU=x64,label=windows/31/tapTestReport/). I haven't had the time to investigate those though.

trevnorris:
I might have missed a floating patch, though I thought that's what "lib: fix stdio/ipc sync i/o regression" was supposed to fix.

orangemocha:
Investigating those Windows failures....

orangemocha:
I haven't completed the investigation yet (I will resume on Monday), but I have determined that the second commit in this PR  (https://github.com/joyent/node/commit/42371f76472c38b91e1353f904be23284f6eb0c1) and not the libuv uddate is causing the WIndows child_process-related failures. 
cc @bnoordhuis

bnoordhuis:
See https://github.com/iojs/io.js/pull/774 - apparently blocking pipe I/O is not (always?) supported on Windows but to what extent is unclear to me.

orangemocha:
I have opened a [PR in libuv](https://github.com/libuv/libuv/pull/238) that addresses the blocking pipe issues on Windows. I think we'll have to float a patch on here. @trevnorris is it ok for me to add a commit to this PR?

trevnorris:
@orangemocha be my guest.

trevnorris:
@orangemocha Thanks for the PR, and great work! I've already cherry-picked your commit onto this PR.

@misterdjules This should be gtg for another test on Jenkins.

orangemocha:
Thanks! Let's see what Jenkins says: http://jenkins.nodejs.org/job/node-accept-pull-request/26/ (testing only)

piscisaureus:
@trevnorris You didn't check in hidden file (.mailmap and .gitignore). Was that deliberate?

trevnorris:
@piscisaureus cp must not grab hidden files by default. Are those files needed?

piscisaureus:
@trevnorris You're right, and they're not needed (but removing them is unnecessary churn).

misterdjules:
The build [still fails on SmartOS](http://jenkins.nodejs.org/job/node-test-commit-unix/22/DESTCPU=ia32,label=smartos/console). https://github.com/misterdjules/node/commit/8ba154b47a97578ad5cbd4b50abaf227699bcb1d, which is included in libuv 1.4.1, fixes it

orangemocha:
Well, at least the Windows failures are gone.

trevnorris:
@misterdjules thanks for the patch. I've landed that commit on this PR.

orangemocha:
There is still one test failing on OSX: http://jenkins.nodejs.org/job/node-test-commit-unix/22/DESTCPU=ia32,label=osx-build/tapTestReport/simple.tap-234/

I looked in Jenkins history, but I couldn't figure out if that's a flaky test or a regression introduced by this update.

orangemocha:
And there is still one test failing on SmartOS: http://jenkins.nodejs.org/view/node/job/node-test-commit-unix/25/DESTCPU=x64,label=smartos/tapTestReport/ even with @misterdjules ' commit.

misterdjules:
@orangemocha `test-fs-watch.js` failing looks like a regression to me, and it doesn't seem to be a flaky test (otherwise there would be `# TODO : Fix flaky test` in the output right?). `test-dgram-multicast-multi-process.js` on SmartOS also seems to be a regression, and doesn't seem to be a flaky test.

orangemocha:
The test is not *marked* as flaky (otherwise, the Jenkins build wouldn't fail even if the test failed). Still the test could be flaky and not have been marked as such, which is what I was wondering.
Anyway, since we cannot find any evidence of these tests failing before, we can assume that they are both regression.

trevnorris:
Sorry. You will have to follow up with SmartOS issues. Not my area. :)

misterdjules:
@trevnorris @orangemocha Yes, I'll look into the tests failing on SmartOS.

misterdjules:
`test-dgram-multicast-multi-process.js` fails on SmartOS 100%, I'm investigating why it's failing and how we can fix it. I can't reproduce the `test-fs-watch.js` timeout on my OS X setup, I'll check with @tjfontaine if I can get a shell on the osx-build agent to investigate more.

trevnorris:
@misterdjules thanks much for your assistance.

misterdjules:
@trevnorris @orangemocha https://gist.github.com/misterdjules/4db906cbcdd95b4e9cb5 seems to fix `test-dgram-multicast-multi-process.js` on SmartOS.

misterdjules:
Submitted [a PR upstream](https://github.com/libuv/libuv/pull/243) that fixes `test-dgram-multicast-multi-process.js` on SmartOS.

misterdjules:
@trevnorris @orangemocha I was able to reproduce `test-fs-watch.js` timing out with current v0.12, so it's not a regression. Also, another run [pass all tests on OSX](http://jenkins.nodejs.org/job/node-test-commit-unix/26/). We should create an issue to fix this test's flakiness though.

So the only thing that remains to be fixed before landing this is `test-dgram-multicast-multi-process.js` on SmartOS. I'll let you know when https://github.com/libuv/libuv/pull/243 is reviewed.

We might want to wait for it to land in a libuv release and upgrade libuv in node to this new release. I just want to avoid floating more patches if we can. What do you think?

saghul:
> We might want to wait for it to land in a libuv release and upgrade libuv in node to this new release. I just want to avoid floating more patches if we can. What do you think?

I wholeheartedly agree.

orangemocha:
@misterdjules given that those failures are understood, I concur that it's a good idea to unblock this update. But I think we should add this commit so that this can get through Jenkins and so that the CI remains healthy: https://github.com/joyent/node/commit/d8a5aaed0e0cb8b69ee033c040102342949d7671

trevnorris:
@misterdjules sounds good. I'll leave this open for now, but seems then we'll be upgrading to 1.4.3.

misterdjules:
@trevnorris @saghul Is there a plan for releasing libuv 1.4.3, and if so when?

@trevnorris @saghul @orangemocha If we can get https://github.com/joyent/node/commit/3c830c884759cb294c29e7e959f1f1584dcee6c6, https://github.com/joyent/node/commit/afcaea7890fbac9ede1276806827594ec59fe929 and https://github.com/libuv/libuv/pull/243 into libuv 1.4.3 before the end of next week, then I think it's worth it to upgrade to libuv 1.4.3 with just one floating patch rather than to libuv 1.4.0 with 4 floating patches.

What do you think?

misterdjules:
@orangemocha Regarding https://github.com/joyent/node/commit/d8a5aaed0e0cb8b69ee033c040102342949d7671, I agree for adding `test-fs-watch`. I would not add `test-dgram-multicast-multi-process` to the flaky tests though, as it's a 100% failure caused by a real bug in libuv that will break some users.

misterdjules:
I will spin up a Jenkins job to test the current libuv's 1.x branch with all commits that are in this PR + https://github.com/libuv/libuv/pull/243, so that we can have a better idea of how that would work. I'll keep you posted.

orangemocha:
@misterdjules 
>I would not add test-dgram-multicast-multi-process to the flaky tests though, as it's a 100% failure caused by a real bug in libuv that will break some users.

Then in principle you'd have to float the corresponding patch (or wait for the 1.4.3 update) or else Jenkins CI jobs will start failing because of test-dgram-multicast-multi-process. That is, assuming that we want to keep using Jenkins to land PRs.

misterdjules:
@orangemocha Right, we would either wait for libuv 1.4.3 or float the SmartOS patch for socket options.

misterdjules:
Ran tests for https://github.com/joyent/node/tree/libuv-upgrade-latest-1x, which is the current tip of libuv's v1.x branch with all commits in this PR + https://github.com/libuv/libuv/pull/243, and the build is all green, or rather blue, [on UNICEs](http://jenkins.nodejs.org/job/node-test-commit-unix/28/) and [on Windows](http://jenkins.nodejs.org/job/node-test-commit-windows/27/).

saghul:
> we would either wait for libuv 1.4.3 or float the SmartOS patch for socket options.

We added a new API, so next version will be 1.5.0.

misterdjules:
Ok, so if the next libuv release is 1.5.0 with new API(s) added, we might want to upgrade libuv incrementally to avoid another round of time consuming code reviews.

Thus, I would suggest to land this PR with the addition of https://github.com/joyent/node/commit/ed644e727db1557e24f09ef511a539c710a38bf2.

When libuv 1.5.0 is released, which will include all floating patches but https://github.com/joyent/node/commit/dbde261a71ed610f8d8f811c0a0475e7b0996e2c, we can submit another PR to upgrade to it and have a smaller set of changes to review.

@joyent/node-coreteam Thoughts?

orangemocha:
>Thus, I would suggest to land this PR with the addition of ed644e7.

+1 for this approach as all the work has already been done. One more floating patch won't be an issue especially since it will be included in the next update.

saghul:
@misterdjules Ok, but then we need a way to know that the bundled libuv is patched, so that bug reports don't become a mess and we avoid chasing ghosts. I suggest the following:

modify include/uv-version.h, set UV_VERSION_IS_RELEASE to 0 and set the UV_VERSION_SUFFIX to something like "node1". If another floating change is added, then you'd bump that number.

This way, process.versions would list libuv as being version "1.4.2-node1", thus avoiding confusion. This is similar to what Ubuntu does with its upstream Debian packages.

trevnorris:
Ref: https://github.com/joyent/node/issues/9342

orangemocha:
FYI, I marked test-fs-watch as flaky on OSX so it won't fail the Jenkins node-accept-pull-request job: https://github.com/joyent/node/commit/e8b21095670543ad6ee8e24e0a1b096b908e20b1

misterdjules:
Just for reference, this would fix #8988.

trevnorris:
@misterdjules I'm getting lost with all the floating patches. Mind helping me make sure we've got them all? I still need to upgrade uv to latest 1.4.x. also.

trevnorris:
Alright. Upgrade to uv 1.4.2 done. Can we get a Jenkins build to test this out?

orangemocha:
Jenkins build in progress: http://jenkins.nodejs.org/view/node/job/node-accept-pull-request/33/

orangemocha:
@trevnorris it looks like the update broke the OSX build:

http://jenkins.nodejs.org/job/node-test-commit-unix/33/DESTCPU=ia32,label=osx-build/console

libtool: unrecognized option `-static'

saghul:
@orangemocha unlikely. All static libraries fail to build, not only libuv. Looks like a problem on the build slave itself.

saghul:
@misterdjules @tjfontaine any comments on https://github.com/joyent/node/pull/9179#issuecomment-78442529?

trevnorris:
I could use some help on this one. I don't have OSX, and can't find a single change regards to a `-static` option.

trevnorris:
@saghul something like c515187

saghul:
> I could use some help on this one. I don't have OSX, and can't find a single change regards to a -static option.

It doesn't seem to be a libuv issue, but an issue on the OSX machine. All static libraries are failing to compile there.

saghul:
@trevnorris yeah, that will do, thanks a lot!

misterdjules:
@orangemocha @trevnorris @saghul The build issue on OSX was due to a problem in the build slave configuration, and I'm the one who broke it when fixing the libuv jenkins job. My apologies for that. It should be fixed now, but let me know if you still have issues with it. Sorry again!

misterdjules:
@saghul @trevnorris FWIW, I'm fine with @saghul's suggestion in https://github.com/joyent/node/pull/9179#issuecomment-78442529 and with @trevnorris' change.

misterdjules:
@orangemocha @trevnorris @saghul [All tests pass on supported unices](http://jenkins.nodejs.org/job/node-test-commit-unix/34/).

deps/uv/include/uv.h
```diff
@@ -138,6 +138,7 @@ extern "C" {
   XX(EOF, "end of file")                                                      \
   XX(ENXIO, "no such device or address")                                      \
   XX(EMLINK, "too many links")                                                \
+  XX(EHOSTDOWN, "host is down")                                               \
```
misterdjules: That seems like another API change that might be a bit more difficult not to use within the node code base, but at that point I don't know what other options we have.

trevnorris: agreed.

misterdjules:
@joyent/node-coreteam besides the API changes, including [the latest one in a 1.4.x patch release](https://github.com/joyent/node/pull/9179#discussion_r26696930), LGTM.

Do we have any option other than just landing these API changes in node? Any other way of moving forward I can think of requires a lot of maintenance from node for which we don't have proper tooling  right now.

saghul:
@misterdjules: That change is somewhere between a bugfix and an API change. libuv doesn't use the symbol but some OS-es generate it so we now export it. The alternative is an assert, which quite some people where getting while downloading packages with NPM, for example, which is not desired.

FWIW this particular change can be "ifdef tested".

misterdjules:
@trevnorris Assigned to you, let me know if you have any trouble running the node-accept-pull-request job on Jenkins. Thank you!

misterdjules:
@saghul OK, my concern was more that someone using `UV_EHOSTDOWN` with a given v1.4.x release will get a compilation error with another v1.4.x release that doesn't define this value, unless I'm missing something.

saghul:
@misterdjules that is indeed a possibility, unfortunately.

misterdjules:
Landed in 7a37910f25a9397ccfb2b4136273db803cb4345f, c0766eb1a4dac9d9b555be271ef84ad401ba91b0, 2411bea0dfa9d48f94f95119bbacce122509681e, 3cce8ab647ff50eb309e87d165f3bcb3eab62e50, 9613ac7cb8cf1c60516a8664f0f8abf7e3ade710, 2fdeb7e9327e84df657bc146fd73aa269d05dfc5 and a93087007fce8466ef58e37af76521468295ccec.

Thank you @trevnorrism @saghul, @orangemocha and everyone involved :+1:

Issue #9181, opened 2015-02-10
"FATAL ERROR: invalid array length Allocation failed - process out of memory" introduced in 0.11.15


SheetJSDev:
The js-xls test script passes in node <= 0.11.14 as well as in io.js but not in node 0.11.15 0.11.16 or 0.12

https://travis-ci.org/SheetJS/js-xls/builds/50244885 shows where the fails occur.  It works in io.js as well as in node 0.10 and versions up to 0.11.14 so I suspect that 0.11.15 changed Buffer behavior.

@trevnorris was there a fix in iojs that hasn't been applied to node?

trevnorris:
@SheetJSDev Can you post the URL to the exact test that's failing?

trevnorris:
I pushed a fix that was created by @bnoordhuis in 04b63e022a477e. Can you let me know if that fixes the issue?

SheetJSDev:
@trevnorris that did not resolve the issue, unfortunately :(  I pulled from the v0.12 branch which appears to have that commit.

The repo is https://github.com/SheetJS/js-xls and the travis commands are https://github.com/SheetJS/js-xls/blob/master/.travis.yml (warning: it downloads a ton of test files)

It's difficult to point to a specific test because each individual test case works on its own (the repo includes a CLI script xls.njs and that passes on each file, even in 0.11.15 and 0.12).  It just blows up after a certain number of tests.

trevnorris:
@SheetJSDev Thanks. I've been able to reproduce the issue. I'll do some testing and let you know what I find.

trevnorris:
Update: this is a race condition, but for what I can tell. If I run the tests using latest v0.12 debug build it doesn't have that issue.

On the flip side, it generated this error:
```
#
# Fatal error in ../deps/v8/src/heap/heap.cc, line 3324
# CHECK(!lo_space()->Contains(object)) failed
#

==== C stack trace ===============================

 1: V8_Fatal
 2: void v8::internal::Heap::RightTrimFixedArray<(v8::internal::Heap::InvocationMode)1>(v8::internal::FixedArrayBase*, int)
 3: _ZN2v88internal20FastElementsAccessorINS0_31FastHoleyObjectElementsAccessorENS0_18ElementsKindTraitsILNS0_12ElementsKindE3EEEE25SetLengthWithoutNormalizeENS0_6HandleINS0_14FixedArrayBaseEEENS7_INS0_7J
[...]

Issue #9147, opened 2015-02-06
Upstream fix to dictionary properties access to V8


misterdjules:
The [lastest update to mdb_v8 in v0.12](https://github.com/joyent/node/commit/4312f8d760bdf574b5a461e426331f0b4d298f18) fixed dictionary properties access. This commit touches `deps/v8/tools/gen-postmortem-metadata.py` with the following changes:
```
    'Map, instance_size, int, kInstanceSizeOffset',
     'Map, bit_field, char, kBitFieldOffset',
     'Map, bit_field2, char, kBitField2Offset',
-    'Map, bit_field3, SMI, kBitField3Offset',
+    'Map, bit_field3, int, kBitField3Offset',
     'Map, prototype, Object, kPrototypeOffset',
     'NameDictionaryShape, prefix_size, int, kPrefixSize',
     'NameDictionaryShape, entry_size, int, kEntrySize',
```
These changes should probably be upstreamed to V8.

misterdjules:
cc @davepacheco

jasnell:
Updated the priority on this to P-1 per https://github.com/nodejs/node/issues/28

misterdjules:
@davepacheco @jasnell Created https://codereview.chromium.org/1296743003.

misterdjules:
Done with https://github.com/nodejs/node/pull/2959 and https://codereview.chromium.org/1296743003.

Issue #9113, opened 2015-01-29
Bug: for loop becomes endless loop in 0.11.15


freiit:
The following one-liner runs in an endless loop on 0.11.15.
node --harmony --use-strict -e 'for (let i = 0; i < 3; ++i) { if (i == 1) { continue; } }'

Replace 'let' with 'var' to avoid the bug.

Bug does not show up in 0.11.14.

Tested with a clean docker image, created like this:
FROM node:0.11.15
CMD node --harmony --use-strict -e 'for (let i = 0; i < 3; ++i) { if (i == 1) { continue; } }'

cjihrig:
This is almost certainly related to the v8 3.28.73 update in 0.11.15.

trevnorris:
I bisected as far back as 3.28.10 and the issue still occurred.

freiit:
Can't reproduce that on 0.12.

Y--:
I do reproduce the bug on 0.12.0.

Y--:
FYI, it's fixed in iojs 1.6.1

misterdjules:
Adding to milestone 0.12.3 as it seems it would be fixed by #14411.

misterdjules:
Fixed in 2a5f4bd7ce40b9537c5d77558687858c7246d896 and cb55a05f528f57e350063f618274b5a182261387, thank you!

Issue #9097, opened 2015-01-26
[regression] NodeOS don't boot with Node.js v0.11.15 (musl conflict?)


piranna:
At shown at https://github.com/NodeOS/NodeOS/issues/92 and found at https://github.com/NodeOS/NodeOS/issues/92#issuecomment-71224206, NodeOS can be build using Node.js v0.11.14 and earlier but not with Node.js v0.11.15, released last week.

When using the latest release, everything compiles fine but when running it Linux boots correctly but Node.js doesn't start nor give you any error. Since NodeOS is a Linux based system the error should be reproducible, so maybe could be related with the usage of musl as libc.

misterdjules:
I'm not familiar with NodeOS, do you think you'd have some bandwidth to find more details about why NodeOS can't run when it uses Node.js v0.11.15?

piranna:
NodeOS is basically Node.js compiled against musl as libc running on top of
a Linux 3.18.3 kernel, that's why I said probably is a problem related to
using musl instead of glibc. Why do you means by "bandwidth"?

misterdjules:
@piranna By "Do you have some bandwidth?" I meant "Do you have some time?", sorry for the confusion. Is it possible to have more details about why Node.js v0.11.15 wouldn't start, like an error message, or a trace output?

piranna:
Sorry, it gives no output :-( I could try to run the compiled binary on my
Ubuntu machine, though...

misterdjules:
@piranna Is the node binary fully statically linked? If not, then running it on Ubuntu would use Ubuntu's libc and libstdc++, and thus would not reproduce the issue, or am I missing something?

piranna:
It's compiled agains musl, so it will not use system glibc, I need to
explicit tell him where is the libc. I can be able to compile it statically
for the test, tough.

misterdjules:
@piranna As long as you're able to make it load the same C and C++ runtimes on Ubuntu than on NodeOS, the test should be valid. It doesn't need to be statically linked.

piranna:
I've just downloaded, compiled everything from scratch and tested v0.11.15 and it works on my Ubuntu system, but inside NodeOS it doesn't show the Node.js REPL prompt. Maybe it's a problem of using ```/dev/console``` as stdio?

On my Ubuntu machine, ```ldd``` outputs:

```bash
[...]

PR #9185, opened 2015-02-11
V8 lateral 3.28 update

trevnorris:
3.28.73 was actually from an unstable branch of V8. This reverts those changes to the actual latest stable 3.28. The reason for this is because of an edge case IC Miss that is causing a specific case to abort (see https://github.com/joyent/node/issues/9181#issuecomment-73807087)

I'm not sure if we want to re-enable handle zapping or not.

R=@tjfontaine @misterdjules


trevnorris:
@richardlau Thanks. Taken care of.

deps/v8/tools/gen-postmortem-metadata.py
```diff
@@ -150,7 +146,7 @@
     'Map, instance_size, int, kInstanceSizeOffset',
     'Map, bit_field, char, kBitFieldOffset',
     'Map, bit_field2, char, kBitField2Offset',
-    'Map, bit_field3, int, kBitField3Offset',
+    'Map, bit_field3, SMI, kBitField3Offset',
```
misterdjules: 4312f8d760bdf574b5a461e426331f0b4d298f18 changed `bit_field3` to be an `int` instead of an `SMI`, so that mdb_v8 can actually inspect its value and determine the correct type depending on which V8 version is used. We'll need to keep this field as an `int` to keep mdb_v8 working.

misterdjules:
It would be interesting to know if the following issues would be fixed by this upgrade:
* https://github.com/joyent/node/issues/9097
* https://github.com/joyent/node/issues/9084
* https://github.com/joyent/node/issues/9113

trevnorris:
#9113 unfortunately is not fixed by this. Still investigating the others.

trevnorris:
This also doesn't help #9084. Don't have the energy today, but may do a bisect of V8 versions until I can find the one that seems to have broken so many things.

trevnorris:
Bisected one of the issues down: https://github.com/joyent/node/issues/9084#issuecomment-75099054

misterdjules:
@trevnorris Just FYI, [I asked @piranna to test this PR with NodeOS](https://github.com/joyent/node/issues/9097#issuecomment-75332519). It will be interesting to know if it fixes this use case.

deps/v8/src/base/platform/platform.h
```diff
@@ -35,7 +35,6 @@ namespace std {
 int signbit(double x);
 }
 # endif
-#include <alloca.h>
```
misterdjules: This had been [added by floating a patch](3589a62104eac1daae782ab479bec09f4df4dc9a) that fixes build issues on SmartOS. In the meantime, I believe most of the changes from this patch except the inclusion of the `alloca.h` header on Solaris derivatives have been upstreamed.

Floating this patch again would solve this build issue on SmartOS: http://jenkins.nodejs.org/job/node-test-commit-unix/DESTCPU=ia32,label=smartos/13/console.

We should however probably upstream it too.

misterdjules:
@trevnorris It seems we're also missing [another floating patch](https://github.com/joyent/node/commit/6ebd85e10535dfaa9181842fe73834e51d4d3e6c).

misterdjules:
@trevnorris There's also [a build error](http://jenkins.nodejs.org/job/node-test-commit-unix/DESTCPU=ia32,label=osx/13/console) on one of the OS X Jenkins agent. I tried to reproduce it on my local OS X setup without success.

piranna:
Ok, I've checked NodeOS 32 bits with the commit ac21a53 and it's not working, so seems it's not fixed, sorry... :-(

misterdjules:
Pasting my comment in https://github.com/joyent/node/issues/9097#issuecomment-75886075 here to provide some context:

The commit you would want to test though is https://github.com/joyent/node/commit/d6249d0b4a47b9a073a06e6d2f183d9a66fc367c, and it requires the other commits here: https://github.com/joyent/node/pull/9185/commits.

https://github.com/joyent/node/commit/ac21a5390c6e02f0bc9052c7794dcabc33a0a0c3 in itself wouldn't solve the issue with NodeOS, as it only brings bug fixes and new features to `mdb_v8`, which is a library that helps mdb to inspect node processes.

Do you have some time to try these commits?

piranna:
According to ```git tree``` the commits are chained. I've tested with the different ones and didn't work, so I assume that no, the don't fix my problem with NodeOS :-(

misterdjules:
@piranna Yes, if you did a git checkout ac21a5390c6e02f0bc9052c7794dcabc33a0a0c3 after fetching joyent/node, then you tested the right set of changes.

I had interpreted your previous message as if you had specifically cherry-picked ac21a53, sorry for the confusion, and thank you again for your help!

piranna:
Yes, that's exactly what I did :-) git cherry-pick is a somewhat advanced command compared to git checkout, so it would be a little bit insane to test it this way, isn't it? ;-)

misterdjules:
@trevnorris Assigning to yourself since you've been originally workin on this, please let me know if you don't have time anymore to work on it.

misterdjules:
@trevnorris I've never used what is provided by handle zapping before. If I understand correctly, that helps when investigating crashes/bugs because handles removed from the handles scope have a very specific value that is easily recognizable.

I don't know what the impact on performance for removing that is (my understanding is that it could improve performance), and I don't know how useful it is for release builds.

@joyent/node-coreteam @indutny @bnoordhuis What do you think?

deps/v8/.gitignore
```diff
@@ -1,82 +0,0 @@
-*.a
```
misterdjules: Do you know why has the `.gitignore` file been removed by this downgrade?

deps/v8/src/base/platform/platform-posix.cc
```diff
@@ -321,15 +321,11 @@ int OS::GetCurrentProcessId() {
...
-#elif V8_OS_LINUX
+#if defined(ANDROID)
   return static_cast<int>(syscall(__NR_gettid));
-#elif V8_OS_ANDROID
-  return static_cast<int>(gettid());
 #else
-  return static_cast<int>(pthread_self());
-#endif
+  return static_cast<int>(syscall(SYS_gettid));
+#endif  // defined(ANDROID)
```
misterdjules: We probably want to float 3589a62104eac1daae782ab479bec09f4df4dc9a again, or cherry-pick another fix from upstream like https://github.com/v8/v8-git-mirror/commit/68d5f916d885462621d52ceafdd605ce1fd28c24.

deps/v8/src/debug-debugger.js
```diff
@@ -430,7 +430,7 @@ ScriptBreakPoint.prototype.set = function (script) {
   if (IS_NULL(position)) return;
 
   // Create a break point object and set the break point.
-  var break_point = MakeBreakPoint(position, this);
+  break_point = MakeBreakPoint(position, this);
```
misterdjules: We may want to float https://github.com/v8/v8-git-mirror/commit/2ad2237507c5b5f9047b8d94d2f4997327eae852 again to avoid issues with `--debug-brk` when used with `--use-strict`.

deps/v8/.gitignore
trevnorris: because `cp` doesn't copy `.*` from the dir. it can be reverted if it's an issue.

misterdjules:
@trevnorris Finally had some time to do a full review. Aside from the comments I added, I'm not familiar enough with V8's implementation to have a very valid opinion, but if that's what's the V8 team recommends, what are the next steps to land this?

trevnorris:
It's been so long I can't remember if it was the V8 team specifically that recommended this be done. It was more because the current version of V8 in Node isn't an actual stable release.

This will have to be rebased, and the changes you commented on will have to be made.

misterdjules:
@trevnorris Someone from the V8 team recommended this version here: https://github.com/joyent/node/issues/9241.

@trevnorris Do you have the time to do what needs to be done to land this (rebasing + floating some of the missing patches again)?

trevnorris:
@misterdjules sorry, i'm on site at Intuit for the week. I could work on this next Wednesday.

bnoordhuis:
>  I've never used what is provided by handle zapping before. If I understand correctly, that helps when investigating crashes/bugs because handles removed from the handles scope have a very specific value that is easily recognizable.
>
> I don't know what the impact on performance for removing that is (my understanding is that it could improve performance), and I don't know how useful it is for release builds.

You would normally turn off handle zapping in release builds and turn it on in debug builds or nightlies.

By the way, you should do this upgrade sooner rather than later.  3.28.73 has a couple of memory corruption bugs that may be remotely exploitable and I think it also has that security bug where `new Function(...)` is vulnerable to arbitrary JS code execution.

There are probably more security patches that you have to back-port from newer releases, what with 3.28 not being maintained.  I'm fairly sure there were fixes in 3.29, 3.30 and 3.31 that haven't been applied to 3.28.

misterdjules:
@bnoordhuis Thank you for the feedback! Is there a resource somewhere where some information about V8's security issues/bugs can be found? I looked at v8-users and v8-dev but didn't find anything. /cc @jakobkummerow.

bnoordhuis:
I don't believe there is anything like a v8-security mailing list.  I just track V8 development and keep an eye out for patches that look like security fixes.

jakobkummerow:
Right. If you use the tip of a supported branch, you don't have to worry about anything. If you insist on using an outdated, unsupported branch, you can watch what we merge back to still-supported branches. If it fixes code that's already present in your branch, you'll want the patch too.

Re handle zapping, as @bnoordhuis said, it flushes out bugs but has a performance penalty. It should be turned off in production. It can be helpful for developing changes either within V8 or in the embedder -- basically anything that uses HandleScopes.

misterdjules:
@bnoordhuis Thanks!

@jakobkummerow Thank you for your feedback too! Regarding V8's branches and tags: I use this application to determine what the current V8 stable is: https://omahaproxy.appspot.com. Today, it seems that the current stable V8 is `4.2.77.14`.

In my clone of V8's github mirror repository, I can see [the corresponding tag](https://github.com/v8/v8-git-mirror/commit/e65bd3efa19fb96186900ea88809a78347bf23aa), and it seems that it corresponds to this commit: https://github.com/v8/v8-git-mirror/commit/d7f25f557c20afe13a383a331208f4a8f05db3a5.

However this commit doesn't seem to be on the 4.2.77 branch, but on the master branch. What branch(es) should I watch to know about changes that land in stable versions?

deps/v8/.gitignore
misterdjules: I think the less changes we have from upstream the better, even though in this case it has no impact on functionality, so I would say that we'd like to put it back in.

misterdjules:
@trevnorris The commit message for https://github.com/joyent/node/commit/ac21a5390c6e02f0bc9052c7794dcabc33a0a0c3 seems to include too many things. Something like `mdb_v8: fix dictionary properties access` seems to represent more what this change does.

There's an issue to upstream part of that change here: https://github.com/joyent/node/issues/9147.

I also wonder why we have two different symbols `prop_type_mask` and `fieldindex_mask` for the same value (`PropertyDetails::FieldIndexField::kMask`). Same for `PropertyDetails::FieldIndexField::kShift`. @indutny @trevnorris @davepacheco Are they duplicates?

jakobkummerow:
@misterdjules Close :-)

4.2.77.14 is the version of V8 shipped in the latest stable Chrome (and a reasonable approximation of what you want to ship). The actual latest V8 version on the 4.2 branch is at https://chromium.googlesource.com/v8/v8.git/+/branch-heads/4.2. As of this writing the branch's tip is 4.2.77.15, one back-merge ahead of what Chrome ships -- the next Chrome update will pick it up. This is the authoritative source of the branch that you'll want to watch.

Note how the tag you found on github says "Merged d7f25f5" -- so yeah, d7f25f5 is indeed a master commit that was merged to the 4.2 branch. The merge commit is e65bd3efa19fb96186900ea88809a78347bf23aa (you can find this hash a little further down on the page).
Also note that the branch you/we care about is called "4.2", not "4.2.77". GitHub doesn't show it as a branch; that's an artifact of the mirroring process.

piranna:
> 4.2.77.14 is the version of V8 shipped in the latest stable Chrome (and a reasonable approximation of what you want to ship). The actual latest V8 version on the 4.2 branch is at https://chromium.googlesource.com/v8/v8.git/+/branch-heads/4.2. As of this writing the branch's tip is 4.2.77.15, one back-merge ahead of what Chrome ships -- the next Chrome update will pick it up. This is the authoritative source of the branch that you'll want to watch.

Then maybe we should follow the updates and use the latest stable release, isn't it? Maybe a good aproach would be to use the same major version on the current branch, so since v0.12 ships a 3.x v8 we should use the latest 3.x v8 stable release but the 4.x on Node.js v0.13, isn't it?

jakobkummerow:
@piranna Yes you should follow the updates and use the latest stable release, which right now is 4.2.77.15. You can see on https://chromium.googlesource.com/v8/v8.git/+/branch-heads/4.2 when the tip of the 4.2 branch updates, and you can see on omahaproxy.appspot.com when you should switch from 4.2 to 4.3. (I wonder how many more times I'll have to repeat that.)

V8 branch naming works somewhat similar to Linux kernel versions:
- 3.29 -> 3.30 -> 4.1 -> 4.2 -> 4.3 -> ... happens predictably every six weeks, and every one of those branches, a while after having been created, becomes the stable branch you should update to. 
- 3.x -> 4.x -> 5.x happens irregularly every couple of years when we feel like it, and is just naming. There's absolutely no reason to stick with "the latest stable 3.x release" -- this concept has no meaning. Today, 3.30 is just as outdated and unsupported as 4.1 is.

piranna:
I was not asking to do it myself, I don't have enough Node.js internals to do so, but instead about the path Node.js dev should follow. According to your message, they should release the latest and greatest release of v8, similar to what seems io.js is currently doing, isn't it? :+1: to this.

misterdjules:
@trevnorris @joyent/node-coreteam Created https://github.com/joyent/node/pull/18206 to continue this work.

piranna:
Do you suggest we test against that brach to see if it works for us (for
example on NodeOS?

misterdjules:
@piranna No, unfortunately there's no reason that the PR at #18206 would fix issues with NodeOS if this one does not, but thank you very much for asking! :+1:

misterdjules:
@trevnorris Closing this one in favor of #18206 to avoid confusion, but of course let me know if we should keep it open. Thanks!

misterdjules:
@jakobkummerow Thank you very much for the clarification on the branching model, I will use https://chromium.googlesource.com/v8/v8.git/ from now on, since it seems to have all the information I need contrary to the GitHub mirror. 

Also, to clarify why the current V8 version in Node.js is outdated: I don't think anyone "insist[s] on using an outdated, unsupported branch". We would use the latest and greatest if we could and are very much aware that the differences between V8's release cycle and Node.js' release cycle make this difficult. We are working on solving that problem. However, this is not the only problem we need to solve. It takes time and our time is limited.

I hope you'll be happy to hear that you won't have to repeat yourself :)

Thank you!

piranna:
You are welcome :-)

PR #9201, opened 2015-02-12
In all native modules, `require` should be called.

ghost:
The NativeModule system passes `NativeModule.require` transparently there. All other require calls in this module are without `NativeModule.` prefix, these three are last to finally finish what should be done back there when lib/module.js was being extracted.

The only one which _should_ have the prefix is the in line 295, where actually implements a big fs-based module system and actually requires a native module. That is left unchanged.

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -20,11 +20,11 @@
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 var NativeModule = require('native_module');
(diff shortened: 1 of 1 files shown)


herby:
Fixes #2009.

trevnorris:
Thanks. Landed in  b8604fa480 with minor commit message reword.

PR #9205, opened 2015-02-12
mdb_v8: fix cstyle nits

davepacheco:
The illumos platform (which shares the same code) has stricter cstyle guidelines.  These changes bring the canonical Node code in sync with illumos.


brendanashworth:
This is change to a dependency, and all dependency patches must be upstreamed first before we accept them into core (this is a policy, not mine!). Thanks for putting the time into this, but I'm going to close this as I don't think we can accept it.

davepacheco:
This has been upstream in mdb_v8 for a while.

davepacheco:
That said, mdb_v8 itself has evolved quite a bit since the last sync, and we should probably resync the whole thing.  @misterdjules has been doing a bunch of the recent work on this.

brendanashworth:
Ah! If you wanted to float a specific patch you could, but you'd need to provide more info on the patch shas and stuff. Perhaps you could bring this up in https://github.com/nodejs/node/issues/2517?

PR #9207, opened 2015-02-13
doc:fix default value of `decodeURIComponent`

h7lin:


misterdjules:
Thank you! LGTM, I would land that on v0.12 though, since it's also a problem in the v0.12 branch.

misterdjules:
I would also add more details to the documentation for `querystring.unescape`. I'm not sure if it should mention that it uses `decodeURIComponent` and falls back to `QueryString.unescapeBuffer`, as these seem to be implementation details.

misterdjules:
Regarding my previous comment, I think mentioning that `querystring.unescape` uses `decodeURIComponent` and falls back to a safer equivalent that doesn't throw on malformed URLs is fine.

robertkowalski:
i added the changes in https://github.com/joyent/node/pull/9259 - the both commits should not conflict and should get merged together

misterdjules:
@h7lin @robertkowalski Thank you both!

@h7lin Do you want to use your real name as the author name for your commit, or do you want to keep "h7lin"?

misterdjules:
@h7lin Thank you! Landed in f1bf88395926d84fb75cc055849624caad8f0144.

PR #9208, opened 2015-02-13
test: integrate v8 test suite

jasnell:
Make it possible to run the basic v8 test suite.

To get started, include `--with-v8-tests` on `configure`. Running
`make` will then build the test suite.

``` bash
./configure --with-v8-tests
make
make test-v8
```

If you include i18n when you build, you can run `make test-v8-intl`

``` bash
./configure --with-v8-tests --with-intl=full-icu
make
make test-v8-intl
```

Quick checks only. No presubmits. Release mode.

The commit includes some patches to the v8 dependency. It seems 
that the .gitignore filters out some necessary log files and 
intl/collator/default-locale appears to be working on macos 10.10.

/cc @misterdjules @tjfontaine - Please take a look and let me know if
this is a good approach..

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,6 +176,25 @@ test-timers:
 test-timers-clean:
 	$(MAKE) --directory=tools clean
 
+test-v8:
+	# note: only does a quickcheck
+	deps/v8/tools/run-tests.py --arch=$(ARCH) --mode=release --noi18n \
+	  --no-presubmit --quickcheck --shell-dir=out/Release
+
+test-v8-intl:
+	# note: only does a quickcheck...
+	deps/v8/tools/run-tests.py --arch=$(ARCH) --mode=release --no-presubmit \
+	  --quickcheck --shell-dir=out/Release intl
+
+test-v8-benchmarks:
+	# note: this runs with --download-data so it'll go out and
+	#       fetch the additional bits it needs to run the benchmarks
+	deps/v8/tools/run-tests.py --arch=$(ARCH) --mode=release --download-data \
+	  --no-presubmit --shell-dir=out/Release benchmarks
+
+test-v8-all: test-v8 test-v8-intl test-v8-benchmarks
+	# runs all v8 tests
+
 apidoc_sources = $(wildcard doc/api/*.markdown)
 apidocs = $(addprefix out/,$(apidoc_sources:.markdown=.html)) \
           $(addprefix out/,$(apidoc_sources:.markdown=.json))
--- a/configure
+++ b/configure
@@ -318,6 +318,11 @@ parser.add_option('--xcode',
     dest='use_xcode',
     help='generate build files for use with xcode')
 
+parser.add_option('--with-v8-tests',
+    action='store_true',
+    dest='with_v8_tests',
+    help='Build v8 test suite')
+
 (options, args) = parser.parse_args()
 
 # set up auto-download list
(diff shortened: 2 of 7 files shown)


jasnell:
Oh... one very important additional point: one of the tests (cctest) requires C++11 in order to compile. To build successfully on mac I had to set `std=c++11`. Obviously this is a departure from the current build...

```
export CXX="`which clang++` -std=c++11"
export LINK="`which clang++` -std=c++11"
export GYP_DEFINES="clang=1"
```

trevnorris:
Nice work.

/cc @tjfontaine @cjihrig

cjihrig:
+1

trevnorris:
@misterdjules Know if this would change anything with Jenkins?

jasnell:
cc @joyent/node-coreteam ... where are we at with this?

misterdjules:
@jasnell @trevnorris Please accept my apologies for the delay in my response.

First, thank you very much James for this!

I tried to run `./configure --with-v8-tests && make` and I got the following output:
```
LD_LIBRARY_PATH=/Users/JulienGilli/dev/node/v0.12/out/Release/lib.host:/Users/JulienGilli/dev/node/v0.12/out/Release/lib.target:$LD_LIBRARY_PATH; export LD_LIBRARY_PATH; cd ../.; mkdir -p deps/v8/testing; svn checkout --force https://googletest.googlecode.com/svn/trunk/ deps/v8/testing/gtest --revision 692
Error validating server certificate for 'https://googletest.googlecode.com:443':
 - The certificate is not issued by a trusted authority. Use the
   fingerprint to validate the certificate manually!
Certificate information:
 - Hostname: *.googlecode.com
 - Valid: from Fri, 27 Feb 2015 20:12:30 GMT until Thu, 28 May 2015 00:00:00 GMT
 - Issuer: Google Inc, US
 - Fingerprint: 9c:df:52:b3:2a:2c:11:16:a1:c2:ba:78:66:ba:f6:84:31:0e:72:5d
(R)eject, accept (t)emporarily or accept (p)ermanently? t
svn: E175002: Server sent unexpected return value (502 Bad Gateway) in response to PROPFIND request for '/svn/trunk'
make[1]: *** [deps/v8/testing/gtest] Error 1
make: *** [node] Error 2
```

I had tested it before (around the time the PR was submitted), and it had worked well, so I'm not sure if I'm doing something wrong or if something changed on google's side in the meantime. Did you try to build it recently?

Other than that, in my opinion it would be better to download gtest only if we want to run V8's tests. In other words, building the default target would not do anything different than what it does currently, and gtest would be donwloaded only when building `test-v8*` targets.

I would also put the additional GYP configuration in a separate file and keep only what is needed to build the node binary in `node.gyp`.

As for the requirement on C++11, do we know exactly what code needs what feature(s) of C++11?

Thank you very much again!

jasnell:
I can go back and tweak the build so that the test stuff is built only if
the test-v8 options are selected. Give me about a week on that.

There is one test class that fails without the C++11 stuff. It may be
possible to exclude it.

misterdjules:
Sounds good, thank you @jasnell!

piscisaureus:
> Other than that, in my opinion it would be better to download gtest only if we want to run V8's tests. In other words, building the default target would not do anything different than what it does currently, and gtest would be donwloaded only when building test-v8* targets.

I would recommend against downloading stuff as part of a build step. I suggest just putting the tests in git.

jasnell:
Replaced with #14185

PR #9215, opened 2015-02-13
console: allow Object.prototype fields as labels

misterdjules:
This is a backport of 6c3647c38d73f729ce85633a0440cd939d93dea2 from
v0.12 to v0.10.

Console.prototype.timeEnd() returns NaN if the timer label
corresponds to a property on Object.prototype. In v0.12, this was fixed
by using Object.create(null) to construct the _times object

However, the version of V8 in the v0.10 branch makes this fix not work
as expected. In v0.10, this commit changes the _times object into a
array of objects of the form:

{ label: someLabel, time: staringWallClockTime }

someLabel can thus be any string, including any string that represents
any Object.prototype field.

Fixes #9116.

### DIFF
--- a/lib/console.js
+++ b/lib/console.js
@@ -20,6 +20,7 @@
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 var util = require('util');
+var assert = require('assert');
 
 function Console(stdout, stderr) {
   if (!(this instanceof Console)) {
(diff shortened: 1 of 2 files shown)


cjihrig:
LGTM

misterdjules:
@cjihrig Thanks! Landed in c8239c0.

Issue #9118, opened 2015-01-30
if package.json's "main" does not resolve, no error reported and resolution continues


nicks:
Create a node_modules/a/package.json that looks like this:

``` json
{
  "name": "a",
  "main": "typo.js"
}
```

where `typo.js` does not resolve to a file. Then execute `require('a')`

Expected behavior:
- something should warn that "typo.js" does not resolve to anything
- module resolution should halt and report an error

Actual behavior:
- require() fails quietly, and continues to search all the module paths for a module

This is particularly bad due to the OS-dependent case-sensitivity problems with require(), because you can have a module that resolves to one version on one platform, and to another version on another platform.

jasnell:
Confirmed and agree that it's rather odd. @tjfontaine thoughts?

danny2394:
I am using v0.13.0-pre on Ubuntu 14.04 where there is an error popping up on terminal.
What version are you getting this bug on. @tjfontaine

cjihrig:
This seems like a valid bug to me, but the module system is the most frozen thing in Node, AFAIK.

PR #9225, opened 2015-02-16
Fix a bug in module loading when package.json's main does not resolve.

nicks:
Fixes https://github.com/joyent/node/issues/9118.

Before this change, the test would load the wrong module 'b'.

### DIFF
--- a/lib/module.js
+++ b/lib/module.js
@@ -91,13 +91,17 @@ function statPath(path) {
 // check if the directory is a package.json dir
 var packageMainCache = {};
 
+function getPackageJsonPath(requestPath) {
+  return path.resolve(requestPath, 'package.json');
+}
+
 function readPackage(requestPath) {
   if (hasOwnProperty(packageMainCache, requestPath)) {
     return packageMainCache[requestPath];
   }
 
   try {
-    var jsonPath = path.resolve(requestPath, 'package.json');
+    var jsonPath = getPackageJsonPath(requestPath);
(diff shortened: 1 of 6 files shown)


jasnell:
@cjihrig @trevnorris @misterdjules ... thoughts on this one?

trevnorris:
No comments from me.

/cc @othiym23 Any comment from the npm side of how this would affect things?

othiym23:
I don't think this will affect npm too much, because it already insists that `main` resolve as part of publishing. That said, this could have all sorts of weird breaking effects in practice, so even though it's arguably a by fix, it should probably be tagged semver-major.

jasnell:
Ok, given that, this would need to land in nodejs/node if it lands at all. Going to close this off here. @nicks .. if you'd like to pursue this further, may I ask that you open a new PR against nodejs/node master?

nicks:
sure, will do. one moment

nicks:
opened a PR as https://github.com/nodejs/node/pull/2568

lib/module.js
```diff
@@ -91,13 +91,17 @@ function statPath(path) {
...
+}
+
 function readPackage(requestPath) {
   if (hasOwnProperty(packageMainCache, requestPath)) {
     return packageMainCache[requestPath];
   }
 
   try {
-    var jsonPath = path.resolve(requestPath, 'package.json');
+    var jsonPath = getPackageJsonPath(requestPath);
```
mscdex: Why is this change necessary?

PR #9237, opened 2015-02-18
src: fix builtin modules failing with --use-strict

### LINKED ISSUE #9187: Node 0.12 stable fails on strict mode
I found this error when I run node with `node --use_strict myapp.js`. What can I do in this case?

syntaxerroronstrictmode

misterdjules:
Currently, lib/_tls_legacy.js and lib/crypto.js cannot be loaded when
--use-strict is passed to node. This change adds a test that makes sure
every external built-in module can be loaded with require when
--use-strict is enabled, and fixes the built-in modules that are
currently broken.

Fixes #9187.

### DIFF
--- a/lib/_tls_legacy.js
+++ b/lib/_tls_legacy.js
@@ -426,6 +426,24 @@ CryptoStream.prototype.end = function(chunk, encoding) {
   stream.Duplex.prototype.end.call(this, chunk, encoding);
 };
 
+/*
+ * Wait for both `finish` and `end` events to ensure that all data that
+ * was written on this side was read from the other side.
+ */
+function _destroyWhenReadAndWriteEndsDone(cryptoStream) {
+  var waiting = 1;
+
+  function finish() {
+    if (--waiting === 0) cryptoStream.destroy();
+  }
+
+  cryptoStream._opposite.once('end', finish);
+
+  if (!cryptoStream._finished) {
+    cryptoStream.once('finish', finish);
+    ++waiting;
+  }
+}
 
 CryptoStream.prototype.destroySoon = function(err) {
   if (this === this.pair.cleartext) {
@@ -440,18 +458,7 @@ CryptoStream.prototype.destroySoon = function(err) {
   if (this._writableState.finished && this._opposite._ended) {
     this.destroy();
   } else {
-    // Wait for both `finish` and `end` events to ensure that all data that
-    // was written on this side was read from the other side.
-    var self = this;
-    var waiting = 1;
-    function finish() {
-      if (--waiting === 0) self.destroy();
-    }
-    this._opposite.once('end', finish);
-    if (!this._finished) {
-      this.once('finish', finish);
-      ++waiting;
-    }
+    _destroyWhenReadAndWriteEndsDone(this);
   }
 };
 
--- a/test/simple/test-use-strict-builtin-modules.js
+++ b/test/simple/test-use-strict-builtin-modules.js
@@ -1,0 +1,34 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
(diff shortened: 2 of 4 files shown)


misterdjules:
As @cjihrig pointed out, it's been implemented in io.js with https://github.com/iojs/io.js/commit/21130c7d6fcaee666f33735768c060be2e06614a.

I think there is still some value in keeping the test added by this PR, as it makes sure that any new built-in module that could be added in the future works in strict mode.

Also, https://github.com/iojs/io.js/commit/21130c7d6fcaee666f33735768c060be2e06614a fixes the `"function can only be declared or top-level or witihin another function"` error by declaring a variable that holds a function object instead. But I think there's an opportunity to make the code clearer here, as the changes in this PR does.

What do you think?

cjihrig:
I'm all for adding `'use strict';` to all of the files in `/lib`. I'm also +1 for the test, although I think it would be fine to just `require()` all the files in `/lib` instead of filtering certain ones out.

misterdjules:
@cjihrig Added https://github.com/misterdjules/node/commit/c0bb576cbdd0108947edc694022133313a212f26 to this PR according to your comments.

misterdjules:
@tjfontaine @trevnorris @cjihrig @orangemocha @srl295 @jasnell @mdawsonibm I added this to the0.12.1 milestone, but I would like to get your thoughts on doing so.

lib/_tls_legacy.js
```diff
@@ -426,6 +428,24 @@ CryptoStream.prototype.end = function(chunk, encoding) {
   stream.Duplex.prototype.end.call(this, chunk, encoding);
 };
 
+/*
+ * Wait for both `finish` and `end` events to ensure that all data that
+ * was written on this side was read from the other side.
+ */
+CryptoStream.prototype._destroyWhenReadAndWriteEndsDone = function() {
```
cjihrig: I don't think we really need to expose another method on the prototype just to deal with the `finish()` function.

test/simple/test-use-strict-builtin-modules.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  return true;
+}
+
+if (process.argv[2] !== 'child') {
+  child_process.fork(__filename, ['child'], { execArgv: ['--use-strict']});
+} else {
+  var natives = process.binding('natives');
+  Object.keys(natives).filter(shouldBeTested).forEach(require);
```
cjihrig: Is filtering really necessary here? I ran this locally on 0.12 without the filter, and it worked fine.

cjihrig:
I think it should be fine to go into 0.12.1.

lib/_tls_legacy.js
misterdjules: Agreed, thanks!

test/simple/test-use-strict-builtin-modules.js
misterdjules: Excellent point, thank you!

It was needed before because the built-in modules were loaded by passing their relative paths to `require`, hence causing them to be loaded twice. Now that we iterate over `process.binding('natives')`, the modules are loaded using their name and not their path, and thus already loaded modules (like the module named `module`) are not loaded another time.

misterdjules:
@cjihrig I realized that enabling `'use strict';` in all built-ins modules would actually break a lot of existing code.

Failure to be strict-mode compliant is reported at runtime only when non-compliant code is executed, not when it's loaded. Thus, in order to be confident that enabling strict mode won't break a lot of existing applications, we'll need to make such a change sit in the code base for a while, let users report issues and fix them.

The current tests suite passes with all built-ins modules in strict mode, but it's not sufficient since it doesn't have 100% code coverage. Even if we reached such a number, there could still be a common code path/input/state combination that would break strict-mode and that is not tested by the tests suite.

For this reason I would suggest enabling strict-mode in all built-in modules in master and not in v0.12.

What do you think?

cjihrig:
It's your call. My personal preference would be to put it in 0.12.1, but I understand your concern.

misterdjules:
@cjihrig Thank you for your feedback!

To illustrate my point and to provide more context, I broke node-inspector (and probably other code) by enabling strict mode in `deps/v8/src/debug-debugger.js` with https://github.com/joyent/node/commit/45f1330425b671905a02fe30ee7a1dd9544c2709.

That change broke node-inspector even though the file was loading fine and the built-in debugger was working without any issue. The reason I hadn't seen the problem is that node-inspector simply has a different coverage than the built-in debugger or any of the tests in node's tests suites.

I will keep the PR as it is for the v0.12 branch, and submit another PR targeted to master that enables strict mode in all built-in modules.

misterdjules:
@cjihrig Could you please rubber-stamp it if you're ok with merging the current PR in v0.12? Thank you!

cjihrig:
LGTM

trevnorris:
LGTM

misterdjules:
@cjihrig @trevnorris Thanks for the review!

The tests suite failed for `test/simple/test-http-curl-chunk-problem.js`, but it should be fixed with https://github.com/misterdjules/node/commit/7eaeddb93cd344bfa2386a0eb620e92e3ec5c2c9. If that commits makes the tests suite pass, I'll land this PR without this fix and submit another PR for that fix.

misterdjules:
[All tests passing](http://jenkins.nodejs.org/job/node-accept-pull-request/24/) and removed the additional commit, will land soon.

misterdjules:
Landed in b233131, thank you!

trevnorris:
Can this be closed?

misterdjules:
Yes, closing now.

PR #9244, opened 2015-02-18
fs: properly handle fd passed to truncate()

charlanxcc:
Currently, fs.truncate() silently fails when a file descriptor
is passed as the first argument. This commit changes this
behavior to properly call fs.ftruncate().

PR-URL: https://github.com/joyent/node/pull/9161
Reviewed-By: Colin Ihrig cjihrig@gmail.com
Reviewed-By: Timothy J Fontaine tjfontaine@gmail.com

### DIFF
--- a/doc/api/fs.markdown
+++ b/doc/api/fs.markdown
@@ -106,7 +106,8 @@ Synchronous ftruncate(2).
 ## fs.truncate(path, len, callback)
 
 Asynchronous truncate(2). No arguments other than a possible exception are
-given to the completion callback.
+given to the completion callback. A file descriptor can also be passed as the
+first argument. In this case, `fs.ftruncate()` is called.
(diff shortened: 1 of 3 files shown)


charlanxcc:
Fix for build issue in 64 bit freebsd.

There are two build fixes. The first is in ```OS::GetSharedLibraryAddresses()```. In this method two automatic variables have the same name: ```result```. I changed one of them to ```addr``` to resolve the issue.

The second one is in ```OS::GetCurrentThreadId()```. In here, the code tried to cast ```pthread_t``` (64 bit) to ```int``` (32 bit), resulting in error. Proper fix should be return type change to ```long```. But as workaround for now, I modified it to ```((int) (int64) (void *) pthread_self())``` only for freebsd.

PR #9257, opened 2015-02-20
Revert "url: reslove urls with . and .."

misterdjules:
This reverts commit ad0684807c474db5cda7d28592e34e19910eb7ab.

Initially, this bug fix targeted master, and I pushed to have it
included in v0.10. In retrospect, I'm not sure it should have made into
v0.10 as it seems it could break a lot of existing working code.

In my opinion, this change is still a bug fix, and it is not backward
incompatible per se. However, I'm not sure that taking the risk to break
a lot of users with a new 0.10.x release that would include this fix is
reasonable, especially now that 0.10.x releases are entering
maintenance mode.


misterdjules:
@trevnorris I apologize for the confusion, I should have put more thoughts into that before pushing to land it in v0.10.

@trevnorris  @tjfontaine @orangemocha @cjihrig What do you think?

The follow-up question is: do we want to release 0.12.1 with this fix too? I still haven't made up my mind about this, although I would rather err on the side of caution.

trevnorris:
Alright on reverting the v0.10 patch, though I think landing it in v0.12.1 would be alright.

cjihrig:
+1 on what @trevnorris said.

misterdjules:
@trevnorris @cjihrig I agree that it would break less users, but technically since it hasn't been included in the v0.12.0 release, this would still be a behavior change that would break users.

I think that if we want to be consistent we would unfortunately need to revert it from v0.12 too and land it in master.

misterdjules:
Thank you, landed in a5be84f!

PR #9263, opened 2015-02-22
fs: fix link and symlink error messages

schnittstabil:
Currently, the error messages for missing arguments of symlink, link, symlinkSync and linkSync are swapped. Additionally they are named inconsistently all over the sources and the docs.

From the docs:

> `fs.link(srcpath, dstpath, callback)`
> `fs.linkSync(srcpath, dstpath)`
> `fs.symlink(srcpath, dstpath, callback)`
> `fs.symlinkSync(srcpath, dstpath)`

But the error messages doesn't match:

``` js
var fs = require('fs');
> fs.symlink('srcpath')
TypeError: src path must be a string
```

There are two things to mention here:
1. `dstpath` is missing, not the `srcpath` argument
2. it's not `src`, it's `srcpath`

In addition, #8920 shows that the meaning of `srcpath` and `dstpath` isn't clear and the docs aren't comprehensive enough.

This PR changes `srcpath` to `target` and `dstpath` to `path` in the docs and the sources, and make the code more consistent.

### DIFF
--- a/doc/api/fs.markdown
+++ b/doc/api/fs.markdown
@@ -200,25 +200,33 @@ Synchronous lstat(2). Returns an instance of `fs.Stats`.
 
 Synchronous fstat(2). Returns an instance of `fs.Stats`.
 
-## fs.link(srcpath, dstpath, callback)
+## fs.link(target, path, callback)
 
 Asynchronous link(2). No arguments other than a possible exception are given to
 the completion callback.
 
-## fs.linkSync(srcpath, dstpath)
+## fs.linkSync(target, path)
 
 Synchronous link(2).
 
-## fs.symlink(srcpath, dstpath[, type], callback)
+## fs.symlink(target, path[, type], callback)
 
-Asynchronous symlink(2). No arguments other than a possible exception are given
-to the completion callback.
+Asynchronous symlink(2). Creates a symbolic link at `path`, which links to `target`.
+No arguments other than a possible exception are given to the completion callback.
 The `type` argument can be set to `'dir'`, `'file'`, or `'junction'` (default
 is `'file'`) and is only available on Windows (ignored on other platforms).
-Note that Windows junction points require the destination path to be absolute.  When using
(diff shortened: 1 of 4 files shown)


piscisaureus:
In io.js this was further improved: https://github.com/iojs/io.js/commit/bc2c85ceef7ac034830e4a4357d0aef69cd6e386

jasnell:
@piscisaureus does the io.js patch include the commits in PR, complement them, or provide an alternate for them?

piscisaureus:
@jasnell my patch predates this one. It fixes more things, including an incorrect invocation of TrowUVException(). Look at the [commit message](https://github.com/iojs/io.js/commit/bc2c85ceef7ac034830e4a4357d0aef69cd6e386) to see all the details. And feel free to judge for yourself. 

We may want to land the test and doc changes from this PR.

schnittstabil:
I like the improvements of @piscisaureus, but haven't known them.

The only thing that I personally would prefer are better parameter names for link **and symlink**, consistently used within sources and docs, [so that error messages aren't such misleading](https://github.com/joyent/node/issues/8920#issuecomment-67978600) (see also [here](https://github.com/joyent/node/pull/8936#issuecomment-75467288)).

From [ln](http://unixhelp.ed.ac.uk/CGI/man-cgi?ln):
> ln [OPTION]... [-T] TARGET LINK_NAME
> ln [OPTION]... TARGET... DIRECTORY

From [mklink](https://technet.microsoft.com/de-de/library/cc753194%28v=ws.10%29.aspx):
> mklink [[/d] | [/h] | [/j]] &lt;Link&gt; &lt;Target&gt; // swapped order regarding ln

Therefore I'd vote for `target` and `[link]path` - instead of `src`/`dest` and `destination`/`path`.

jasnell:
@schnittstabil ... if you'd like to pursue this change further, opening a new updated one against http://github.com/nodejs/node master would be the right approach. This is not going to be able to land here in it's current state. I'm going to go ahead and close this but can reopen if necessary

PR #9266, opened 2015-02-23
Extend non-standard Python documentation

imogensharp:
Add additional documentation regarding building with a Python interpreter installed to a non-standard location.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,12 @@ make
 make install
 ```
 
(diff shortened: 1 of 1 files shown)


jasnell:
@misterdjules @orangemocha ... thoughts?

@imogensharp ...to pursue this, you may want to close here and open an issue on http://github.com/nodejs/build or /nodejs/node

jasnell:
Given that this is not going to land here, a new PR would need to be opened at nodeje/node master or against v0.12. Closing this here for now.

PR #9268, opened 2015-02-23
net: allow port 0 in connect()

### LINKED ISSUE #9194: Random port: 0 doesn’t work on Node 0.12.0
https://github.com/gruntjs/grunt-contrib-connect/issues/126

See the above issue link in the gruntjs project for details. This was apparently an issue in node v0.11.0, but no one filed it in the node project because v0.11.x was the experimental branch. This issue is now appearing in node v0.12.0.

To be clear, this issue doesn't appear to be an issue with gruntjs because I am seeing it in a completely different context in a project using the node-spdy library, which doesn't use gruntjs at all.

cjihrig:
This allows port 0 to be passed to `Socket.prototype.connect()` and `net.createConnection()`. However, by allowing port 0, we are eliminating all type checking since `var port = options.port | 0;` converts everything to a number. More thorough checking could be added by verifying that `options.port` is `undefined`, a number, or a numeric string (which must be supported for compatibility with `url.parse()` results), but at what point does it become overkill.

Closes #9194. cc: @trevnorris

### DIFF
--- a/lib/net.js
+++ b/lib/net.js
@@ -892,7 +892,7 @@ Socket.prototype.connect = function(options, cb) {
   } else {
     var dns = require('dns');
     var host = options.host || 'localhost';
-    var port = options.port | 0;
+    var port = 0;
     var localAddress = options.localAddress;
     var localPort = options.localPort;
     var dnsopts = {
@@ -906,8 +906,16 @@ Socket.prototype.connect = function(options, cb) {
     if (localPort && !util.isNumber(localPort))
       throw new TypeError('localPort should be a number: ' + localPort);
 
-    if (port <= 0 || port > 65535)
-      throw new RangeError('port should be > 0 and < 65536: ' + port);
+    if (typeof options.port === 'number')
+      port = options.port;
+    else if (typeof options.port === 'string')
+      port = options.port.trim() === '' ? -1 : +options.port;
+    else if (options.port !== undefined)
+      throw new TypeError('port should be a number or string: ' + options.port);
+
+    if (port < 0 || port > 65535 || isNaN(port))
+      throw new RangeError('port should be >= 0 and < 65536: ' +
+                           options.port);
 
     if (dnsopts.family !== 4 && dnsopts.family !== 6)
       dnsopts.hints = dns.ADDRCONFIG | dns.V4MAPPED;
--- a/src/tcp_wrap.cc
+++ b/src/tcp_wrap.cc
@@ -428,7 +428,7 @@ void TCPWrap::Connect(const FunctionCallbackInfo<Value>& args) {
 
   assert(args[0]->IsObject());
   assert(args[1]->IsString());
-  assert(args[2]->Uint32Value());
+  assert(args[2]->IsUint32());
 
   Local<Object> req_wrap_obj = args[0].As<Object>();
   node::Utf8Value ip_address(args[1]);
@@ -460,7 +460,7 @@ void TCPWrap::Connect6(const FunctionCallbackInfo<Value>& args) {
 
   assert(args[0]->IsObject());
   assert(args[1]->IsString());
-  assert(args[2]->Uint32Value());
+  assert(args[2]->IsUint32());
 
   Local<Object> req_wrap_obj = args[0].As<Object>();
   node::Utf8Value ip_address(args[1]);
--- a/test/simple/test-net-create-connection.js
+++ b/test/simple/test-net-create-connection.js
@@ -24,33 +24,99 @@ var assert = require('assert');
 var net = require('net');
 
 var tcpPort = common.PORT;
+var expectedConnections = 7;
 var clientConnected = 0;
 var serverConnected = 0;
 
 var server = net.createServer(function(socket) {
   socket.end();
-  if (++serverConnected === 4) {
+  if (++serverConnected === expectedConnections) {
     server.close();
   }
 });
+
 server.listen(tcpPort, 'localhost', function() {
   function cb() {
     ++clientConnected;
+  }
+
+  function fail(opts, errtype, msg) {
+    assert.throws(function() {
+      var client = net.createConnection(opts, cb);
+    }, function (err) {
+      return err instanceof errtype && msg === err.message;
+    });
   }
 
   net.createConnection(tcpPort).on('connect', cb);
   net.createConnection(tcpPort, 'localhost').on('connect', cb);
   net.createConnection(tcpPort, cb);
   net.createConnection(tcpPort, 'localhost', cb);
+  net.createConnection(tcpPort + '', 'localhost', cb);
+  net.createConnection({port: tcpPort + ''}).on('connect', cb);
+  net.createConnection({port: '0x' + tcpPort.toString(16)}, cb);
 
-  assert.throws(function () {
-    net.createConnection({
-      port: 'invalid!'
-    }, cb);
-  });
+  fail({
+    port: true
+  }, TypeError, 'port should be a number or string: true');
+
+  fail({
+    port: false
+  }, TypeError, 'port should be a number or string: false');
+
+  fail({
+    port: []
+  }, TypeError, 'port should be a number or string: ');
+
+  fail({
+    port: {}
+  }, TypeError, 'port should be a number or string: [object Object]');
+
+  fail({
+    port: null
+  }, TypeError, 'port should be a number or string: null');
+
+  fail({
+    port: ''
+  }, RangeError, 'port should be >= 0 and < 65536: ');
+
+  fail({
+    port: ' '
+  }, RangeError, 'port should be >= 0 and < 65536:  ');
+
+  fail({
+    port: '0x'
+  }, RangeError, 'port should be >= 0 and < 65536: 0x');
+
+  fail({
+    port: '-0x1'
+  }, RangeError, 'port should be >= 0 and < 65536: -0x1');
+
+  fail({
+    port: NaN
+  }, RangeError, 'port should be >= 0 and < 65536: NaN');
+
+  fail({
+    port: Infinity
+  }, RangeError, 'port should be >= 0 and < 65536: Infinity');
+
+  fail({
+    port: -1
+  }, RangeError, 'port should be >= 0 and < 65536: -1');
+
+  fail({
+    port: 65536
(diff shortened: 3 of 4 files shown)


misterdjules:
I think that adding additional checks to throw except for valid numerical strings and numbers is worthwhile, especially if the original code was doing that.

misterdjules:
Adding to the 0.12.1 milestone as it seems like a significant regression to me.

cjihrig:
@misterdjules added type checking back. PTAL

lib/net.js
```diff
@@ -906,8 +906,11 @@ Socket.prototype.connect = function(options, cb) {
     if (localPort && !util.isNumber(localPort))
       throw new TypeError('localPort should be a number: ' + localPort);
 
-    if (port <= 0 || port > 65535)
-      throw new RangeError('port should be > 0 and < 65536: ' + port);
+    if (!util.isUndefined(options.port) && options.port != port)
```
trevnorris: Do `options.port >>> 0 === port && options.port >= 0`. That's a uint check that excludes `NaN` and prevents wrap around.

trevnorris: This also means you can remove the additional below check. But you should do `port >>>= options.port` to make sure number strings (i.e. `'0xfe'`) are properly coerced.

trevnorris:
Couple comments, otherwise LGTM.

test/simple/test-net-localerror.js
```diff
@@ -23,27 +23,17 @@ var common = require('../common');
...
-  }, 'localAddress should be a valid IP: foobar');
-
-  connect({
-    host: 'localhost',
-    port: 65536
-  }, 'port should be > 0 and < 65536: 65536');
-
-  connect({
-    host: 'localhost',
-    port: 0
```
misterdjules: Instead of removing this test, can we put it back and assert it doesn't throw?

lib/net.js
misterdjules: @trevnorris I'm not sure I understand what you mean in the previous two comments. Could you please elaborate?

test/simple/test-net-create-connection.js
```diff
@@ -38,19 +38,34 @@ server.listen(tcpPort, 'localhost', function() {
...
   net.createConnection(tcpPort, cb);
   net.createConnection(tcpPort, 'localhost', cb);
 
-  assert.throws(function () {
-    net.createConnection({
-      port: 'invalid!'
-    }, cb);
-  });
+  fail({
+    port: 'invalid!'
```
misterdjules: I think it can't hurt to add more tests with that cover the rest of invalid values such as `null`, `NaN`, etc. What do you think?

misterdjules:
@cjihrig Thanks Colin! I added my comments inline.

lib/net.js
cjihrig: @trevnorris I don't think that will work. I added more tests at @misterdjules request, and that validation failed on the empty string IIRC. What does seem to handle all cases is `if (options.port !== undefined && parseInt(options.port, 10) !== port)`. I'm not big on using `parseInt()`, but it handles numeric strings well, which is the most annoying part of this check.

cjihrig:
@misterdjules I made the test changes your requested.

lib/net.js
```diff
@@ -906,8 +906,11 @@ Socket.prototype.connect = function(options, cb) {
     if (localPort && !util.isNumber(localPort))
       throw new TypeError('localPort should be a number: ' + localPort);
 
-    if (port <= 0 || port > 65535)
-      throw new RangeError('port should be > 0 and < 65536: ' + port);
+    if (options.port !== undefined && parseInt(options.port, 10) !== port)
```
misterdjules: Using `parseInt` we would accept `"42foo"` as a valid port, which we wouldn't accept before this change. I don't think we should accept these values, what do you think?

cjihrig: If `options.port` was `"42foo"`, then `parseInt()` would return 42, but `port` would be 0, and an exception would be thrown.

src/tcp_wrap.cc
```diff
@@ -428,7 +428,7 @@ void TCPWrap::Connect(const FunctionCallbackInfo<Value>& args) {
 
   assert(args[0]->IsObject());
   assert(args[1]->IsString());
-  assert(args[2]->Uint32Value());
+  assert(args[2]->Uint32Value() >= 0);
```
misterdjules: I may be missing something: how can `Uint32Value()` return a value that is not `>= 0`?

misterdjules: Instead, we might want to use `IsUint32()`.

cjihrig: Ah, you're right. I added the `>=` because when `Uint32Value()` was equal to 0, the assertion failed.

lib/net.js
trevnorris: @cjihrig ah, my checks are to see if they _are_. so instead it would be:
```javascript
if (!(options.port >>> 0 === options.port && options.port >= 0))
```

cjihrig: @trevnorris I'm still having some trouble with your checks, and variations on them:

`if ((options.port >>> 0 === options.port && options.port >= 0))` throws on valid port 12346
`if (!(options.port >>> 0 === options.port && options.port >= 0))` throws on valid port '12346'
`if (!(options.port >>> 0 === port && options.port >= 0))` does not throw on invalid port ''
`if ((options.port >>> 0 === port && options.port >= 0))` throws on valid port 12346

cjihrig:
@misterdjules updated to use `IsUint32()`

lib/net.js
misterdjules: Right, I thought `"42foo" | 0` would evaluate to `42`, which it does not. Thanks for the clarification!

Using `parseInt` only with a radix of `10` makes strings representing hexadecimal numbers such as `0x12` throw. The current version of the code does, and it seems that it matches with the concept of "undefined, a number, or a string representing a number".

I'm not sure if we want to support it though. I added support for strings representing hexadecimal numbers [in a gist](https://gist.github.com/misterdjules/7e4f25997b3805792aac), and the problem is that it would accept invalid values such as `0x123R`, since `0x123R | 0 === 0` and `parseInt(0x123R, 10) === 0` too.

Other than that, I think the above-mentioned gist improves the current changes by making the checks a bit clearer though, so feel free to reuse it without the hexadecimal string support.

lib/net.js
```diff
@@ -906,8 +906,11 @@ Socket.prototype.connect = function(options, cb) {
...
+      throw new TypeError('port should be a number: ' + options.port);
```
misterdjules: We might also want to change the error message to `'port should be undefined or should represent a valid number: ' + options.port'`, but I'm not sure how much of a breaking change that would be for users who consume error messages.

lib/net.js
misterdjules: I created another gist that I think makes the check even clearer: https://gist.github.com/misterdjules/741cb578638b18f139ef.

cjihrig: Unfortunately, I thought that `parseInt('0xff', 10)` would return 255, but that is not the case. I think this could be the answer:

`if (options.port !== undefined && parseFloat(options.port) !== port)`

cjihrig: @misterdjules OK, I found this, which appears to handle everything I've tried so far. It's less efficient than I'd like, but could also be added to `util`. Thoughts?

```javascript
function isNumeric(n) {
  return !isNaN(parseFloat(n)) && isFinite(n);
}
```

misterdjules: As mentioned in #libuv, `isNumeric` would not handle negative hexadecimal number strings, and thus `isNumeric` would be inconsistent. But we could have:
```
/*
 * This function returns `true` if `value` is a string that represents a valid
 * decimal or hexadecimal non-negative number, false otherwise.
 */
function isNonNegativeNumeric(n) {
  return !isNaN(parseFloat(n)) && isFinite(n) && n >= 0;
}
``` 
and use it like in the following gist: https://gist.github.com/misterdjules/7e4f25997b3805792aac.

cjihrig:
@misterdjules updated.

misterdjules:
@cjihrig Pasting what I said in #libuv to make sure it's not lost:

Another concern I had is that actually `parseFloat` doesn’t handle hexadecimal strings at all, for instance `parseFloat(‘0xff’) === 0`. In the current state of the PR, it works out correctly, but because we’re lucky.

It seems to me that your original method (comparing coercion to int’s result with parseInt’s output) would be more robust and consistent. We could add some additional checks and comments like something closer to what I had in this gist: https://gist.github.com/misterdjules/741cb578638b18f139ef with the error message mentioning that numbers and numeric strings are accepted instead of `'port shoud be a number'`.

What do you think?

trevnorris:
The best way for number coercion is `+number`. Use that instead of `parseFloat()`.

cjihrig:
Updated to a slightly new approach. Added more tests, and incorporated @trevnorris `+number` suggestion.

jasnell:
LGTM

misterdjules:
@cjihrig I like it, thank you very much for persisting!

LGTM and tests running here: http://jenkins.nodejs.org/job/node-test-commit-unix/ and here: http://jenkins.nodejs.org/job/node-test-commit-windows/. I don't expect any of them to fail since all tests pass locally, but better safe than sorry.

misterdjules:
More precisely, tests for UNICEs here: http://jenkins.nodejs.org/job/node-test-commit-unix/23/ and Windows here: http://jenkins.nodejs.org/job/node-test-commit-windows/22/.

misterdjules:
One test failing on SmartOS: http://jenkins.nodejs.org/job/node-test-commit-unix/23/DESTCPU=ia32,label=smartos/tapTestReport/simple.tap-437/, [`TCPWrap::Connect6`](https://github.com/joyent/node/blob/master/src/tcp_wrap.cc#L463) also needs to be updated.

cjihrig:
@misterdjules updated `Connect6`

misterdjules:
@cjihrig Tests running on [UNIX](http://jenkins.nodejs.org/job/node-test-commit-unix/24/) and [Windows](http://jenkins.nodejs.org/job/node-test-commit-windows/23/).

misterdjules:
@cjihrig All tests passed.

cjihrig:
Thanks! Landed in 1a2a4dac232673987fcea4471b19bb82502f0205

PR #9289, opened 2015-02-26
path: refactor for performance and consistency

nwoltman:
Improve performance by:
- Not leaking the `arguments` object! (in [`win32.join`](https://github.com/joyent/node/pull/9289/files#diff-c0fae59301ef49fb71b43278db9ad881L197))
- Getting the last character of a string by index, instead of with `.substr()` ([here](https://github.com/joyent/node/pull/9289/files#diff-c0fae59301ef49fb71b43278db9ad881L445)) or `.slice()` ([here](https://github.com/joyent/node/pull/9289/files#diff-c0fae59301ef49fb71b43278db9ad881L363))

Improve code consistency by:
- Using `[]` instead of `.charAt()` where possible (such as [here](https://github.com/joyent/node/pull/9289/files#diff-c0fae59301ef49fb71b43278db9ad881L428))
- Using a function declaration instead of a var declaration ([here](https://github.com/joyent/node/pull/9289/files#diff-c0fae59301ef49fb71b43278db9ad881R114))
- Using `.slice()` with clearer arguments (such as [here](https://github.com/joyent/node/pull/9289/files#diff-c0fae59301ef49fb71b43278db9ad881L387))

Improve both by:
- Making the reusable `trimArray()` function
- Standardizing getting certain path statistics with the new `win32StatPath()` function
#### Benchmarks: (higher is better)

| function (path.) | current (ops/sec) | this PR (ops/sec) |
| --- | --- | --- |
| win32.resolve | 315,019 | 399,543 |
| win32.normalize | 873,559 | 1,051,170 |
| win32.isAbsolute | 1,915,332 | 1,939,849 |
| win32.join | 452,879 | 815,033 |
| win32.relative | 153,821 | 219,928 |
| win32.format | 13,118,588 | 16,364,384 |
| posix.relative | 244,638 | 304,456 |

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -51,6 +51,29 @@ function normalizeArray(parts, allowAboveRoot) {
   return res;
 }
 
+// returns an array with empty elements removed from either end of the input
+// array or the original array if no elements need to be removed
+function trimArray(arr) {
+  var lastIndex = arr.length - 1;
+  var start = 0;
+  for (; start <= lastIndex; start++) {
+    if (arr[start])
+      break;
+  }
+
+  var end = lastIndex;
+  for (; end >= 0; end--) {
+    if (arr[end])
+      break;
+  }
+
+  if (start === 0 && end === lastIndex)
+    return arr;
+  if (start > end)
+    return [];
+  return arr.slice(start, end + 1);
+}
+
 // Regex to split a windows path into three parts: [*, device, slash,
 // tail] windows-only
 var splitDeviceRe =
@@ -76,9 +99,21 @@ function win32SplitPath(filename) {
   return [device, dir, basename, ext];
 }
 
-var normalizeUNCRoot = function(device) {
+function win32StatPath(path) {
+  var result = splitDeviceRe.exec(path),
+      device = result[1] || '',
+      isUnc = !!device && device[1] !== ':';
+  return {
+    device: device,
+    isUnc: isUnc,
+    isAbsolute: isUnc || !!result[2], // UNC paths are always absolute
+    tail: result[3]
+  };
+}
+
+function normalizeUNCRoot(device) {
   return '\\\\' + device.replace(/^[\\\/]+/, '').replace(/[\\\/]+/g, '\\');
-};
+}
 
 // path.resolve([from ...], to)
 win32.resolve = function() {
@@ -113,11 +148,11 @@ win32.resolve = function() {
       continue;
     }
 
-    var result = splitDeviceRe.exec(path),
-        device = result[1] || '',
-        isUnc = device && device.charAt(1) !== ':',
-        isAbsolute = win32.isAbsolute(path),
-        tail = result[3];
+    var result = win32StatPath(path),
+        device = result.device,
+        isUnc = result.isUnc,
+        isAbsolute = result.isAbsolute,
+        tail = result.tail;
 
     if (device &&
         resolvedDevice &&
@@ -159,11 +194,11 @@ win32.resolve = function() {
 
 
 win32.normalize = function(path) {
-  var result = splitDeviceRe.exec(path),
-      device = result[1] || '',
-      isUnc = device && device.charAt(1) !== ':',
-      isAbsolute = win32.isAbsolute(path),
-      tail = result[3],
+  var result = win32StatPath(path),
+      device = result.device,
+      isUnc = result.isUnc,
+      isAbsolute = result.isAbsolute,
+      tail = result.tail,
       trailingSlash = /[\\\/]$/.test(tail);
 
   // Normalize the tail path
@@ -187,22 +222,21 @@ win32.normalize = function(path) {
 
 
 win32.isAbsolute = function(path) {
-  var result = splitDeviceRe.exec(path),
-      device = result[1] || '',
-      isUnc = !!device && device.charAt(1) !== ':';
-  // UNC paths are always absolute
-  return !!result[2] || isUnc;
+  return win32StatPath(path).isAbsolute;
 };
 
 win32.join = function() {
-  function f(p) {
-    if (!util.isString(p)) {
+  var paths = [];
+  for (var i = 0; i < arguments.length; i++) {
+    var arg = arguments[i];
+    if (!util.isString(arg)) {
       throw new TypeError('Arguments to path.join must be strings');
     }
-    return p;
+    if (arg) {
+      paths.push(arg);
+    }
   }
 
-  var paths = Array.prototype.filter.call(arguments, f);
   var joined = paths.join('\\');
 
   // Make sure that the joined path doesn't start with two slashes, because
@@ -239,25 +273,10 @@ win32.relative = function(from, to) {
   var lowerFrom = from.toLowerCase();
   var lowerTo = to.toLowerCase();
 
-  function trim(arr) {
-    var start = 0;
-    for (; start < arr.length; start++) {
-      if (arr[start] !== '') break;
-    }
+  var toParts = trimArray(to.split('\\'));
 
-    var end = arr.length - 1;
-    for (; end >= 0; end--) {
-      if (arr[end] !== '') break;
-    }
-
-    if (start > end) return [];
-    return arr.slice(start, end + 1);
-  }
-
-  var toParts = trim(to.split('\\'));
-
-  var lowerFromParts = trim(lowerFrom.split('\\'));
-  var lowerToParts = trim(lowerTo.split('\\'));
+  var lowerFromParts = trimArray(lowerFrom.split('\\'));
+  var lowerToParts = trimArray(lowerTo.split('\\'));
 
   var length = Math.min(lowerFromParts.length, lowerToParts.length);
   var samePartsLength = length;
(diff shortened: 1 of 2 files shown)


jasnell:
@woollybogger ... do you have benchmark numbers showing that this provides a net performance improvement? The nature of the changes would indicate an improvement but numbers always help :-).

woollybogger:
@jasnell I updated the PR description with benchmark data. I only included the functions that had a >1% change in performance.

woollybogger:
It's been over a week so...bump?

jasnell:
@woollybogger ... sorry, was catching up on a few other items. Thanks for the profiling numbers. I'll queue this up to review on Monday. 

// cc @joyent/node-coreteam ... can one of you take a look as well?

trevnorris:
Generally LGTM. But let's run this on the servers and make sure all tests pass on all platforms.

lib/path.js
```diff
@@ -51,6 +51,27 @@ function normalizeArray(parts, allowAboveRoot) {
   return res;
 }
 
+// returns an array with empty elements removed from either end of the input
+// array or the original array if no elements need to be removed
+function trimArray(arr) {
+  var lastIndex = arr.length - 1;
+  var start = 0;
+  for (; start <= lastIndex; start++) {
+    if (arr[start]) break;
```
cjihrig: Please move `break` to the next line.

lib/path.js
```diff
@@ -51,6 +51,27 @@ function normalizeArray(parts, allowAboveRoot) {
...
+  }
+
+  var end = lastIndex;
+  for (; end >= 0; end--) {
+    if (arr[end]) break;
```
cjihrig: Same here

lib/path.js
woollybogger: Good catch! I forgot that's the coding style.

lib/path.js
```diff
@@ -115,8 +138,8 @@ win32.resolve = function() {
 
     var result = splitDeviceRe.exec(path),
         device = result[1] || '',
-        isUnc = device && device.charAt(1) !== ':',
-        isAbsolute = win32.isAbsolute(path),
+        isUnc = device && device[1] !== ':',
+        isAbsolute = isUnc || result[2],
```
misterdjules: I'm not a big fan on inlining the call to `win32.isAbsolute` here and in `win32.normalize`. I'm concerned that we could get to a point that `win32.isAbsolute` and these inline versions diverge.

misterdjules:
@woollybogger @trevnorris @jasnell Added a comment inline. Thoughts?

piscisaureus:
I know the damage has already been done, but one of the objections I have against exposing the unix code on windows and vice versa is that `path.resolve()` and maybe other functions produce garbage when relative paths are used.

Does anyone have an idea for fixing this. Maybe throw if the path is relative?

lib/path.js
woollybogger: We could have another function similar to [`win32SplitPath`](https://github.com/joyent/node/blob/v0.12/lib/path.js#L66) that returns the following properties:
+ `device`
+ `isUnc`
+ `isAbsolute`
+ `tail`

Also, is there any reason why [`win32SplitPath`](https://github.com/joyent/node/blob/v0.12/lib/path.js#L66) returns an array? I feel like the code would be more readable if it returned an object instead.

woollybogger: Ah, I'm guessing the answer to my question ^ is so that `win32SplitPath` returns the same data structure as [`posixSplitPath`](https://github.com/joyent/node/blob/v0.12/lib/path.js#L406).

misterdjules:
@piscisaureus I'm not sure I understand what you're saying, could you please give us an example of what you're describing?

piscisaureus:
@misterdjules 

```
> path.posix.resolve('foo/bar')
'C:\\Users\\Bert Belder/foo/bar'
```

I don't know what would happen if you did the opposite on unix, e.g.
```
> path.win32.resolve('\\\\myserver\\myshare\\foo.txt');
???
> path.win32.resolve('c:appdata\\node.js');
???
```

But I am sure the results would not make sense.

misterdjules:
@piscisaureus OK thank you for the clarification!

```
> path.win32.resolve('\\\\myserver\\myshare\\foo.txt');
???
> path.win32.resolve('c:appdata\\node.js');
???
```
work as expected because they are absolute paths, but like you said `path.win32.resolve` with relative paths would fail on any UNIX system.

It seems that only `path.resolve` would be affected by this, as it seems to be the only one calling a platform-specific function (`process.cwd`). `path.relative` uses `path.resolve`, but it doesn't seem to be affected by it since it seems to deal only with relative paths. We would need to investigate more though.

Anyway, I think this should be handled outside of this PR and in my opinion it's worth creating a new issue. Do you want to do that @piscisaureus or should I do it?

Thank you again!

piscisaureus:
@misterdjules 

Ah, oops, you're right.
(The second path `c:appdata\\node.js` isn't absolute btw). It does a drive-specific cwd lookup [here](https://github.com/joyent/node/blob/9b534e2e87489d219f0d0cf5be7f9119704fe424/lib/path.js#L102).

Agreed btw that I should be handled in a separate issue.

misterdjules:
@piscisaureus created new issue here: https://github.com/joyent/node/issues/9430.

lib/path.js
woollybogger: @misterdjules Are things good enough inline or should there be a helper function such as the one I outlined above? I tested Node with the changes in the PR against a version with the helper function and it does slow things down a little (about 2% to 5% slower for the functions the helper touches).

misterdjules: Unless `win32.isAbsolute` is on the critical path of a very common use case in terms of performance, I have a preference for using a helper function such as the one you described [in your comment above](https://github.com/joyent/node/pull/9289#discussion_r26546997). 

I want to avoid subtle regressions due to potential divergence between several versions of inlined `win32.isAbsolute`.

woollybogger:
@misterdjules @jasnell Updated with the helper function as discussed [here](https://github.com/joyent/node/pull/9289#discussion-diff-26521014) and updated the benchmark data.

lib/path.js
```diff
@@ -360,11 +379,10 @@ win32.format = function(pathObject) {
 
   var dir = pathObject.dir;
   var base = pathObject.base || '';
-  if (dir.slice(dir.length - 1, dir.length) === win32.sep) {
-    return dir + base;
-  }
-
   if (dir) {
+    if (dir[dir.length - 1] === win32.sep) {
```
misterdjules: Does this change (moving the check inside `if (dir)`) fixes a bug where sometime `dir` would be undefined and thus the code would throw? If so, then it would be great to add at least a test for that.

misterdjules:
Could you please take a look at my [comment](https://github.com/joyent/node/pull/9289#discussion_r26788172)? Otherwise, LGTM. Thanks @woollybogger! :+1:

Adding to the 0.12.2 milestone as node v0.12.1 will be a security release following the recent OpenSSL releases.

lib/path.js
woollybogger: Yes, I can think of at least two test cases that this fixes. I'll add some tests tonight.

woollybogger:
@roryrjb You wrote the `path.format` and `path.parse` functions so I have a question for you. In `path.format`, [`root`](https://github.com/joyent/node/pull/9289/files#diff-c0fae59301ef49fb71b43278db9ad881R371) isn't used anywhere except to [`throw an error`](https://github.com/joyent/node/pull/9289/files#diff-c0fae59301ef49fb71b43278db9ad881R373). Is there a good reason for this or can that check just be removed?

roryrjb:
@woollybogger no, there's no reason I can think of, it can be removed.

woollybogger:
@roryrjb Cool, thanks! :+1:

woollybogger:
@misterdjules Ran into a little snag when writing tests:
```js
path.posix.format({root:'', dir: 'dir/other dir/', base: 'index.html'});
// -> 'dir/other dir//index.html'
path.win32.format({root:'', dir: 'dir\\other dir\\', base: 'index.html'});
// -> 'dir\\other dir\\index.html'
```
The `posix` version doubles the directory separator while the `win32` version does not.
I know 0.12 is a stable branch, so even though I'd like to fix the posix version to work the same as it's win32 counterpart, I'm guessing that won't be accepted in this brach? Or is it small enough that this would be okay?

misterdjules:
@woollybogger If we want to make changes to `path.format`, let's do that on master and in another issue/PR. The documentation in v0.12 doesn't mention a specific behavior for any platform, and it's not clear how many users currently rely on that behavior.

Thank you very much again!

woollybogger:
@misterdjules Added a new commit with the tests and a couple other changes related to `path.parse` and `path.format`. If it's good I'll squash the commits.

misterdjules:
Thank you again @woollybogger. Added some comments, please let me know what you think. 

For some of the more controversial changes, we can always tackle them later in separate PRs. That way we can land the improvements you already made to the path's module performance and take more time to discuss the rest.

woollybogger:
@misterdjules I removed the controversial code and squashed into one commit.

misterdjules:
@woollybogger LGTM, thank you very much! :+1:

misterdjules:
Landed in c66f8c21f0646aaffd59ddc332381c6824df3e5a.

PR #9301, opened 2015-02-28
tests: fix race in test-http-curl-chunk-problem

misterdjules:
This test setups two event listeners: one on a child process' exit event
, another for the same child process' stdandard output's 'data' event.
The data even listener writes to a stream, and the exit event listener
ends it.

Because the exit event can be emitted before the data event, there is a
chance that something will be written to the stream after it's ended,
and that an error is thrown.

This change makes the test end the stream in the listener for the child
process' standard output's end event, which is guaranteed to be emitted
after the last data event, thus avoiding the race.

### DIFF
--- a/test/simple/test-http-curl-chunk-problem.js
+++ b/test/simple/test-http-curl-chunk-problem.js
@@ -69,13 +69,16 @@ var server = http.createServer(function(req, res) {
     res.write(data);
   });
 
+  cat.stdout.on('end', function onStdoutEnd() {
+    res.end();
+  });
+
   // End the response on exit (and log errors)
   cat.on('exit', function(code) {
     if (code !== 0) {
       console.error('subprocess exited with code ' + code);
       process.exit(1);
     }
-    res.end();
   });
 
 });


misterdjules:
This should be backported to v0.10, but there are other issues for v0.10 (see http://jenkins.nodejs.org/job/node-review-windows-julien/DESTCPU=ia32,label=windows/lastBuild/tapTestReport/simple.tap-240/) in this test that require more changes, so I'll fix that for v0.10 in another PR.

mdawsonibm:
FYI - This test was failing consistently on one of our internal test machines, applying the patch resolved the issue.

misterdjules:
@mdawsonibm Thanks for the heads up!

If you have some time to review the code change and if you think it's the best way to fix it, could you please add "LGTM" in a comment?

Thanks again!

cjihrig:
LGTM

mdawsonibm:
LGTM

piscisaureus:
Another solution to wait for the "close" event on the child process object. It's guaranteed to fire after both the process has exited *and* all the stdio streams are closed.

But this works too - lgtm.

piscisaureus:
Landed in io.js - https://github.com/iojs/io.js/commit/1009130495b4c52d00deed0ee2292039c3f36310

jasnell:
LGTM

misterdjules:
@piscisaureus Indeed, that was my first solution, and then I thought that this one was more explicit. Thanks for bringing that up!

misterdjules:
Thank you, landed in d9a309f718d738eb817195118f4daa06fb6f68ee!

PR #9302, opened 2015-02-28
src: enable strict mode in all builtin modules

misterdjules:
This is a follow-up commit for b233131901dea47132b0748d8e4d6bfcbba28abe.

It enables strict mode in all built-in modules. This will break existing code, but the goal is to progressively fix as many issues as possible that can arise when using `--use-strict`.

It could probably break code that loads these modules within a browser, since any of the `'use strict';` directive would turn all code into strict mode. I'm not sure how much we support this use case though, any information on that would be much appreciated.

### DIFF
--- a/deps/debugger-agent/lib/_debugger_agent.js
+++ b/deps/debugger-agent/lib/_debugger_agent.js
@@ -1,3 +1,5 @@
+'use strict';
+
 var assert = require('assert');
 var net = require('net');
 var util = require('util');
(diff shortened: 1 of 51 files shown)


misterdjules:
Tests started [on UNICEs](http://jenkins.nodejs.org/job/node-test-commit-unix/21/) and [on Windows](http://jenkins.nodejs.org/job/node-test-commit-windows/20/).

cjihrig:
LGTM

trevnorris:
LGTM.

misterdjules:
Thank you @cjihrig and @trevnorris for the review! Landed in ef4344311e19a4f73c031508252b21712b22fe8a.

PR #9311, opened 2015-03-01
http: emit abort event from ClientRequest

### LINKED ISSUE #9278: http: add abort event request()
There doesn't seem to be a good way (or at least I haven't found one) to detect when `ClientRequest.prototype.abort()` is called. There is `ClientRequest.prototype.aborted`, but that would require observing the object. It would be nice to have an `abort` event emitted in this case.

As a test case, I would like to detect the abort in the following code:

```javascript
var http = require('http');
var server = http.createServer(function(req, res) {
  res.end();
});

server.listen(4000, function() {
  function connect(cb) {
    var req = http.request({
      port: 4000
    }, cb);

    return req;
  }

  var req = connect(function() {
    assert(false);
  });


  req.abort();
});
```

cjihrig:
`ClientRequest` will now emit an `abort` event the first time `abort()` is called.

cc: @tjfontaine 
Closes #9278

### DIFF
--- a/doc/api/http.markdown
+++ b/doc/api/http.markdown
@@ -855,7 +855,14 @@ Emitted when the server sends a '100 Continue' HTTP response, usually because
 the request contained 'Expect: 100-continue'. This is an instruction that
 the client should send the request body.
 
-### request.flushHeaders()
+### Event: 'abort'
+
+`function () { }`
+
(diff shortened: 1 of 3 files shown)


jasnell:
LGTM

cjihrig:
This is already in io.js. Deferring to the converged repo.

Issue #9022, opened 2015-01-13
docs: emitter.listeners return a copy of original array rather than an array reference


StevenRoan:
As title, the document use the description 
- **Returns an array of listeners for the specified event.**

However, return an array usually indicating return an array reference, and here is different.
Maybe we can consider to modify this part of document, Thanks!

Example code
```
var EventEmitter = require('events').EventEmitter;
var eventEmitter = new EventEmitter();
var arrListeners = eventEmitter.listeners('test-event');
console.log(arrListeners.length);
eventEmitter.addListener('test-event', function(){
  //Do sth.
});
console.log(arrListeners.length); // log 0 because the this array reference is different with current one. 
arrListeners = eventEmitter.listeners('test-event');
console.log(arrListeners.length);
```

get output
0
0
1

BSteephenson:
https://github.com/joyent/node/blob/master/lib/events.js#L155
addListener is optimized to not even make an array if there's only one listener. Calling .listeners(), it just makes an array on the fly. Changing the documentation rather than the feature makes more sense.

PR #9321, opened 2015-03-03
docs:events clarify that emitter.listener() returns a copy

bsteephenson:
Clarifies that emitter.listener() returns a copy, not a reference
Resolves issue #9022

### DIFF
--- a/doc/api/events.markdown
+++ b/doc/api/events.markdown
@@ -104,7 +104,7 @@ Note that `emitter.setMaxListeners(n)` still has precedence over
 
 ### emitter.listeners(event)
 
-Returns an array of listeners for the specified event.
+Returns a copy of the array of listeners for the specified event.
 
     server.on('connection', function (stream) {
       console.log('someone connected!');


doc/api/events.markdown
```diff
@@ -104,7 +104,7 @@ Note that `emitter.setMaxListeners(n)` still has precedence over
 
 ### emitter.listeners(event)
 
-Returns an array of listeners for the specified event.
+Returns a copy of an array of listeners for the specified event.
```
cjihrig: Maybe change this to `Returns a copy of the array...`

BSteephenson: That would be lying though. Sometimes that array doesn't exist until you call .listeners(event)

cjihrig: That seems like an implementation detail that shouldn't be relevant to a user of the API.

jasnell:
LGTM ... happy with either "an" or "the"

BSteephenson:
I'll change it to "the" haha. Let me just figure out how to make the change without adding an extra commit.

jasnell:
LGTM. Will land shortly in a batch with other doc improvements...

jasnell:
Closing this. Landed in nodejs/node

Issue #9294, opened 2015-02-26
EADDRINUSE does not specify address


kumavis:
My error logs always look like:

```
events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: listen EADDRINUSE
    at errnoException (net.js:905:11)
    at Server._listen2 (net.js:1043:14)
    at listen (net.js:1065:10)
    at net.js:1147:9
    at dns.js:72:18
    at process._tickCallback (node.js:442:13)
```

I understand what it means, but could the error please include the address attempted to be used?

kumavis:
thanks :cat:

jasnell:
This uses util._errnoException to look up the system error. In theory, _errnoException could be modified to accept additional information but that would be an API change, meaning it wouldn't hit until at least v0.13. Would need to check with @joyent/node-coreteam to see what they would think about this kind of change.

That said, if you enable debugging, the debugging output ought to give you all of the information you need and you can obviously catch the error and output those details yourself.

kumavis:
thanks for the response. if i don't know what port it is, than its usually happening somewhere in someone else's code, and harder to catch.

jasnell:
Yes, I agree that's a challenge. Let's see what the core team says. If they feel it would be appropriate to extend _errnoException with additional data then that's a quick addition I can make. Like I said, however, given that it's an API change, it wouldn't land until v0.13 at least

trevnorris:
Interesting idea to automatically append all known fields added to the error. For example, if this error was caught then the port could be seen by `er.port`. I'm +/-0 right now. Anyone else?

cjihrig:
io.js aded `util._exceptionWithHostPort()`, which provides some additional information on the error. I don't think that's exactly the same thing being discussed here, but it's a starting point.

misterdjules:
@jasnell I believe that @kumavis got this error with node v0.10.x, not with node v0.12.x. In node v0.10.x, this error is thrown via [an internal `errnoException` function](https://github.com/joyent/node/blob/v0.10/lib/net.js#L901-L909). So fixing this in v0.10.x would not be an API change.

For v0.12.x, is `util._errnoException` considered as part of the public API?

misterdjules:
@jasnell Ok, I think I understand what you meant by an API change. Adding properties to the error object would definitely be an API change.

However my understanding of what @kumavis meant in his original comment is that the _message_ that is written to the output should have more information, not the error object itself.

[...]

PR #9336, opened 2015-03-05
dgram and net: better error messages

jasnell:
Port of the io.js util._exceptionWithHostPort addition.
Slight change to omit the ex.address and ex.port from the generated
error to avoid the API change. These are commented out for now so
they can be added back in easily later. Pulled over the changes made
to dgram and net to use this. Added an additional
test-net-server-errmsg
test as additional verification

see: https://github.com/iojs/io.js/pull/250 for reference.

Addresses: https://github.com/joyent/node/issues/9294

/cc @joyent/node-coreteam

### DIFF
--- a/lib/dgram.js
+++ b/lib/dgram.js
@@ -37,6 +37,7 @@ var cluster = null;
 var dns = null;
 
 var errnoException = util._errnoException;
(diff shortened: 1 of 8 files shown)


jasnell:
Realized I hadn't rebase properly before hitting this one, just to be safe, redoing.

PR #9339, opened 2015-03-05
dgram and net: better error messages

jasnell:
Port of the io.js util._exceptionWithHostPort addition.
Slight change to omit the ex.address and ex.port from the generated
error to avoid the API change. These are commented out for now so
they can be added back in easily later. Pulled over the changes made
to dgram and net to use this. Added an additional
test-net-server-errmsg
test as additional verification

see: iojs/io.js#250 for reference.

Addresses: #9294

/cc @joyent/node-coreteam

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -749,3 +749,27 @@ exports._errnoException = function(err, syscall, original) {
   e.syscall = syscall;
   return e;
 };
+
+exports._exceptionWithHostPort = function(err,
+                                          syscall,
+                                          address,
+                                          port,
+                                          additional) {
+  var details;
+  if (port && port > 0) {
+    details = address + ':' + port;
+  } else {
+    details = address;
+  }
+
+  if (additional) {
+    details += ' - Local (' + additional + ')';
+  }
+
+  var ex = exports._errnoException(err, syscall, details);
+  //ex.address = address;
+  //if (port) {
+  //  ex.port = port;
+  //}
+  return ex;
+};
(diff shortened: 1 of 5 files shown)


trevnorris:
Since it's a straight port, I'll let the style nits go. LGTM.

lib/util.js
```diff
@@ -749,3 +749,27 @@ exports._errnoException = function(err, syscall, original) {
...
+  } else {
+    details = address;
+  }
+
+  if (additional) {
+    details += ' - Local (' + additional + ')';
+  }
+
+  var ex = exports._errnoException(err, syscall, details);
+  //ex.address = address;
```
misterdjules: Is there any reason why we'd want to keep this commented out code?

test/simple/test-dgram-error-message-address.js
```diff
@@ -0,0 +1,55 @@
...
+var dgram = require('dgram');
+
+// IPv4 Test
+var socket_ipv4 = dgram.createSocket('udp4');
+
+socket_ipv4.on('listening', assert.fail);
+
+socket_ipv4.on('error', common.mustCall(function(e) {
+  assert.equal(e.message, 'bind EADDRNOTAVAIL 1.1.1.1:' + common.PORT);
+  // assert.equal(e.address, '1.1.1.1');
```
misterdjules: Same as above regarding commented out code, and for the rest of the changes.

jasnell:
Going to close this. It's not critical and it's already landed in nodejs/node. Can revisit if someone really wants to get this into v0.12

PR #9345, opened 2015-03-06
Fix strict mode violation

phanect:
Function declaration is not allowed in function block in strict mode.

### DIFF
--- a/lib/crypto.js
+++ b/lib/crypto.js
@@ -596,7 +596,7 @@ exports.pbkdf2Sync = function(password, salt, iterations, keylen, digest) {
 function pbkdf2(password, salt, iterations, keylen, digest, callback) {
   var encoding = exports.DEFAULT_ENCODING;
 
-  function next(er, ret) {
+  var next = function (er, ret) {
     if (ret)
       ret = ret.toString(encoding);
     callback(er, ret);


cjihrig:
Is this actually causing any errors? The following code runs fine on current master:

```javascript
'use strict';
var crypto = require('crypto');

crypto.pbkdf2('secret', 'salt', 4096, 512, 'sha256', function(err, key) {
  assert.strictEqual(err, undefined);
  assert.ok(key instanceof Buffer);
});
```

phanect:
@cjihrig If your example code run with `--use_strict` option, it fails.

```
$ node --use_strict ./test.js # Above your code is in test.js 
crypto.js:604
    function next(er, ret) {
    ^^^^^^^^
SyntaxError: In strict mode code, functions can only be declared at top level or immediately within another function.
    at runInThisContext (node.js:741:18)
    at NativeModule.compile (node.js:804:14)
    at Function.NativeModule.require (node.js:774:18)
    at Function.Module._load (module.js:295:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/home/phanect/Desktop/test.js:3:14)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
```

cjihrig:
Are you sure you're using the latest code on the master branch? I can't reproduce.

```
$ node -v
v0.13.0-pre
$ node --use_strict ./test.js
```

Your error occurs at crypto.js:604. That function is no longer on that line in [master](https://github.com/joyent/node/blob/master/lib/crypto.js#L604), but it was when [0.12](https://github.com/joyent/node/blob/v0.12.0-release/lib/crypto.js#L604) was released.

phanect:
Oops, sorry. I found I used 0.12.0.
I tried latest master and it worked fine.

Issue #6439, opened 2013-10-30
Compilation of v0.11.8 fails on Solaris 10 and some Illumos distros due to missing libproc.h


ptribble:
Building recent v0.11.X source on Solaris 10 fails with:

../deps/mdb_v8/mdb_v8.c:53:21: fatal error: libproc.h: No such file or directory
compilation terminated.

That's expected. The interfaces in libproc.h are private and the header file is explicitly excluded from packaging, so this will be true of Solaris and most Illumos-based distros. (I suspect you don't realize this for SmartOS, because it ends up in the proto area.)

I can work around this by explicitly turning off node_use_mdb in the configure script, but this ought to be handled properly.

jasnell:
@ptribble @davepacheco ... any updates on this one?

ptribble:
I'm manually patching configure to disable mdb on Solaris 10, which seems to work.

On other illumos variants I'm temporarily dropping in a copy of libproc.h (and currently adding an unversioned symlink for libavl.so.1, although that should shortly be unnecessary). Ideally the fix would be on the illumos side, to make libproc a public and committed interface.

rmustacc:
Which is something I'm slowly working on from the illumos side.

dago:
Would it be possible to merge the above commits to a release version?

jasnell:
@joyent/node-tsc @orangemocha @misterdjules ... any thoughts? should we land this here or defer to the converged repo?

cgalibern:
I have proposed this pull request #9349 , it allow configure --without-mdb.
Is it possible to merge it ?
thanks

jasnell:
@misterdjules @orangemocha ... what do you think?

misterdjules:
@jasnell #9349 seems like it would be useful.

orangemocha:
I am really out of my depth here, but would it make sense to set node_use_mdb to false by default for Solaris? It's the only platform for which it's set to true.

The original change is here: https://github.com/joyent/node/commit/e851fef60f5368927ea0a6b7bfc984ef18e93d6f

Also, https://github.com/joyent/node/pull/9349 is against master. It should probably be rebased against v0.12.

misterdjules:
@orangemocha Let's continue the discussion in #9349.

misterdjules:
Added the P-1 label because it prevents Solaris and derivatives' users from building Node.js on their platform.

ip1981:
libproc is used to be a private interface.

[...]

PR #9349, opened 2015-03-06
configure: add --without-mdb flag (fix issue #6439)

cgalibern:
We need a way to disable node_use_mdb on non Illumos derivated, because of
missing libproc.h.

Example:
    ./configure --dest-cpu=x64  --dest-os=solaris --prefix=/opt/node --without-mdb
    creating  ./icu_config.gypi
    { 'target_defaults': { 'cflags': [],
                           'default_configuration': 'Release',
                           'defines': [],
                           'include_dirs': [],
                           'libraries': []},
      'variables': { 'clang': 0,
                     'gcc_version': 48,
                     'host_arch': 'ia32',
                     'icu_small': 'false',
                     'node_install_npm': 'true',
                     'node_prefix': '/opt/node',
                     'node_shared_cares': 'false',
                     'node_shared_http_parser': 'false',
                     'node_shared_libuv': 'false',
                     'node_shared_openssl': 'false',
                     'node_shared_v8': 'false',
                     'node_shared_zlib': 'false',
                     'node_tag': '',
                     'node_use_dtrace': 'true',
                     'node_use_etw': 'false',
                     'node_use_mdb': 'false',
                     'node_use_openssl': 'true',
                     'node_use_perfctr': 'false',
                     'openssl_no_asm': 0,
                     'python': '/usr/bin/python',
                     'target_arch': 'x64',
                     'uv_library': 'static_library',
                     'uv_parent_path': '/deps/uv/',
                     'uv_use_dtrace': 'true',
                     'v8_enable_gdbjit': 0,
                     'v8_enable_i18n_support': 0,
                     'v8_no_strict_aliasing': 1,
                     'v8_optimized_debug': 0,
                     'v8_random_seed': 0,
                     'v8_use_snapshot': 'true',
                     'want_separate_host_toolset': 1}}
    creating  ./config.gypi
    creating  ./config.mk

### DIFF
--- a/configure
+++ b/configure
@@ -282,6 +282,11 @@ parser.add_option('--without-etw',
     dest='without_etw',
     help='build without ETW')
 
+parser.add_option('--without-mdb',
+    action='store_true',
+    dest='without_mdb',
+    help='build without mdb')
+
 parser.add_option('--without-npm',
     action='store_true',
     dest='without_npm',
@@ -554,7 +559,7 @@ def configure_node(o):
   # if we're on illumos based systems wrap the helper library into the
   # executable
   if flavor == 'solaris':
-    o['variables']['node_use_mdb'] = 'true'
+    o['variables']['node_use_mdb'] = b(not options.without_mdb)
   else:
     o['variables']['node_use_mdb'] = 'false'
 


misterdjules:
Adding @orangemocha's comments from #6439 here to avoid confusion:

> I am really out of my depth here, but would it make sense to set node_use_mdb to false by default for Solaris? It's the only platform for which it's set to true.

> The original change is here: e851fef

> Also, #9349 is against master. It should probably be rebased against v0.12.

@orangemocha Changing the default value for `node_use_mdb` would be a breaking change, as it would require packagers and users to change how they build node.js on SmartOS, I don't think we should do that in v0.12.

misterdjules:
Added the P-1 label because it prevents Solaris and derivatives' users from building Node.js on their platform.

orangemocha:
@misterdjules : sure, I defer to your expertise on the matter since I have none when it comes to these platforms :)

misterdjules:
LGTM. Thank you @cgalibern :+1:

misterdjules:
Tested on SmartOS with `--without-mdb` and with the default options and verified that it works as expected. Also created https://github.com/joyent/node/pull/25707 that is identical, but with a commit message that follows our guidelines so that I can land this with our node-accept-pull-request job on Jenkins.

PR #9356, opened 2015-03-08
build: use doc-tool module

robertkowalski:
delete old checked in doc-tool. run `npm i -g node-doc-generator`
before building the docs.

all the best from newark :)


robertkowalski:
@misterdjules i don't remember how you double checked if something for the doc changed exactly (it was a rsync dry run) can you send me a hint - so far i just verified by looking at the results and by writing the tests

misterdjules:
@robertkowalski I use ` rsync -v -n -c -u -r out/doc/ node@nodejs.org:~/web/nodejs.org/` to check what changes would be deployed after making changes to the API docs.

misterdjules:
Same comments as in https://github.com/joyent/node-website/pull/84:

Instead of invoking a global `node-doc-generator`, would it be better to install it locally and update/install it every time the documentation is generated so that we don't run the risk of generating documentation with an outdated generator?

jasnell:
@robertkowalski @misterdjules ... any reason to keep this open here? can we move to one of the other repos?

robertkowalski:
good point, closing

PR #9366, opened 2015-03-10
net: use cached peername to resolve remote props

### LINKED ISSUE #9287: socket.remoteAddress after close is undefined
`socket.remoteAddress` after `close` is undefined because `this._handle` is null. Since we're caching `this._peername` can we just change `_getpeername` to return `this._peername` if it already exists and not always look for a `this._handle`.

Related stackoverflow posts:
http://stackoverflow.com/questions/12444598/why-is-socket-remoteaddress-undefined-on-end-event
http://stackoverflow.com/questions/12126403/getting-the-ip-address-of-a-socket-inside-a-close-event-handler

jameshartig:
Allows socket.remote\* properties to still be accessed even after the
socket is closed. Fixes #9287

### DIFF
--- a/test/simple/test-net-remote-address-port.js
+++ b/test/simple/test-net-remote-address-port.js
@@ -41,6 +41,10 @@ var server = net.createServer(function(socket) {
   socket.on('end', function() {
     if (++conns_closed == 2) server.close();
   });
+  socket.on('close', function() {
+    assert.notEqual(-1, remoteAddrCandidates.indexOf(socket.remoteAddress));
+    assert.notEqual(-1, remoteFamilyCandidates.indexOf(socket.remoteFamily));
+  });
   socket.resume();
 });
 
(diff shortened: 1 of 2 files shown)


test/simple/test-net-remote-address-port.js
```diff
@@ -53,12 +57,20 @@ server.listen(common.PORT, 'localhost', function() {
     assert.equal(common.PORT, client.remotePort);
     client.end();
   });
+  client.on('close', function() {
+    assert.notEqual(-1, remoteAddrCandidates.indexOf(client2.remoteAddress));
```
cjihrig: Should these two lines be checking `client` instead of `client2`?

cjihrig:
One question about the test. I haven't run the code, but if the CI is happy, then this LGTM.

test/simple/test-net-remote-address-port.js
fastest963: Good catch. Woops!

fastest963:
Good catch. Fixed the test and it still passes.

cjihrig:
Thanks! Landed in8c38b07252d86f4eef91344f8df4f0c3e38bca35.

jasnell:
@cjihrig ... do you happen to know why this was reopened?

cjihrig:
I think that was a malfunction in the TJ Fontaine bot.

PR #9372, opened 2015-03-10
prevented overwriting encoding when chunk is a buffer in writeOrBuffer

jandrieu:
_stream_writable ignores the incoming encoding in writeOrBuffer if the chunk is a Buffer. That led to "unknown encoding' errors later, despite the chunk initially having a good encoding. 

FWIW, upstream was a process.stdin with encoding to utf8. That encoding was getting lost in writeOrBuffer.


jandrieu:
FWIW, I can confirm this bug goes back to at least v0.10.32. It is prevening couchdb/nodejs integration with couchjs.

jasnell:
@chrisdickinson ... thoughts on this one?

PR #9381, opened 2015-03-11
test: Environment varible to specify directory for pipes

mhdawson:
At the uv layer pipes are connected with uv_pipe_connect.
The current spec for this method indicates that the maximum
length is limited to the size of length of
sizeof(sockaddr_un.sun_path), typically between 92 and
108 bytes. Anything longer than that just gets truncated.

The simple testsuite currently creates pipes in directories
under the directory where node was built.  In our jenkins
jobs this sometimes ends up being a deep enough path that
the path for the pipes is getting truncated.  The result
is that tests using pipes fail with errors that don't
make it obvious what the problem is.

Even if the errors were helpful, we still need a way
to avoid the truncation.

This patch adds the environment variable NODE_PIPE_DIR.
If set the tests create pipes in this directory instead of
the current defaults.  In addition the test harness is
updated to remove/delete this directory before/after
each test is run.

```
modified:   test/common.js
modified:   test/simple/test-net-pipe-connect-errors.js
modified:   test/testpy/__init__.py
modified:   test/simple/test-cluster-eaccess.js
```

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -28,6 +28,11 @@ exports.testDir = path.dirname(__filename);
 exports.fixturesDir = path.join(exports.testDir, 'fixtures');
 exports.libDir = path.join(exports.testDir, '../lib');
 exports.tmpDir = path.join(exports.testDir, 'tmp');
+if (process.env.NODE_PIPE_DIR === undefined) {
+  exports.pipeTmpDir = exports.tmpDir;
+} else {
+  exports.pipeTmpDir = path.join(process.env.NODE_PIPE_DIR, 'NodePipeTmp');
+}
 exports.PORT = +process.env.NODE_COMMON_PORT || 12346;
 
 exports.opensslCli = path.join(path.dirname(process.execPath), 'openssl-cli');
@@ -35,7 +40,7 @@ if (process.platform === 'win32') {
   exports.PIPE = '\\\\.\\pipe\\libuv-test';
   exports.opensslCli += '.exe';
 } else {
-  exports.PIPE = exports.tmpDir + '/test.sock';
+  exports.PIPE = exports.pipeTmpDir + '/test.sock';
 }
 if (!fs.existsSync(exports.opensslCli))
   exports.opensslCli = false;
--- a/test/simple/test-cluster-eaccess.js
+++ b/test/simple/test-cluster-eaccess.js
@@ -27,7 +27,12 @@ var path = require('path');
 var fs = require('fs');
 var net = require('net');
 
-var socketPath = path.join(common.fixturesDir, 'socket-path');
+var socketPath = common.PIPE;
+if (process.platform === 'win32') {
+  // for windows pipes are handled specially and we can't write to
+  // common.PIPE with fs.writeFileSync like we can on other platforms
+  socketPath = path.join(common.fixturesDir, 'socket-path');
+}
 
 if (cluster.isMaster) {
   var worker = cluster.fork();
--- a/test/simple/test-net-pipe-connect-errors.js
+++ b/test/simple/test-net-pipe-connect-errors.js
@@ -31,7 +31,9 @@ var accessErrorFired = false;
 
 // Test if ENOTSOCK is fired when trying to connect to a file which is not
 // a socket.
-var emptyTxt = path.join(common.fixturesDir, 'empty.txt');
+var emptyTxt = path.join(common.pipeTmpDir, 'empty.txt');
+fs.closeSync(fs.openSync(emptyTxt, 'w'));
+
 var notSocketClient = net.createConnection(emptyTxt, function() {
   assert.ok(false);
 });
--- a/test/testpy/__init__.py
+++ b/test/testpy/__init__.py
@@ -48,16 +48,22 @@ class SimpleTestCase(test.TestCase):
     self.mode = mode
     self.tmpdir = join(dirname(self.config.root), 'tmp')
     self.additional_flags = additional
+    if "NODE_PIPE_DIR" in os.environ:
+      self.pipeTmpDir = join(os.environ["NODE_PIPE_DIR"], 'NodePipeTmp')
   
   def AfterRun(self, result):
     # delete the whole tmp dir
     try:
       rmtree(self.tmpdir)
+      if self.pipeTmpDir:
+        rmtree(self.pipeTmpDir);
     except:
       pass
     # make it again.
     try:
       mkdir(self.tmpdir)
+      if self.pipeTmpDir:
+        mkdir(self.pipeTmpDir);
     except:
       pass
 
@@ -65,6 +71,8 @@ class SimpleTestCase(test.TestCase):
     # delete the whole tmp dir
     try:
       rmtree(self.tmpdir)
+      if self.pipeTmpDir:
+        rmtree(self.pipeTmpDir);
     except:
       pass
     # make it again.
@@ -72,6 +80,8 @@ class SimpleTestCase(test.TestCase):
     while not os.path.exists(self.tmpdir):
       try:
         mkdir(self.tmpdir)
+        if self.pipeTmpDir:
+          mkdir(self.pipeTmpDir);
       except:
         pass
   


test/common.js
```diff
@@ -35,7 +35,11 @@ if (process.platform === 'win32') {
   exports.PIPE = '\\\\.\\pipe\\libuv-test';
   exports.opensslCli += '.exe';
 } else {
-  exports.PIPE = exports.tmpDir + '/test.sock';
+  if (typeof process.env.NODE_PIPE_DIR == 'undefined') {
```
cjihrig: Can you replace the `typeof` checks with direct comparisons to `undefined`?

cjihrig: This should be `===`.

test/simple/test-cluster-eaccess.js
```diff
@@ -28,6 +28,9 @@ var fs = require('fs');
 var net = require('net');
 
 var socketPath = path.join(common.fixturesDir, 'socket-path');
```
misterdjules: Can this test use `common.PIPE` also instead of  `socketPath`?

test/testpy/__init__.py
```diff
@@ -53,11 +53,15 @@ def AfterRun(self, result):
     # delete the whole tmp dir
     try:
       rmtree(self.tmpdir)
+      if "NODE_PIPE_DIR" in os.environ:
+        rmtree(os.environ["NODE_PIPE_DIR"]);
```
misterdjules: I'm concerned with what happens if users set `NODE_PIPE_DIR` to a directory that contains other data in it that is not temporary and that is not related to Node.js' tests.

Maybe we should always generate a temporary directory for pipes using something like `mktemp -d -t prefix`? I'm not sure how to do that on Windows though.

misterdjules:
@mdawsonibm As discussed during the core team meeting, please add this to the 0.12.1 or 0.12.2 milestone as you see fit.

test/testpy/__init__.py
mdawsonibm: Another wrinkle is that mktemp as an executable that can be run from the shell does not exist. 
For windows it looked to me like it already used a different directory for pips

test/simple/test-cluster-eaccess.js
mdawsonibm: I don't know the history of the test but I don't see why not.

test/testpy/__init__.py
misterdjules: Right on Windows it already uses a directory that is outside of the source directory, so we're fine. But on all other supported platforms, we should be able to use `mktemp` or something similar, or am I missing something?

mdawsonibm: I missed out  above that its on AIX that mktemp does not exist as an executable on AIX and I've not found a good alternative yet.

mdawsonibm: One alternative if we are worried about deleting the contents of the directory is to just delete the files we create.  If we think that would be ok I can try that out.

mdawsonibm: Or we could create a sub-directory in the directory specified by NODE_PIPE_DIR and only delete that sub-directory.

misterdjules: @mdawsonibm Yes, I think any of the two solutions you mentioned would work, with a preference for the sub-directory since it would probably require less maintenance over time (no need to track which files need to be removed if the list of files created changes).

mdawsonibm: Ok, address comments to date and validate all works as expected

mdawsonibm:
Updated per comments and all tests pass on linux with/without NODE_PIPE_DIR in a shallow directory.  In a deep directory all pass with NODE_PIPE_DIR and without some fail as expected.

test/testpy/__init__.py
```diff
@@ -53,11 +53,15 @@ def AfterRun(self, result):
     # delete the whole tmp dir
     try:
       rmtree(self.tmpdir)
+      if "NODE_PIPE_DIR" in os.environ:
+        rmtree(join(os.environ["NODE_PIPE_DIR"], 'NodePipeTmp'));
```
misterdjules: Can we avoid the repetition of `join(os.environ["NODE_PIPE_DIR"], 'NodePipeTmp')` by having a `self.pipeTmpDir` member variable?

test/simple/test-net-pipe-connect-errors.js
```diff
@@ -32,6 +32,11 @@ var accessErrorFired = false;
 // Test if ENOTSOCK is fired when trying to connect to a file which is not
 // a socket.
 var emptyTxt = path.join(common.fixturesDir, 'empty.txt');
+if (process.env.NODE_PIPE_DIR !== undefined) {
```
misterdjules: We might be able to always use:
```
emptyTxt = path.join(common.pipeTmpDir, 'empty.txt');
fs.closeSync(fs.openSync(emptyTxt, 'w'));
```
and never use the `empty.txt` file from the fixtures directory.

mdawsonibm: makes sense to me.

test/testpy/__init__.py
mdawsonibm: will do

test/testpy/__init__.py
```diff
@@ -48,16 +48,22 @@ def __init__(self, path, file, mode, context, config, additional=[]):
     self.mode = mode
     self.tmpdir = join(dirname(self.config.root), 'tmp')
     self.additional_flags = additional
+    if "NODE_PIPE_DIR" in os.environ:
+      self.pipeTmpDir = join(os.environ["NODE_PIPE_DIR"], 'NodePipeTmp')
```
misterdjules: `self.pipeTmpDir = None

misterdjules:
@mdawsonibm Once this comment: https://github.com/joyent/node/pull/9381/files#r26866770 is addressed, LGTM :+1: Thank you very much!

mdawsonibm:
When I ran the accept job  test-cluster-eaccess failed on windows 32bit.  I think its because what's used for common.PIPE ('\\\\.\\pipe\\libuv-test')  is special and can't be used by fs.writeFileSync and we end up with this error:

  #Error: ENOENT, no such file or directory '\\.\pipe\libuv-test'

It seems like test already special cases for windows  32 in that it expects a different error.  

I think it makes the most sense to special case what we use for socketPatch for windows. In this way we share the common solution to the depth paths on non-windows were we say the original problem while keeping the test on windows running in the way it did perviously

mdawsonibm:
Pushed additional change to address failure seen on windows

test/simple/test-cluster-eaccess.js
```diff
@@ -27,7 +27,12 @@ var path = require('path');
 var fs = require('fs');
 var net = require('net');
 
-var socketPath = path.join(common.fixturesDir, 'socket-path');
+var socketPath = common.PIPE;
+if (process.platform === 'win32') {
+  // for windows pipes are handled specially and we can't write to
+  // common.PIPE with fs.writeFileSync like we can on other platforms
+  socketPath = path.join(common.fixturesDir, 'socket-path');
+}
```
misterdjules: If this test was not failing before when run in a deep directory and we can't use `common.PIPE` on all platforms, then I suggest reverting to the old behavior of using `path.join(common.fixturesDir, 'socket-path');` on all platforms. Sorry for the confusion.

test/testpy/__init__.py
misterdjules: `"NODE_PIPE_DIR"` is still hardcoded in several places, the approach suggested in [my initial comment](https://github.com/joyent/node/pull/9381/files#r26866770) avoids that.

misterdjules:
Added two comments, can you please take a look and let me know what you think? Also, we'll need to squash these commits into just one before we merge them with the node-accept-pull-request Jenkins job.

mdawsonibm:
It was failing on unix due to the deep paths, that is why I modified it in the first patch.  On windows it would have been ok because its not actually a pipe but a directory that's being used.

misterdjules:
@mdawsonibm OK, thank you for the clarification. Did you have some time to take a look at this comment: https://github.com/joyent/node/pull/9381#discussion_r27139364?

mdawsonibm:
Sorry did not look far up enough to see that one.  Will update  and squash into one commit.

mdawsonibm:
Ok updated to reflect that comment and squashed into 1 commit

misterdjules:
Thank you @mdawsonibm! The first line of the commit should be 50 columns or less, as [indicated in the contribution guidelines](https://github.com/joyent/node/blob/master/CONTRIBUTING.md#commit). Otherwise, LGTM :+1:

mdawsonibm:
Ok shrunk length of first line, will land now

mdawsonibm:
landed as 5dd5ce75d7d9a1470f4e35b6b4046bc933714f2d

PR #9410, opened 2015-03-13
watchdog: fix timeout for early polling return

mhdawson:
Switch from running the loop with UV_RUN_ONCE to UV_RUN_DEFAULT, because
it's possible that the poll returns earlier than expected and thus the
timer is not run on a single interation.

The loop is not stopped either from the timer callback or from the async
handle's.

PR-URL: https://github.com/iojs/io.js/pull/622
Reviewed-By: Ben Noordhuis info@bnoordhuis.nl
Reviewed-By: Rod Vagg rod@vagg.org

### DIFF
--- a/src/node_watchdog.cc
+++ b/src/node_watchdog.cc
@@ -87,8 +87,9 @@ void Watchdog::Destroy() {
 void Watchdog::Run(void* arg) {
   Watchdog* wd = static_cast<Watchdog*>(arg);
 
-  // UV_RUN_ONCE so async_ or timer_ wakeup exits uv_run() call.
-  uv_run(wd->loop_, UV_RUN_ONCE);
+  // UV_RUN_DEFAULT the loop will be stopped either by the async or the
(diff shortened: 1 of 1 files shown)


misterdjules:
LGTM.

misterdjules:
Also, please add this to the appropriate milestone (as well as the appropriate issue). This will allow anybody else looking at the issues/PRs in that milestone to also find it an review it, and will communicate in which release we want this to be merged.

Currently we have two 0.12.x milestones, 0.12.1 and 0.12.2. If we want to integrate this fix asap and thus have people review it asap, let's add it to 0.12.1, otherwise let's add it to 0.12.2.

Even if it's merged quickly, sometimes it's handy to be able to look back to what had been done for a particular release on GitHub.

Thank you!

mdawsonibm:
Landed in 7e9d2f8de8194e5f4d950eb87288bf97b2f3e509

Issue #9404, opened 2015-03-12
url.parse does not handle IPv6 link-local addresses correctly


cthayer:
In version 0.12.0, `url.parse()` does not parse a URL containing an IPv6 link-local address correctly.

Specifically, adding a zone to the address in the host portion of the URL causes `url.parse()` to fail to detect the host portion of the address and considers everything after the protocol to be part of the path.
#### Example of failure with IPv6 zone:

``` javascript
var url = require('url');

url.parse('ws://[fe80::2e0:4cff:fe68:23ff%en7]:51877/engine.io/?EIO=3&transport=websocket');
{ protocol: 'ws:',
  slashes: true,
  auth: null,
  host: '',
  port: null,
  hostname: '',
  hash: null,
  search: '?EIO=3&transport=websocket',
  query: 'EIO=3&transport=websocket',
  pathname: '/[fe80::2e0:4cff:fe68:23ff%en7]:51877/engine.io/',
  path: '/[fe80::2e0:4cff:fe68:23ff%en7]:51877/engine.io/?EIO=3&transport=websocket',
  href: 'ws:///[fe80::2e0:4cff:fe68:23ff%en7]:51877/engine.io/?EIO=3&transport=websocket' }
```
#### Example of success without the IPv6 zone:

``` javascript
var url = require('url');

url.parse('ws://[fe80::2e0:4cff:fe68:23ff]:51877/engine.io/?EIO=3&transport=websocket');
{ protocol: 'ws:',
  slashes: true,
  auth: null,
  host: '[fe80::2e0:4cff:fe68:23ff]:51877',
  port: '51877',
  hostname: 'fe80::2e0:4cff:fe68:23ff',
  hash: null,
  search: '?EIO=3&transport=websocket',
  query: 'EIO=3&transport=websocket',
  pathname: '/engine.io/',
  path: '/engine.io/?EIO=3&transport=websocket',
  href: 'ws://[fe80::2e0:4cff:fe68:23ff]:51877/engine.io/?EIO=3&transport=websocket' }
```
#### Expected behavior

I would expect `url.parse()` of `ws://[fe80::2e0:4cff:fe68:23ff%en7]:51877/engine.io/?EIO=3&transport=websocket` to produce the following result:

``` javascript
{ protocol: 'ws:',
  slashes: true,
  auth: null,
  host: '[fe80::2e0:4cff:fe68:23ff%en7]:51877',
  port: 51877,
[...]

PR #9411, opened 2015-03-14
url: % is a valid host character in IPv6 addresses

cthayer:
% is a valid host character in IPv6 link-local addresses.

The % separates the address from the zone within the hostname section of the URL.

This fixes issue #9404

### DIFF
--- a/lib/url.js
+++ b/lib/url.js
@@ -282,6 +282,15 @@ Url.prototype.parse = function(url, parseQueryString, slashesDenoteHost) {
           }
         }
       }
+    } else {
+      // IPv6 hostname
+      
+      // for link local addresses:
+      // - make sure the zone id is present after the %
+      // - make sure the hostname does not end in '%]'
+      if (this.hostname.indexOf('%]') !== -1) {
+        this.hostname.replace('%', '');
+      }
     }
 
     if (this.hostname.length > hostnameMaxLen) {
--- a/lib/url.js
+++ b/lib/url.js
@@ -70,6 +70,7 @@ var protocolPattern = /^([a-z0-9.+-]+:)/i,
     // are the ones that are *expected* to be seen, so we fast-path
     // them.
     nonHostChars = ['%', '/', '?', ';', '#'].concat(autoEscape),
+    nonHostCharsIPv6 = ['/', '?', ';', '#'].concat(autoEscape),
     hostEndingChars = ['/', '?', '#'],
     hostnameMaxLen = 255,
     hostnamePartPattern = /^[+a-z0-9A-Z_-]{0,63}$/,
@@ -216,11 +217,19 @@ Url.prototype.parse = function(url, parseQueryString, slashesDenoteHost) {
       rest = rest.slice(atSign + 1);
       this.auth = decodeURIComponent(auth);
     }
+    
+    // if hostname begins with [ and is followed by ]
+    // assume that it's an IPv6 address.
+    // :: is the smallest IPv6 address possible, so ] must be > 2 positions to the right of [
+    var ipv6Hostname = rest[0] === '[' &&
+        rest.indexOf(']') > 2;
 
     // the host is the remaining to the left of the first non-host char
+    // use the appropriate non-host char set
+    var invalidHostChars = ipv6Hostname ? nonHostCharsIPv6 : nonHostChars;
     hostEnd = -1;
-    for (var i = 0; i < nonHostChars.length; i++) {
-      var hec = rest.indexOf(nonHostChars[i]);
+    for (var i = 0; i < invalidHostChars.length; i++) {
+      var hec = rest.indexOf(invalidHostChars[i]);
       if (hec !== -1 && (hostEnd === -1 || hec < hostEnd))
         hostEnd = hec;
     }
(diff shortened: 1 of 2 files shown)


misterdjules:
Adding to milestone 0.12.2 for the same reason its corresponding issue [has been added to 0.12.2 too](https://github.com/joyent/node/issues/9404#issuecomment-78688847).

jbergstroem:
Isn't `%` supposed to be represented using percent-encoding (`%25`) in urls? http://tools.ietf.org/html/rfc6874#section-2

trevnorris:
Do any browsers support this?

bahamat:
Curl, at least, does.

````
$ curl -i 'http://[fe80::253:a3ff:fe13:7081%en4]:80/'
HTTP/1.1 200 OK
Date: Mon, 23 Mar 2015 23:34:29 GMT
Server: Apache/2.4.9 (Unix)
Content-Location: index.html.en
Vary: negotiate
TCN: choice
Last-Modified: Mon, 11 Jun 2007 18:53:14 GMT
ETag: "2d-432a5e4a73a80"
Accept-Ranges: bytes
Content-Length: 45
Content-Type: text/html

<html><body><h1>It works!</h1></body></html>
````

cthayer:
Node.js allows the % character as well (just use a `net.Socket` to try it out)

This is just an issue with `url.parse` correctly parsing a URL with an IPv6 Link Local address in it.

misterdjules:
Corresponding issues [in Chromium's issues tracker](https://code.google.com/p/chromium/issues/detail?id=70762) and [in Firefox's issue tracker](https://bugzilla.mozilla.org/show_bug.cgi?id=700999).

It seems that both of them are seriously considering implementing [RFC 6874](tools.ietf.org/html/rfc6874), but it hasn't been done yet.

Removing from the 0.12.4 milestone as currently it's not supported by most major browsers, and thus doesn't seem to be a pressing issue.

cthayer:
I must say that it's a little frustrating having feature support (or bug fixes) for a server side language dictated by browser support.

Currently, it's impossible to bind a server or connect a client to a link local address on the same machine when using any npm module that sanitizes it's URL's via `url.parse` because `url.parse` does not respect that everything within the `[` and `]` is the IPv6 address.

Take `url.parse` out of the equation and everything works.  To me, that means that node.js already supports RFC 6874 it's just `url.parse` that doesn't.

This would sit much better with me if my tiny changes were breaking tests or something else rather than this change not being implemented (or deemed important) because browsers don't support this yet.  Last I checked, node.js was a server side language and this issue essentially removes npm as a resource for a project I'm working on.

`</rant>`

Just wanted to get my thoughts/feelings out there.  Just frustrated as this issue is a big issue for me.

trevnorris:
If the 0.12.4 release is pressing then let's at least get the fix in for the 0.12.5 release.

misterdjules:
@cthayer @trevnorris v0.12.4 is not pressing, as we're currently working on v0.12.3. I don't have anything against adding this PR back to the 0.12.4 milestone if the impact of its changes are well understood. My understanding was that it was not the case.

I also initially thought we had more issues still open in the 0.12.3 and 0.12.4 milestones and didn't want to add much more to it except for major issues that are well understood, but looking at them again it seems there's still some room for other issues.

I apologize for the confusion.

trevnorris:
If this helps the parser more closely follow the spec, and it's only an additive change, then I don't see this merging this as an issue.

@misterdjules You cool if this is added to the 0.12.4 milestone?

misterdjules:
@trevnorris Yes, I'm ok with adding it to the 0.12.4 milestone as long as we consider it's not an addition to the API, but a bug fix. In my opinion it is a bug fix.

@trevnorris @cthayer As usual we still need to do a thorough review and after briefly looking at the code changes again I would already recommend adding more tests for invalid input, like an url with a valid IPv6 address, a `%` character but without a zone id.

cthayer:
@trevnorris @misterdjules Thanks for the additional input.  I'm glad to see that this is being considered for the 0.12.4 release again.  I'm happy to add more tests.

I hadn't done actual validation of the hostname portion when a IPv6 zone-id is involved because it didn't appear that `url.parse` does any validation of the hostname portion, rather it just finds the left boundary of the hostname portion and I was trying to make my change as minimal as possible.  So I know that my code would fail a test for an IPv6 address with a `%` and no zone-id!  :)

I'm all for making sure things are correct, so I'll add more tests and more logic for validating the zone-id.  Also, since I'm validating the zone-id, is it cool if I add a parameter to the output called `zoneId` that contains just the IPv6 zone-id (if present)?

misterdjules:
@cthayer Thank you for considering to add more tests! Just to make sure we're on the same page, I was not suggesting to validate IPv6 addresses if that's not already done for host names, just that a zone-id is provided after a `%` character if that's required by the RFC. In other words, having just one test seems like it's probably not enough.

I would suggest adding that additional `zoneId` property to the output in another PR targeted to `master`. Even if it wouldn't break backward compatibility, I'd rather keep changing the API on development versions, unless @trevnorris or @joyent/node-coreteam disagrees.

Thank you!

cthayer:
@misterdjules I've added a test (and supporting code) for a url with a valid IPv6 address, a `%` character but without a zone id.

My solution was to strip the `%` character from the hostname and continue on.  So, for example, `[::1%]` becomes `[::1]`.

I wasn't sure how much hostname validation you want to add, since there wasn't much there before, so I left it at that, but let me know if you'd like to be more thorough.

PR #9427, opened 2015-03-17
url: resolve urls with . and ..

### LINKED ISSUE #8992: url resolve(href, '.') sometimes leaves off trailing slash

misterdjules:
'.' and '..' are directory specs and resolving urls with or without the
hostname with '.' and '..' should add a trailing slash to the end of the
url.

Fixes #8992.


misterdjules:
@trevnorris @cjihrig After merging v0.10 into v0.12 and v0.12 in master, the [revert])(https://github.com/joyent/node/pull/9257) of the [original PR](https://github.com/joyent/node/pull/9010) has landed into v0.12 and master.

According to my [comment](https://github.com/joyent/node/pull/9257#issuecomment-76500324) a while ago, I suggest landing the [original PR](https://github.com/joyent/node/pull/9010) in master, which is why this PR is created against master.

trevnorris:
SGTM.

cjihrig:
Works for me too. LGTM

misterdjules:
@amir-s @trevnorris @cjihrig Thank you! Finally landed for good in 9b534e2 :+1:

Issue #9441, opened 2015-03-19
Upgrade OpenSSL to 1.0.1m


misterdjules:
See the [announcement of the upcoming 1.0.1m release](https://mta.openssl.org/pipermail/openssl-announce/2015-March/000020.html).

When the node v0.10.38 release that contains the OpenSSL upgrade is done, please merge into v0.12 and do the v0.12.1 release.

misterdjules:
@jasnell As discussed this morning, the plan for now is to first upgrade OpenSSL in v0.10, then to merge  v0.10 into v0.12.1 branched off of v0.12.0, and finally to merge v0.12.1-release into v0.12.

@joyent/node-coreteam Please let us know if this is correct.

jasnell:
Sounds good.

jasnell:
@joyent/node-coreteam ...

For review:
Patch against: v0.10.37-release -  https://github.com/jasnell/node/commit/8a917ca403e8a9e9c0b167018b17b628217a65a2 (my branch: https://github.com/jasnell/node/tree/v0.10.37-opensslm)

Patch against: v0.12.0-release - https://github.com/jasnell/node/commit/5cc430fb94392bbb50da8069900bb2f1b1c4ba25 (my branch: https://github.com/jasnell/node/tree/v0.12.0-opensslm)

The patch followed the process documented in the io.js openssl-m update (https://github.com/iojs/io.js/pull/1206) by @shigeki

Builds on Mac OSx 10.10 and Ubuntu 14.10. Tests pass. Needs additional review and verification.

shigeki:
I've just reviewed the v0.10.37-opensslm branch and verified there is no difference from my patch. It can work on Node-v0.10.x. LGTM.

jasnell:
@shigeki ... I appreciate your work. You made it significantly easier for
me :+1:

misterdjules:
@shigeki @jasnell Looking at the changes, I think it would be better to make changes to `deps/openssl/openssl.gyp` instead of changes to the OpenSSL source itself to work around the symlinked headers. That makes us have to maintain one less floating patch.

I would also like to keep at least these commits:
* https://github.com/shigeki/io.js/commit/a3e0da2e0c4ffcb7272a8a3f41d8772b0969e706.
* https://github.com/shigeki/io.js/commit/ac89f8ac5b036b857afb27d9d0e7bc78a8a732cd.

as separate commits so that we can track them separately from the rest of the changes.

How does that sound?

[...]

PR #9446, opened 2015-03-20
deps: update to openssl-1.0.1m

jasnell:
There are several updates pulled from:
  https://github.com/iojs/io.js/pull/1206
1. Expand the distribution tarball into deps/openssl/openssl
2. replace all headers in openssl

change all openssl/include/openssl/*.h to include resolved symbolic
links and openssl/crypto/opensslconf.h to refer config/opensslconf.h
1. deps: separate sha256/sha512-x86_64.pl for openssl

sha256-x86_64.pl does not exist in the origin openssl distribution. It
was copied from sha512-x86_64.pl and both sha256/sha512 scripts were
modified so as to generates only one asm file specified as its key
hash length.
1. fix openssl assembly error on ia32 win32

`x86masm.pl` was mistakenly using .486 instruction set, why `cpuid` (and
perhaps others) are requiring .686 .
1. remove vpaesni-x86_64.asm in x64-win32-masm

It is no longer used.
1. remove unused backup file in openssl asm

deps/openssl/asm/x64-win32-masm/x86_64cpuid.asm.orig was accidentally
committed before.
1. fix keypress requirement in apps on win32

reapply b910613792dac946b295855963869933a9089044

/cc: @joyent/node-coreteam


indutny:
LGTM, if it builds and tests are passing :)

misterdjules:
Pasting my comments from #9441 so that we can continue the discussion here:

@shigeki @jasnell Looking at the changes, I think it would be better to make changes to `deps/openssl/openssl.gyp` instead of changes to the OpenSSL source itself to work around the symlinked headers. That makes us have to maintain one less floating patch.

I would also like to keep at least these commits:
* https://github.com/shigeki/io.js/commit/a3e0da2e0c4ffcb7272a8a3f41d8772b0969e706.
* https://github.com/shigeki/io.js/commit/ac89f8ac5b036b857afb27d9d0e7bc78a8a732cd.

as separate commits so that we can track them separately from the rest of the changes.

How does that sound?

jasnell:
I'm good with whichever way you want to go.

misterdjules:
Closing in favor of #9451.

PR #9448, opened 2015-03-20
test: fix test-process-active-wraps for IPv4

CGavrila:
**Problem**

The test-process-active-wraps test found in test/simple/test-process-active-wraps.js fails if the machine it runs on does not have IPv6 enabled. This is due to the fact that Node first tries to use IPv6 and then falls back to using IPv4 if it is not enabled, as it can be seen in [lib/net.js#L1100-L1105](https://github.com/joyent/node/blob/9b534e2e87489d219f0d0cf5be7f9119704fe424/lib/net.js#L1100-L1105). Even though binding to an IPv6 address fails, there is an extra handle created in the process. This causes a couple of assertions in the test to fail.

**Fix**

The fix involves checking whether the server started on an IPv6 address or IPv4 address when deciding what to assert. If the protocol used is IPv4, then the numbers used for the assertions can simply be bumped by one.

As an example, this is how the code has been changed in the test:

```
...

var isIPv6 = true;

(function() {
  expect(0, 0);
  var server = net.createServer().listen(common.PORT);
  if (server.address().family !== 'IPv6') isIPv6 = false;
  isIPv6 ? expect(1, 0) : expect(2, 0);

...
```

### DIFF
--- a/test/simple/test-process-active-wraps.js
+++ b/test/simple/test-process-active-wraps.js
@@ -31,12 +31,17 @@ function expect(activeHandles, activeRequests) {
 
 var handles = [];
 
+/* Start from the assumption that IPv6 is enabled, just like Node does;
+ * if it is disabled, a different number of handles will be created. */
+var isIPv6 = true;
+
 (function() {
   expect(0, 0);
   var server = net.createServer().listen(common.PORT);
-  expect(1, 0);
+  if (server.address().family !== 'IPv6') isIPv6 = false;
+  isIPv6 ? expect(1, 0) : expect(2, 0);
   server.close();
-  expect(1, 0); // server handle doesn't shut down until next tick
+  isIPv6 ? expect(1, 0) : expect(2, 0); // server handle doesn't shut down until next tick
   handles.push(server);
 })();
 
@@ -47,13 +52,13 @@ var handles = [];
     });
   };
 
-  expect(1, 0);
+  isIPv6 ? expect(1, 0) : expect(2, 0);
   var conn = net.createConnection(common.PORT);
   conn.on('lookup', onlookup);
   conn.on('error', function() { assert(false); });
-  expect(2, 1);
+  isIPv6 ? expect(2, 1) : expect(3, 1);
   conn.destroy();
-  expect(2, 1); // client handle doesn't shut down until next tick
+  isIPv6 ? expect(2, 1) : expect(3, 1); // client handle doesn't shut down until next tick
   handles.push(conn);
 })();
 


mdawsonibm:
I discussed this with Cristian, the only question I'd have for other reviewers is if they are aware of if we do similar IPv6 checks elsewhere

misterdjules:
@mdawsonibm @CGavrila Yes, we do similar tests in other tests. For this one, we might also be able to use `common.hasIPv6()` and bind to `127.0.0.1` or `::1` depending on its output.

CGavrila:
@misterdjules: sure, that would work too, I just wanted to make it as less intrusive as possible. As long as the end goal of having the test pass for  both IPv4 and IPv6 is achieved, I'm fine with either option.

misterdjules:
@CGavrila Basically, I just want to avoid having tests for IPv6 at many different locations in this test. Binding to different addresses depending on `common.hasIPv6()` is one way to do that. We would need to add a comment that explains why we bind to a specific address too.

Thank you!

misterdjules:
@CGavrila Then let's use my suggestion of listening explicitly on `::1` if `common.hasIPv6` returns true, and on `127.0.0.1` otherwise. That will make the code simpler. Thank you!

mhdawson:
I was trying out the suggestion, looks to me like the doc is wrong.  It says this:

  server.listen(port[, hostname][, backlog][, callback])#

but the test failed if I put hostname after port. It passes the other way around.  Will look at that as a separate issue.

This is the patch I have in mind, now just need to track down the machine were we were failing,  reproduce and then validate this fixes the problem.

I have validated the test still passes when run on linux where the loopback has both IPv4 and IPv6

diff --git a/test/simple/test-process-active-wraps.js b/test/simple/test-process-active-wraps.js
index ae12986..c26e74e 100644
--- a/test/simple/test-process-active-wraps.js
+++ b/test/simple/test-process-active-wraps.js
@@ -24,6 +24,16 @@ var assert = require('assert');
 var spawn = require('child_process').spawn;
 var net = require('net');

+// Node first tries to use IPV6 when the host is not specified
+// and then falls back to IPV4 if that fails.  That causes
+// the test to fail if IPV6 is not configured as we end up with
+// an active handle for each of IPV6 and IPV4.  We will therefore
+// bind specifically to the IP4 or IPV6 loopback as appropriate
+var loopback = '::1';
+if(!common.hasIPv6) {
+  loopback = '127.0.0.1';
+}
+
 function expect(activeHandles, activeRequests) {
   assert.equal(process._getActiveHandles().length, activeHandles);
   assert.equal(process._getActiveRequests().length, activeRequests);
@@ -33,7 +43,7 @@ var handles = [];

 (function() {
   expect(0, 0);
-  var server = net.createServer().listen(common.PORT);
+  var server = net.createServer().listen(loopback, common.PORT);
   expect(1, 0);
   server.close();
   expect(1, 0); // server handle doesn't shut down until next tick
@@ -48,7 +58,7 @@ var handles = [];
   };

   expect(1, 0);
-  var conn = net.createConnection(common.PORT);
+  var conn = net.createConnection(loopback, common.PORT);
   conn.on('lookup', onlookup);
   conn.on('error', function() { assert(false); });
   expect(2, 1);

misterdjules:
@mhdawson The patch looks good, thank you for moving this forward :+1:

mhdawson:
Found the machine were it was failing and validated that it failed without the fix and passed with.

Since Cristian is away, created new pull request:

https://github.com/joyent/node/pull/25278

misterdjules:
Closing in favor of #25278.

PR #9450, opened 2015-03-20
Fix issue 9434

tyleranton:
Changed child_process.exec documentation to reflect that the parameters stdout and stderr are instances of Buffer only if {encoding: null}. Otherwise, they are instances of String with the encoding.


misterdjules:
@tyleranton Thank you!

This PR needs to be created against the v0.12, not master. Also, only one commit that represents the changes made to the documentation should be present.

Closing as it's not possible to change the target branch for a given PR, but please reopen one with your documentation change.

PR #13862, opened 2015-03-24
WIP: Verify core events and patterns

tjfontaine:
Before we can start stamping out what it means to be "compatible" with Node.js, we have to start by writing down how we think Node.js works today, and then verifying that scenario. (cc @dshaw @trevnorris @jasnell @mdawsonibm As we've all had conversations about API compatibility in the past)

Since we're event driven the key abstraction/data structure that impacts all of Node.js is `EventEmitter`. While we have spent the time to get that useful and fast, we haven't spent as much time making sure how Node.js presents its interfaces via the `EventEmitter` are meeting our expectations.

There are many places we can improve Node.js, but one of our friction points today (internally and externally) involves around what events will happen, when, and in what order.

This branch/PR adds the notion of a `StrictEE` implementation of `EventEmitter` that provides facilities to describe your events.

Instead of inheriting from plain old `EE` you inherit from `StrictEE`, and then when constructing you pass in an object which defines the events and the orders that can be passed in.

I myself am not convinced if this should be used in production, there is definitely plenty of low hanging fruit for improving its performance (it should create way fewer object literals). But really, I think (unless the user has asked for "strict mode") normal operating procedure should be that most of the strict checks are `noop`s.

Instead, I think this is critical while running our test suite, and verifying our software (and our downstream users code).

A definition of a rule looks like this currently:

``` javascript
{
  myEvent: {
    minCount: 1,
    maxCount: 2,
    after: ['event1', 'event2', 'event3'],
    notAfter: ['event4', 'event5'],
  },
}
```

and running our test suite you can grep the output and get something like this:

```
   2 #AssertionError: Event "close" must fire after at least one of ["error","end","finish","destroy","aborted"], fired []
   2 #AssertionError: Event "close" must fire after at least one of ["error","upgrade","aborted","finish","end"], fired ["socket","drain","response"]
   2 #AssertionError: Event "close" must fire after at least one of ["listening","error"], fired []
   2 #AssertionError: Event "listening" fired 2 times, more than 1
   2 #AssertionError: Event "unpipe" must **not** fire after "error", fired ["pipe","error"]
```

It's a very simple description, `after` is better defined: "after any", and `notAfter` is "not after any", and there is a mechanism for counting the times which an event has fired. We probably want the concept of `afterAll`.

In practice `minCount` is not used anywhere, and its overhead and potential to leak (in its current implementation) should discourage use of this without also growing some concept of a finalization step that could be used to verify this instead of the current global data structure.

You can also use `obj.strictEERegister` to extend rule definitions in child classes.

There probably should also be an `obj.strictEEReset` mode that resets the "state machine" for reuse (this already found an issue in `net.connect` re-usage).

This is currently based on v0.12 because I found it helpful while debugging other current bugs. In its current state it's a public API addition that wouldn't be suitable for v0.12.1, but I do think it's useful to finish this work on the current stable branch and rename APIs and any additional events such that we can continue to squash bugs.

Other obvious oversights include missing tests and documentation.

I'm publishing this PR as I'm looking for help from someone in the community who wants to come and help complete this. As a result, you'll end up with a lot of more knowledge about how Node.js actually works. I myself learned quite a bit bootstrapping into this. For everyone else, we'll have described and along the way documented quite a bit of how Node.js works, which will make it easier for the next person to come and contribute a fix.

### DIFF
--- a/lib/_http_agent.js
+++ b/lib/_http_agent.js
@@ -63,6 +63,8 @@ function Agent(options) {
     var name = self.getName(options);
     debug('agent.on(free)', name);
 
+    // This seems an absurd check, we should not be firing this event after
+    // it has been closed/destroyed
     if (!socket.destroyed &&
         self.requests[name] && self.requests[name].length) {
       self.requests[name].shift().onSocket(socket);
--- a/lib/_http_client.js
+++ b/lib/_http_client.js
@@ -42,6 +42,29 @@ var Agent = require('_http_agent');
 function ClientRequest(options, cb) {
   var self = this;
   OutgoingMessage.call(self);
+  this.strictEERegister({
+    connect: {
+      maxCount: 1,
+      after: ['socket'],
+    },
+    close: {
+      maxCount: 1,
+      after: ['error', 'upgrade', 'aborted', 'finish', 'end'],
+    },
+    error: {
+      maxCount: 1,
+      notAfter: ['close'],
+    },
+    continue: {
+      maxCount: 1,
+      notAfter: ['error', 'close'],
+      after: ['socket'],
+    },
+    response: {
+      notAfter: ['error', 'close'],
+      maxCount: 1,
+    },
+  }, true);
 
   if (util.isString(options)) {
     options = url.parse(options);
--- a/lib/_http_incoming.js
+++ b/lib/_http_incoming.js
@@ -38,6 +38,15 @@ exports.readStop = readStop;
 /* Abstract base class for ServerRequest and ClientResponse. */
 function IncomingMessage(socket) {
   Stream.Readable.call(this);
+  this.strictEERegister({
+    aborted: {
+      maxCount: 1,
+      notAfter: ['error', 'close', 'end'],
+    },
+    timeout: {
+      notAfter: ['error', 'close'],
+    },
+  });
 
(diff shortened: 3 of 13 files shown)


jasnell:
I recently sat down with Bert and Michael Dawson and had a conversation
around improvements around the internal APIs. That conversation and this
one pair nicely. I'll help where I can. Either later on this evening or
tomorrow I'll dig in a bit deeper in response to your write up here.

tjfontaine:
Related to this, but not included (and not necessarily in scope of this change) is the idea of separating out what is an internal state machine and what are external events for consumers to subscribe.

While working on getting the PR this far, it's clear that at times we go to great lengths to avoid the scenario where external consumers of add/remove listeners won't adversely effect our internal state machine.

It shouldn't be that during the normal consumption of the `EE` APIs (i.e. `removeAllListeners`) results in node not being able to make forward progress.

It also shouldn't be the case that Node.js is worried about adding internal events (especially during patch releases) that might step on external users.

We should be working to clearly delineate between internal and external, and this sort of change gets us closer to even knowing what our surface area actually is.

jasnell:
@trevnorris ... any reason you can see to keep this open?

trevnorris:
Nope.

PR #14002, opened 2015-03-25
correct example code

AlexKVal:


misterdjules:
Thanks @AlexKVal for your contribution! The sample code describes the normal usage of `stream.Writable.end()`, so I don't think we want it to trigger an error.

However I agree that it's slightly confusing since the example is right after the sentence `Calling write() after calling end() will raise an error.`.

Maybe we could have the normal use case sample code before that sentence, and have a specific code sample for the write-after-end error after that sentence?

What do you think?

AlexKVal:
Agree. I had written it without thinking some steps ahead.

Something like this ?

>The right way:
`right example code`

>The wrong way:
`wrong example code`

misterdjules:
@AlexKVal Right, something like:
```
Call this method when no more data will be written to the stream. If supplied, the callback is attached as a listener on the finish event.

// write 'hello, ' and then end with 'world!'
var file = fs.createWriteStream('example.txt');
file.write('hello, ');
file.end('world!');

Calling write() after calling end() will raise an error:

// end with 'world!' and then write with 'hello, ' will raise an error
var file = fs.createWriteStream('example.txt');
file.end('world!');
file.write('hello, ');
```

AlexKVal:
I'm closing this PR because of #15517.

PR #14086, opened 2015-03-25
configure: added --with-arm-fpu option

Ferk:
The configure script is currently using arm-fpu='vfpv3' as the hardcoded default.

This patch will add a --with-arm-fpu option that will allow the choice of other FPU hardware. Otherwise we would get an "Illegal instruction" error on runtime when using the v8 engine in ARM hardware that has a different floating point unit (I was having this problem on an at91sama5 microcontroller).

It might be advisable to use as default value "default" instead of "vfpv3". For consistency and to avoid having to update the default value in the future if v8 ever decides to drop support of vfpv3 like it did with vfpv2. 

But since this would change the current default behavior ("default" currently falls back to "vfpv3-d16" in the version of v8 you are using), I would like to know your opinion about it, perhaps you have good reasons to have "vfpv3" as the default arm_fpu, instead of "vfpv3-d16".

### DIFF
--- a/configure
+++ b/configure
@@ -232,6 +232,12 @@ parser.add_option('--with-arm-float-abi',
     help='specifies which floating-point ABI to use. Valid values are: '
          'soft, softfp, hard')
 
+parser.add_option('--with-arm-fpu',
+    action='store',
+    dest='arm_fpu',
+    help='specifies which vector floating point unit to use. Valid values are: '
+         'vfpv3, vfpv3-d16, neon')
+
 parser.add_option('--with-dtrace',
     action='store_true',
     dest='with_dtrace',
@@ -490,6 +496,11 @@ def configure_arm(o):
   else:
     arm_float_abi = 'default'
 
+  if options.arm_fpu:
+    arm_fpu = options.arm_fpu
+  else:
+    arm_fpu = 'vfpv3'
+
   if is_arch_armv7():
     o['variables']['arm_version'] = '7'
   elif is_arch_armv6():
@@ -497,7 +508,7 @@ def configure_arm(o):
   else:
     o['variables']['arm_version'] = 'default'
 
-  o['variables']['arm_fpu'] = 'vfpv3'  # V8 3.18 no longer supports VFP2.
+  o['variables']['arm_fpu'] = arm_fpu
   o['variables']['arm_neon'] = int(is_arm_neon())
   o['variables']['arm_thumb'] = 0      # -marm
   o['variables']['arm_float_abi'] = arm_float_abi


misterdjules:
Just FYI, `vfpv3` will not the default anymore when github.com/joyent/node/pull/14193 lands.
I'm not familiar with ARM, but why aren't vfpv1, vfpv2 and vfpv4 in the list of available floating point units?

Ferk:
When that pull requests lands, vfpv3 will still be the default for armv7 systems. It won't fix the compilation issues on hardware using vfpv3-d16.

I didn't add vfpv2 because it was said to be "unsupported" in a comment. I also didn't want to add too many untested options, but I could add all the other possibilities, not just vfpvx:  ‘vfp’, ‘vfpv3’, ‘vfpv3-fp16’, ‘vfpv3-d16’, ‘vfpv3-d16-fp16’, ‘vfpv3xd’, ‘vfpv3xd-fp16’, ‘neon’, ‘neon-fp16’, ‘vfpv4’, ‘vfpv4-d16’, ‘fpv4-sp-d16’, ‘neon-vfpv4’, ‘fpv5-d16’, ‘fpv5-sp-d16’, ‘fp-armv8’, ‘neon-fp-armv8’, and ‘crypto-neon-fp-armv8’.

misterdjules:
> When [that pull request](http://github.com/joyent/node/pull/14193) lands, vfpv3 will still be the default
> for armv7 systems. It won't fix the compilation issues on hardware using vfpv3-d16 (node doesn't
> compile on my at91sama5 microcontroller using the vfpv3 option).

Right, just to clarify the intent of my comment was to give an update of what's coming soon, not to say that all issues for building Node.js on ARM are solved.

Thanks for the clarifications!

jasnell:
@misterdjules ... am I right to assume that this would not land here? This PR likely needs to be revisited.

jasnell:
At this point, anything done here would need to happen in nodejs/node. Closing this here.

not-a-user:
Thank you! Why is this not merged yet? This patch is required to build node.js v0.12.7 on my nvidia tegra 2 (cortex-a9, vfpv3-d16, no neon, ubuntu 12.04, gcc 4.6, linux 3.4). My `/proc/cpuinfo`:

```
Processor       : ARMv7 Processor rev 0 (v7l)
processor       : 0
BogoMIPS        : 1987.37

processor       : 1
BogoMIPS        : 1993.93

Features        : swp half thumb fastmult vfp edsp thumbee vfpv3 vfpv3d16 tls 
CPU implementer : 0x41
CPU architecture: 7
CPU variant     : 0x1
CPU part        : 0xc09
CPU revision    : 0

Hardware        : trimslice
Revision        : 0000
Serial          : 0000000000000000
```

PR #14088, opened 2015-03-25
doc: fixed child_process.exec doc

### LINKED ISSUE #9434: child_process.exec doc is inconsistent with actual implementation
[child_process.exec' documentation](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback) states that the `stdout` and `stderr` parameters are instances of `Buffer` by default, but it's actually not the case. They are instances of `Buffer` only if `{ encoding: null}` is passed as options.

The documentation of v0.12.x should be fixed to reflect this behavior.

The original intent was that `stdout` and `stderr` parameters would be instances of `Buffer`. If that still holds true, then the change in behavior should be done in master post v0.12.x.

Adding to milestone 0.12.2 as it's a quick change.

tyleranton:
Changed child_process.exec documentation to reflect that the parameters stdout and stderr are instances of Buffer only if {encoding: null}. Otherwise, they are instances of String with the encoding.

Fixes #9434

### DIFF
--- a/doc/api/child_process.markdown
+++ b/doc/api/child_process.markdown
@@ -460,11 +460,11 @@ index corresponds to a fd in the child.  The value is one of the following:
    ignored node will open `/dev/null` and attach it to the child's fd.
 4. `Stream` object - Share a readable or writable stream that refers to a tty,
    file, socket, or a pipe with the child process. The stream's underlying
-   file descriptor is duplicated in the child process to the fd that 
+   file descriptor is duplicated in the child process to the fd that
    corresponds to the index in the `stdio` array. Note that the stream must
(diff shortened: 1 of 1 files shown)


KenanY:
@tyleranton Can you remove the `.DS_Store` file?

tyleranton:
Yea, sorry about that.

jasnell:
LGTM.. @misterdjules, can you take a look as well

jasnell:
@piscisaureus ... look good to you?

piscisaureus:
lgtm

jasnell:
Landed in 2860c53

PR #14089, opened 2015-03-25
fix install source path for openssl headers

obastemur:
Seems node 0.12.x and node 0.10.x / jxcore use the same install script. This must be broken for a very long time. Moving the commit from https://github.com/jxcore/jxcore/commit/fc1023f48d046f5731cb5899581b9353db5c7607

### DIFF
--- a/tools/install.py
+++ b/tools/install.py
@@ -161,7 +161,7 @@ def files(action):
 
   if 'false' == variables.get('node_shared_openssl'):
     action(['deps/openssl/config/opensslconf.h'], 'include/node/openssl/')
-    subdir_files('deps/openssl/include/openssl', 'include/node/openssl/', action)
+    subdir_files('deps/openssl/openssl/include/openssl', 'include/node/openssl/', action)
 
   if 'false' == variables.get('node_shared_v8'):
     subdir_files('deps/v8/include', 'include/node/', action)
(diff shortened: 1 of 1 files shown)


obastemur:
ping! this issue is breaking the native addons using builtin openssl.

jasnell:
LGTM

trevnorris:
LGTM

misterdjules:
@obastemur Thank you!

LGTM, however it also applies to v0.10 and v0.12, so we may want to close this PR and submit another one against v0.10 that will get merged eventually into v0.12 and master.

Also, I'm not familiar with writing binary add-ons that require OpenSSL, but was this issue not raised before because it "only" impacts developers who use Node.js' OpenSSL headers without using node-gyp?

obastemur:
`Also, I'm not familiar with writing binary add-ons that require OpenSSL, but was this issue not raised before because it "only" impacts developers who use Node.js' OpenSSL headers without using node-gyp?`

Yes. IMHO this part should be removed entirely. openssl and the headers from other dependencies should be copied into include/deps folder (similar to node-gyp include packages). I have this item on my todo list. Once I have something proper for node 0.10, 0.12 I will be bringing it here at least for a discussion.

misterdjules:
@obastemur Great thank you!

misterdjules:
Landed in 9d19dfbfdb2253e0bf42b55c7c1bbd618f40bf3c and 4028669531bcd30d6082bf0dfdb8d73ddb0717c6. These will get merged in v0.12 eventually (sooner than later). Thank you again @obastemur!

PR #14146, opened 2015-03-26
URL: Fix resolving from non-file to file URLs.

jagoda:
When resolving a reference URL with the 'file' scheme an no host
against a base URL without the 'file' scheme, the first path element
of the reference URL is used as the host for the target URL. This
results in an invalid target URL.

This change makes an exception for file URLs so that the host is not
mangled during URL resolution.

### DIFF
--- a/lib/url.js
+++ b/lib/url.js
@@ -510,7 +510,9 @@ Url.prototype.resolveObject = function(relative) {
     }
 
     result.protocol = relative.protocol;
-    if (!relative.host && !hostlessProtocol[relative.protocol]) {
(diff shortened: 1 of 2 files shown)


jagoda:
I believe this issue is also present in v0.10.

jasnell:
@jagoda ... I know it's been a while, but can you please update this to match what was landed in https://github.com/nodejs/node/pull/1277. I'd like to go ahead and land this in v0.12 but need the same style fixes made.

jagoda:
@jasnell I think this should be good to go now.

PR #14182, opened 2015-03-27
l10n: Initial l10n support

jasnell:
The initial implementation of the node localization support.

If --with-intl is enabled, this will make use of ICU's resource
bundle mechanism. If --with-intl is disabled, a printf based
fallback will be used. Within the code, everything is macro
driven.

See the README for more detail.
A test case is provided.

Build tested on Mac, Ubuntu 14.10 and Win 7. Tests pass.

This does not yet include the externalized strings and no part
of node has yet been modified to use this mechanism. This is
just the initial implementation support. Step two will be to begin
externalizing the strings both within the c/c++ native code and
the JS core library.


jasnell:
cc @joyent/node-coreteam

jasnell:
Will be revisiting this in nodejs/node

PR #14185, opened 2015-03-28
test: run v8 tests

jasnell:
2nd pass with the v8 tests. moves the new targets
into a separate gyp as requested. uses an alternate
build for cutest that excludes the one file that requires
c++11 (there's always that _one_ test that uses auto
and ruins it for everyone else).

(replaces https://github.com/joyent/node/pull/9208)

To use:

```
./configure --with-v8-tests
make
make test-v8
```

```
./configure --with-intl=full-icu --with-v8-tests
make
make test-v8-intl
```

A rudimentary start at support on win32 is included
but it likely needs more work.

### DIFF
--- a/vcbuild.bat
+++ b/vcbuild.bat
@@ -38,6 +38,7 @@ set noperfctr_msi_arg=
 set i18n_arg=
 set download_arg=
 set build_release=
+set v8_tests=
 
 :next-arg
 if "%1"=="" goto args-done
@@ -62,6 +63,8 @@ if /i "%1"=="test-message"  set test=test-message&goto arg-ok
 if /i "%1"=="test-gc"       set test=test-gc&set buildnodeweak=1&goto arg-ok
 if /i "%1"=="test-all"      set test=test-all&set buildnodeweak=1&goto arg-ok
 if /i "%1"=="test"          set test=test&goto arg-ok
+if /i "%1"=="test-v8"       set test=test-v8&goto arg-ok
+if /i "%1"=="test-v8-intl"  set test=test-v8-intl&goto arg-ok
 @rem Include small-icu support with MSI installer
 if /i "%1"=="msi"           set msi=1&set licensertf=1&set download_arg="--download=all"&set i18n_arg=small-icu&goto arg-ok
 if /i "%1"=="upload"        set upload=1&goto arg-ok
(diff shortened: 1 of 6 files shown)


jasnell:
This version still does the download for gtest and gmock. @piscisaureus suggested that we just add those in to the tree if we're going to pull the v8 tests in.

jasnell:
@joyent/node-coreteam

indutny:
A question... why do you want to do this instead of always using stable v8 version?

jasnell:
This is largely intended as a sanity check. Since we're building v8 and may end up floating patches on it, adding the additional check of running the v8 tests is useful.

indutny:
@jasnell ok...

indutny:
I mean, makes sense. But kind of sad :) I'd rather remove the tests to shave off some mbs from the tarball.

jasnell:
I don't disagree :-)

orangemocha:
This looks useful. One slight concern I have is whether it's the best approach to add the logic to node's build scripts themselves (makefile, vcbuild.bat), or whether it would be better to isolate this stuff in their own scripts under tools/v8. After all, you are not testing node at all.. what you want is to test a dependency, and this would probably relevant only in the context of updating the dependency.

jasnell:
At the moment, adding this to node's build scripts is the easiest thing but it can be isolated. The v8 tests require a few additional pieces to be built (specifically, d8) and it's just easier to do that while we're building everything else. That said, there's no harm in isolating it.

orangemocha:
@jasnell , you are right, it's probably easier to roll it out this way. Fine by me.

vcbuild.bat
```diff
@@ -208,6 +212,16 @@ if "%test%"=="test-simple" set test_args=%test_args% simple
 if "%test%"=="test-message" set test_args=%test_args% message
 if "%test%"=="test-gc" set test_args=%test_args% gc
 if "%test%"=="test-all" set test_args=%test_args%
+if "%test%"=="test-v8" goto test-v8
+if "%test%"=="test-v8-intl" goto test-v8-intl
+
```
orangemocha: Note that the control here was falling through to :build-node-weak before your change. Now, it will fall through to :test-v8, causing v8 tests to **always** run.

You can either add a goto :build-node-weak after these checks or move :test-v8 and :-test-v8-intl somewhere below where they are not in the default control flow path, essentially making them subroutines.

jasnell: Good catch.

jasnell:
Will be revisiting this in nodejs/node

PR #14352, opened 2015-03-31
correct text

AlexKVal:
its -> it's -> it is
a HTTP -> we read 'H' as 'eich' -> an HTTP

### DIFF
--- a/doc/api/cluster.markdown
+++ b/doc/api/cluster.markdown
@@ -24,7 +24,7 @@ all share server ports.
       });
     } else {
       // Workers can share any TCP connection
-      // In this case its a HTTP server
+      // In this case it is an HTTP server
       http.createServer(function(req, res) {
         res.writeHead(200);
         res.end("hello world\n");


doc/api/cluster.markdown
```diff
@@ -24,7 +24,7 @@ all share server ports.
       });
     } else {
       // Workers can share any TCP connection
-      // In this case its a HTTP server
+      // In this case it is an HTTP server
```
misterdjules: @joyent/node-coreteam Could a native English speaker please confirm the usage of "an" instead of "a" here? Thank you!

tresf: Confirmed per BBC:

> Take the eighth letter of the alphabet, pronounce it haitch and then look for the slightly agonised look in some people's eyes. One suggestion is that it touches on a long anxiety in English over the letter aitch. In the 19th Century, it was normal to pronounce hospital, hotel and herb without the h.Oct 28, 2010

So "an HTTP" should be similar to "an hour". :)

jasnell:
Will land this in a separate commit and PR in http://github.com/nodejs/node master

jasnell:
Closing here. New PR opened.

PR #14383, opened 2015-04-01
tls: Update cipher-list defaults, add command-line switches

jasnell:
1. Pull over the io.js default cipher list updates
2. Adds new `--cipher-list` and `--enable-legacy-cipher-list` command line switches. 
   `--cipher-list` allows a new default cipher list to be specified  
   `--enable-legacy-cipher-list` tells node to use specific older cipher lists
3. Adds new `NODE_CIPHER_LIST` and `NODE_LEGACY_CIPHER_LIST`
   environment variables as alternatives to the command line switches
4. Adds a new `getLegacyCiphers` method to the TLS module
5. A test case is included.


jasnell:
(This is just for master... separate PR's will be coming for backporting changes into
v.0.12.x and v0.10.x)

/cc @joyent/node-coreteam

jasnell:
There will be just a couple more updates on this coming soon.. just a few minor late breaking tweaks...

src/node.cc
```diff
@@ -2929,6 +2929,12 @@ static void PrintHelp() {
 #endif
          "  --enable-ssl2        enable ssl2\n"
          "  --enable-ssl3        enable ssl3\n"
+         "  --cipher-list=val    specify the default TLS cipher list\n"
+         "  --enable-legacy-cipher-list=val \n"
+         "                       set to v0.10.38 to use the v0.10.38 list,\n"
+         "                       set to v0.10.39 to use the v0.10.39 list.\n"
+         "                       set to v0.12.2 to use the v0.12.2 list.\n"
+         "                       set to v0.12.3 to use the v0.12.3 list.\n"
```
trevnorris: I think it would be enough to simply list possible options. users could probably figure out from context that each one corresponds to a given version.

src/node.h
```diff
@@ -223,6 +223,18 @@ NODE_EXTERN void RunAtExit(Environment* env);
   }                                                                           \
   while (0)
 
+#define NODE_DEFINE_STRING_CONSTANT(target, constant)                         \
```
trevnorris: instead I would recommend that the `Isolate` be the first passed argument.

src/node_crypto.h
```diff
@@ -59,10 +60,51 @@
...
+                                "!RC4:!MD5:!aNULL"
+
+#define DEFAULT_CIPHER_LIST_HEAD "ECDHE-RSA-AES256-SHA384:"                   \
+                                 "DHE-RSA-AES256-SHA384:"                     \
+                                 "ECDHE-RSA-AES256-SHA256:"                   \
+                                 "DHE-RSA-AES256-SHA256:"                     \
+                                 "ECDHE-RSA-AES128-SHA256:"                   \
+                                 "DHE-RSA-AES128-SHA256:"                     \
+                                 "HIGH:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!MD5:" \
+                                 "!PSK:!SRP:!CAMELLIA"
```
trevnorris: alignment issues on the ending `/`. Check that all the `DEFAULT_CIPHER_LIST_*` are aligned at the end the same.

trevnorris:
Few small things, but looking good.

src/node.cc
jasnell: ;)  one hopes, yes.

mdawsonibm:
Reviewed for consistency with 0.12 and 0.10 patches looks good.  Differences between them also make sense.  This looks like changes from original io.js patch plus the changes for 0.10.X and 0.12.X.  Will need to squash and ensure final commit comment is less than 50 char limit but otherwise lgtm.

src/node.cc
```diff
@@ -3040,6 +3047,20 @@ static void ParseArgs(int* argc,
     } else if (strcmp(arg, "--v8-options") == 0) {
       new_v8_argv[new_v8_argc] = "--help";
       new_v8_argc += 1;
+    } else if (strncmp(arg, "--cipher-list=", 14) == 0) {
```
misterdjules: Could this be done in `src/node.js` instead of in `src/node.cc`? It seems it would just be less error prone and shorter.

test/simple/test-tls-cipher-list.js
```diff
@@ -0,0 +1,69 @@
...
+function doTest(checklist, env, useswitch) {
+  var options;
+  if (env && useswitch === 1) {
+    options = {env:env};
+  }
+  var args = ['-e', 'console.log(require(\'tls\').DEFAULT_CIPHERS)'];
+
+  switch(useswitch) {
+    case 1:
+      // Test --cipher-test
```
misterdjules: Very minor, but I guess this should read `// Test --cipher-list`.

test/simple/test-tls-cipher-list.js
```diff
@@ -0,0 +1,69 @@
...
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+var spawn = require('child_process').spawn;
+var assert = require('assert');
+var tls = require('tls');
+
+function doTest(checklist, env, useswitch) {
+  var options;
+  if (env && useswitch === 1) {
```
misterdjules: It seems that if `usesswitch === 1`, it means that we want to test `--cipher-list`, so why set `options` in this case?

src/node_crypto.cc
```diff
@@ -4851,6 +4852,26 @@ static void array_push_back(const TypeName* md,
...
+void DefaultCiphers(const v8::FunctionCallbackInfo<v8::Value>& args) {
+  Environment* env = Environment::GetCurrent(args.GetIsolate());
+  HandleScope scope(env->isolate());
+  v8::String::Utf8Value key(args[0]);
+  const char * list = legacy_cipher_list(ToCString(key));
+  if (list != NULL) {
+    args.GetReturnValue().Set(
+      v8::String::NewFromUtf8(args.GetIsolate(), list));
+  } else {
+    args.GetReturnValue().Set(
```
misterdjules: Should a warning be displayed if a given legacy ciphers identifier doesn't exist?

test/simple/test-tls-dhe.js
```diff
@@ -47,6 +47,7 @@ function test(keylen, expectedCipher, cb) {
   var options = {
     key: key,
     cert: cert,
+    ciphers: ciphers,
```
misterdjules: Is it actually needed? I thought that `'DHE-RSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256'` were in the default ciphers list and that this test tested that the server should work with these ciphers out of the box, without using the `ciphers` option.

misterdjules:
@jasnell Added some comments, sorry again for taking so long in reviewing this!

misterdjules:
@jasnell Running the tests suite in `test/external/ssl-options/test.js` with these changes now, I'll keep you posted about the results.

misterdjules:
@jasnell All tests in `test/external/ssl-options/test.js` pass.

jasnell:
+1... my time tonight is disappearing quick so I will work on landing these
in the morning. Thank you very much for the additional review Julien. It's
very helpful

jasnell:
Landed

PR #14412, opened 2015-04-02
src: support for release candidates

misterdjules:
This is a pull-request to get the ball rolling on doing release candidates for _stable_ versions of node. Development releases would not go through this process.

These release candidates would be made available when all the fixes for a given version have landed in the stable branch. Users of node would have one week to report feedback on a given release candidate. If no regression is found in a given release candidate after this period of one week, then the release process of the actual release can start.

The goal is to avoid releasing stable versions which have severe bugs that affect a lot of users. In the last few months, we have had releases which broke a lot of users (and I was responsible for most of the changes that broke them). When that happens, it puts more pressure on the release management team, and overall puts both maintainers and users in a difficult position.

The current state of this pull request allows to release a release candidate version by [adding the definition of the `NODE_TAG` macro and setting the `NODE_VERSION_IS_RELEASE` macro to `1`  in `src/node_version.h`](https://github.com/misterdjules/node/commit/3bfc35bf70aad62688f4078ad0fabada02690e49).

The `NODE_TAG` macro is not a new macro, and thus to my knowledge these changes do not introduce any new API for users who depend on the macros defined in `src/node_version.h`.

With this change committed, we can trigger the [`nodejs-release` job on Jenkins](http://jenkins.nodejs.org/job/nodejs-release/) and then run `node-release-post-build.sh` and, using the next stable release as an example, have [a new release named `v0.12.3-rc1` available at nodejs.org](https://nodejs.org/dist/v0.12.3-rc1/). Using the same example, a [v0.12.3-rc1 tag](https://github.com/misterdjules/node/releases/tag/v0.12.3-rc1) is also created.

The change in version name/number has a subtle impact on important parts of our release process, and I'm currently testing various scenarios such as using nvm to install release candidates or upgrading from a given release candidate to the actual corresponding release on Windows and making sure all these scenarios work as expected. I'm also currently making sure that the release process is not made more difficult or error prone by the addition of this change.

In the meantime, any feedback would be very much appreciated!

/cc @joyent/node-coreteam. I would also be very interested in getting feedback from @wraithan who has helped us a lot for many releases in the past, @hunterloftis and @ploer who also helped us a lot when we had some issues with previous v0.10.x releases.


misterdjules:
Adding to the 0.12.3 milestone as I think it would be great to start that process with the v0.12.3 release.

misterdjules:
Created [a PR in creationix/nvm](https://github.com/creationix/nvm/pull/712) to support the release candidate process described in this PR.

ljharb:
The creation of the release candidate has broken `nvm` - see https://github.com/creationix/nvm/issues/711. It's really unfortunate that this decision was made without even checking with version managers first.

misterdjules:
@lljharb This decision has _not_ been made yet. The purpose of this PR _is_ to check with version managers first, like the original comment says:
> This is a pull-request to get the ball rolling

The reason why the v0.12.3-rc1 is still available on nodejs.org is a mistake on my side, it should not have stayed there and was used for testing purposes. I'm removing it now, my apologies for the confusion.

Your feedback is needed and appreciated, and this is also why I created https://github.com/creationix/nvm/pull/712.

Thank you.

ljharb:
@misterdjules thanks, glad to hear it :-)

trevnorris:
+1 for having rc's.

jasnell:
I'm +1 on this also. Looking forward tho, I'm needing to get my head around
how this would work in the converged foundation project plan.

hunterloftis:
@misterdjules thanks for considering the downstream impact! I'm not sure that we will use rcs much ourselves; right now I'm focusing on better messaging so users understand what's happening when latest stable gets a bump. It's unlikely that our automated tests will catch issues that yours don't, I think, but our thousands of users spot them immediately ;)

ljharb:
Since this is asking for feedback, here are my thoughts as "release candidates" pertain to `nvm`:
 - One option is to keep `nodejs.org/dist` solely for actual releases, and use `nodejs.org/dist-rc` or something for RCs. Due to the way I've implemented `iojs`, it wouldn't be very complex to add a `noderc` prefix such that `nvm install noderc-0.13` would do what you expected. However, this only works seamlessly if there's just one RC for a given version number, which clearly won't be the case.
 - Another option is to use a suffix on the version number, ie `nvm install 0.12-rc1`. However, this imo will greatly complicate `nvm`, because all of the code is written to assume (now with an optional prefix) a normal semver version number. https://github.com/creationix/nvm/pull/712 is a solid first start but I'm pretty sure that's just the tip of the iceberg.
 - Another alternative is to not use nodejs.org at all for release candidates - users that are testing those can install from source by checking out the tag on the git repo. What are the reasons that wouldn't be sufficient?

Whatever format or distribution channel is selected, my primary concern is that the existence of `rc` builds does not in any way break `nvm` (or `nave` or `n`, for that matter).

tools/getnodeversion.py
```diff
@@ -13,4 +19,41 @@
...
+  version_string = '%(major)s.%(minor)s.%(patch)s'% locals()
+  if tag:
+    if msi:
+      rc_number_match = re.match('.*(\d+)$', tag)
+      if rc_number_match and rc_number_match.group(1):
+        rc_number = rc_number_match.group(1)
+        version_string += '.%(rc_number)s'% locals()
+      else:
+        sys.stderr.write('Could not determine proper rc number for msi ' +
+                         'version, exiting')
```
orangemocha: Wondering if this will work for tagged releases that are not RCs. My python regexp is a bit rusty, but it seems that out of -for example- 0.12.1, the regexp would match 1 as rc_number_match, and add it again to the version string. Or if it cannot match an rc number, the else path will fail.

wraithan: At a glance this looks great. I didn't see this until now and an wiped out
from work. I'll look at this more in depth tomorrow.

jasnell:
@misterdjules ... at this point is there reason to keep this open? Is this still something we want to pursue?

ljharb:
I believe that `io.js` already has a release candidate process, that will become the node process - and the work I've begun on `nvm` to support `io.js` release candidates will easily be usable with post-merger `node` release candidates.

misterdjules:
Closing as this was done by @rvagg as part of the nodejs/node release process.

PR #14413, opened 2015-04-02
tls: disable RC4, add --cipher-list command line switch

jasnell:
Disable RC4 in the default cipher list

Add the `--cipher-list` command line switch and `NODE_CIPHER_LIST`
environment variable to completely override the default cipher list.

Add the `--enable-legacy-cipher-list` and `NODE_LEGACY_CIPHER_LIST`
environment variable to selectively enable the default cipher list from
previous node.js releases.

(Targets v0.10.38-release)

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -2567,8 +2567,7 @@ static void PrintHelp() {
          "  --enable-ssl2        enable ssl2\n"
          "  --enable-ssl3        enable ssl3\n"
          "  --cipher-list=val    specify the default TLS cipher list\n"
-         "  --enable-legacy-cipher-list=val \n"
-         "                       set to v0.10.38 to use the v0.10.38 list\n"
+         "  --enable-legacy-cipher-list=v0.10.38 \n"
          "\n"
          "Environment variables:\n"
 #ifdef _WIN32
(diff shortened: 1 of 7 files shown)


mdawsonibm:
I had reviewed the 0.10.X version of the patch with James and we have run it through our internal builds.  FV passed without any regressions, we are starting SVT now. 

lgtm

jasnell:
Ok. So far the only nit appears to be making the PrintHelp output less
verbose. I can make that change then will work with Julien on getting this
landed.

src/node.cc
```diff
@@ -2652,6 +2657,21 @@ static void ParseArgs(int argc, char **argv) {
     } else if (strcmp(arg, "--throw-deprecation") == 0) {
       argv[i] = const_cast<char*>("");
       throw_deprecation = true;
+    } else if (strncmp(arg, "--cipher-list=", 14) == 0) {
```
misterdjules: I'm wondering if we _need_ to support a `--cipher-list` command line option in v0.10.x. Adding support for `--enable-legacy-cipher-list` already breaks the contract of not adding additional API in patch releases, and it seems sufficient to handle the use case of users not wanting to use the new default ciphers list.

misterdjules:
@jasnell Added one additional comment specific to v0.10.x and this PR, and my comments in #14383 also apply here.

Thank you for doing all this work @jasnell :+1:

src/node.cc
jasnell: Adding --cipher-list here is intended to make it so we don't have to make this kind of change in v0.10.x and v0.12.x again. If another one of the ciphers comes into question, we can tell users to simply pass in the new cipher list rather than going in and changing it.

jasnell:
Landed in #67d9a56 and #02a549e

PR #14414, opened 2015-04-02
tls: disable RC4, add --cipher-list command line switch

jasnell:
Disable RC4 in the default cipher list

Add the `--cipher-list` command line switch and `NODE_CIPHER_LIST`
environment variable to completely override the default cipher list.

Add the `--enable-legacy-cipher-list` and `NODE_LEGACY_CIPHER_LIST`
environment variable to selectively enable the default cipher list from
previous node.js releases.

(Targets v0.12.2-release)

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -2936,6 +2936,11 @@ static void PrintHelp() {
 #endif
          "  --enable-ssl2        enable ssl2\n"
          "  --enable-ssl3        enable ssl3\n"
+         "  --cipher-list=val    specify the default TLS cipher list\n"
+         "  --enable-legacy-cipher-list=val \n"
+         "                       set to v0.10.38 to use the v0.10.38 list,\n"
+         "                       set to v0.10.39 to use the v0.10.39 list.\n"
+         "                       set to v0.12.2 to use the v0.12.2 list.\n"
          "\n"
          "Environment variables:\n"
 #ifdef _WIN32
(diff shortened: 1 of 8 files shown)


Fishrock123:
Why not just use one PR? You can keep pushing to it without creating extra noise..

jasnell:
There are three separate PR's for three separate branches (v0.10.38,
v0.12.2 and Master).

Fishrock123:
Ah, makes sense, but then again, isn't it usually just merged forwards from the latest (0.10. in this case)?

jasnell:
Typically yes. In this case there are key differences in the patches that
make that a bit difficult to review. Separating them out like this allows
the changes to be reviewed in isolation. If the changes look good, I'll
work with Julian to land the changes in whatever way makes the most sense.

mdawsonibm:
Looks consistent with 0.10.38 and where different makes sense.  Will need to squash the two commits and shorten the title to fit under the 50 char limit.  Otherwise lgtm.

src/node.cc
```diff
@@ -3047,6 +3054,20 @@ static void ParseArgs(int* argc,
     } else if (strcmp(arg, "--v8-options") == 0) {
       new_v8_argv[new_v8_argc] = "--help";
       new_v8_argc += 1;
+    } else if (strncmp(arg, "--cipher-list=", 14) == 0) {
```
misterdjules: Paraphrasing my [previous comment](https://github.com/joyent/node/pull/14413#discussion_r27933665) for the corresponding PR targeted to v0.10.x:

I'm wondering if we _need_ to support a `--cipher-list` command line option in v0.12.x. Adding support for `--enable-legacy-cipher-list` already breaks the contract of not adding additional API in patch releases, and it seems sufficient to handle the use case of users not wanting to use the new default ciphers list.

misterdjules:
@jasnell Same comments as in https://github.com/joyent/node/pull/14383, plus https://github.com/joyent/node/pull/14414#discussion_r27934008.

Thank you very much again for doing this work :+1:

src/node.cc
jasnell: The intent on introducing --cipher-list in v0.10.x and v0.12.x is to avoid
the possibility of having to make these kinds of changes in the future. If,
for instance, another one of the default ciphers comes under scrutiny and
we have to switch it off, rather than having to bump the maintenance
release again we can simply tell folks to use the command line switch or
envar to override the list entirely and go from there.

jasnell:
Landed in #f9291a9 and #b5737bb

Issue #7766, opened 2014-06-10
bad error message "TypeError: Usage: Buffer.concat(list, [length])"


tj:
"array expected, passed \<type\>" or something would be more obvious to assert that it's a type issue, something consistent throughout core for type issues would be nice. AFAIK no others use "Usage: ..."

misterdjules:
Fixed in d01a900 for v0.10, and merged in v0.12 and master with b0425f0 and a995a6a. Thank you @chrisdickinson!

PR #14421, opened 2015-04-03
deps: upgrade npm to 2.7.5

othiym23:
This release of npm is a little more important than most, because it contains two security fixes:
- [`300834e`](https://github.com/npm/npm/commit/300834e91a4e2a95fb7fb59c309e7c3fc91d2312)
  `tar@2.0.0`: Normalize symbolic links that point to targets outside the
  extraction root. This prevents packages containing symbolic links from
  overwriting targets outside the expected paths for a package. Thanks to [Tim
  Cuthbertson](http://gfxmonk.net/) and the team at [Lift
  Security](https://liftsecurity.io/) for working with the npm team to identify
  this issue. ([@othiym23](https://github.com/othiym23))
- [`0dc6875`](https://github.com/npm/npm/commit/0dc68757cffd5397c280bc71365d106523a5a052)
  `semver@4.3.2`: Package versions can be no more than 256 characters long.
  This prevents a situation in which parsing the version number can use
  exponentially more time and memory to parse, leading to a potential denial of
  service. Thanks to Adam Baldwin at Lift Security for bringing this to our
  attention.  ([@isaacs](https://github.com/isaacs))

The severity is medium and the impact is low, at least for now. We have no evidence that either of these vulnerabilities were being exploited in the wild. All the same, it would be good to get people upgrading.

This version also contains two important fixes to npm's git / GitHub support:
- [`eab6184`](https://github.com/npm/npm/commit/eab618425c51e3aa4416da28dcd8ca4ba63aec41)
  [#7766](https://github.com/npm/npm/issues/7766) One last tweak to ensure that
  GitHub shortcuts work with private repositories.
  ([@iarna](https://github.com/iarna))
- [`a840a13`](https://github.com/npm/npm/commit/a840a13bbf0330157536381ea8e58d0bd93b4c05)
  [#7746](https://github.com/npm/npm/issues/7746) Only fix up git URL paths when
  there are paths to fix up. ([@othiym23](https://github.com/othiym23))

Unfortunately, `npm@2.7.5` is missing the final piece of the git puzzle:
- [`b747593`](https://github.com/npm/npm/commit/b7475936f473f029e6a027ba1b16277523747d0b)
  [#7630](https://github.com/npm/npm/issues/7630) Don't automatically log all
  git failures as errors. `maybeGithub` needs to be able to fail without
  logging to support its fallback logic.
  ([@othiym23](https://github.com/othiym23))

In most cases, `npm@2.7.5` will work fine with private GitHub repositories, but it will log an ugly and misleading message about what it's doing. The above change (which is part of `npm@2.7.6`) takes care of that, but I still don't think it's a good idea to jump the gun with npm releases. I could maybe be persuaded otherwise.


othiym23:
Closing in favor of #15617.

PR #14572, opened 2015-04-08

jasnell:

### DIFF
--- a/doc/api/tls.markdown
+++ b/doc/api/tls.markdown
@@ -174,8 +174,7 @@ v0.10.38 behavior of only using the default cipher list on the server.
 ### Cipher List Precedence
 
 Note that the `--enable-legacy-cipher-list`, `NODE_LEGACY_CIPHER_LIST`,
-`--cipher-list` and `NODE_CIPHER_LIST` options are mutually exclusive. Only
-_one_ should be used at a time.
+`--cipher-list` and `NODE_CIPHER_LIST` options are mutually exclusive.
 
 If the `NODE_CIPHER_LIST` and `NODE_LEGACY_CIPHER_LIST` environment variables
 are both specified, the `NODE_LEGACY_CIPHER_LIST` setting will take precedence.
--- a/doc/api/tls.markdown
+++ b/doc/api/tls.markdown
@@ -164,6 +164,35 @@ Reverting back to the defaults used by older releases can weaken the security
 of your applications. The legacy cipher suites should only be used if absolutely
 necessary.
 
+NOTE: Due to an error in Node.js v0.10.38, the default cipher list only applied
+to servers using TLS. The default cipher list would _not_ be used by clients.
+This behavior has been changed in v0.10.39 and the default cipher list is now
+used by both the server and client when using TLS. However, when using
+`--enable-legacy-cipher-list=v0.10.38`, Node.js is reverted back to the
+v0.10.38 behavior of only using the default cipher list on the server.
+
+### Cipher List Precedence
+
+Note that the `--enable-legacy-cipher-list`, `NODE_LEGACY_CIPHER_LIST`,
+`--cipher-list` and `NODE_CIPHER_LIST` options are mutually exclusive. Only
+_one_ should be used at a time.
+
+If the `NODE_CIPHER_LIST` and `NODE_LEGACY_CIPHER_LIST` environment variables
+are both specified, the `NODE_LEGACY_CIPHER_LIST` setting will take precedence.
+
+The `--cipher-list` and `--enable-legacy-cipher-list` command line options
+will override the environment variables. If both happened to be specified, the
+right-most (second one specified) will take precedence. For instance, in the
+example:
+
+    node --cipher-list=ABC --enable-legacy-cipher-list=v0.10.38
+
+The v0.10.38 default cipher list will be used.
+
+    node --enable-legacy-cipher-list=v0.10.38 --cipher-list=ABC
+
+The custom cipher list will be used.
+
 ## tls.getCiphers()
 
 Returns an array with the names of the supported SSL ciphers.
@@ -173,6 +202,18 @@ Example:
     var ciphers = tls.getCiphers();
     console.log(ciphers); // ['AES128-SHA', 'AES256-SHA', ...]
 
+
+## tls.getLegacyCiphers(version)
+
+Returns a default cipher list used in a previous version of Node.js. The
+version parameter must be a string whose value identifies previous Node.js
+release version. The only value currently supported is `v0.10.38`.
+
+A TypeError will be thrown if: (a) the `version` is any type other than a
+string, (b) the `version` parameter is not specified, or (c) additional
+parameters are passed in. An Error will be thrown if the `version` parameter is
+passed in as a string but the value does not correlate to any known Node.js
+release for which a default cipher list is available.
 
 ## tls.createServer(options, [secureConnectionListener])
 
--- a/doc/api/tls.markdown
+++ b/doc/api/tls.markdown
@@ -181,7 +181,7 @@ If the `NODE_CIPHER_LIST` and `NODE_LEGACY_CIPHER_LIST` environment variables
 are both specified, the `NODE_LEGACY_CIPHER_LIST` setting will take precedence.
 
 The `--cipher-list` and `--enable-legacy-cipher-list` command line options
-will override the environment variables. If both happened to be specified, the
+will override the environment variables. If both happen to be specified, the
 right-most (second one specified) will take precedence. For instance, in the
 example:
 
--- a/lib/tls.js
+++ b/lib/tls.js
@@ -1330,6 +1330,20 @@ function normalizeConnectArgs(listArgs) {
   return (cb) ? [options, cb] : [options];
 }
 
+// return true if the --enable-legacy-cipher-list command line
+// switch, or the NODE_LEGACY_CIPHER_LIST environment variable
+// are set to v0.10.38 and the DEFAULT_CIPHERS equal the v0.10.38
+// list.
+function usingV1038Ciphers() {
+  var argv = process.execArgv;
+  if ((argv.indexOf('--enable-legacy-cipher-list=v0.10.38') > -1 ||
+       process.env['NODE_LEGACY_CIPHER_LIST'] === 'v0.10.38') &&
+      DEFAULT_CIPHERS === _crypto.getLegacyCiphers('v0.10.38')) {
+        return true;
+  }
+  return false;
+}
+
 exports.connect = function(/* [port, host], options, cb */) {
   var args = normalizeConnectArgs(arguments);
   var options = args[0];
@@ -1338,7 +1352,8 @@ exports.connect = function(/* [port, host], options, cb */) {
   var defaults = {
     rejectUnauthorized: '0' !== process.env.NODE_TLS_REJECT_UNAUTHORIZED
   };
-  if (DEFAULT_CIPHERS != _crypto.getLegacyCiphers('v0.10.38')) {
+  if (!usingV1038Ciphers()) {
+    // only set the default ciphers
     defaults.ciphers = DEFAULT_CIPHERS;
   }
   options = util._extend(defaults, options || {});
--- a/lib/tls.js
+++ b/lib/tls.js
@@ -1353,7 +1353,13 @@ exports.connect = function(/* [port, host], options, cb */) {
     rejectUnauthorized: '0' !== process.env.NODE_TLS_REJECT_UNAUTHORIZED
   };
   if (!usingV1038Ciphers()) {
-    // only set the default ciphers
+    // only set the default ciphers if we are _not_ using the
+    // v0.10.38 legacy cipher list. Node v0.10.38 had a bug
+    // that failed to set the default ciphers on the default
+    // options. This has been fixed in v0.10.39 and above.
+    // However, when the user explicitly tells node to revert
+    // back to using the v0.10.38 cipher list, node should
+    // revert back to the original v0.10.38 behavior.
     defaults.ciphers = DEFAULT_CIPHERS;
   }
   options = util._extend(defaults, options || {});
--- a/src/node_crypto.cc
+++ b/src/node_crypto.cc
@@ -4199,12 +4199,18 @@ const char* ToCString(const node::Utf8Value& value) {
 
 Handle<Value> DefaultCiphers(const Arguments& args) {
   HandleScope scope;
+  unsigned int len = args.Length();
+  if (len != 1 || !args[0]->IsString()) {
+    return ThrowException(Exception::TypeError(String::New("A single string parameter is required")));
+  }
   node::Utf8Value key(args[0]);
   const char * list = legacy_cipher_list(ToCString(key));
-  if (list == NULL) {
-    list = DEFAULT_CIPHER_LIST_HEAD;
+  if (list != NULL) {
+    return scope.Close(v8::String::New(list));
+  } else {
+    return ThrowException(Exception::Error(String::New(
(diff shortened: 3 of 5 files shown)


jasnell:
@misterdjules @mdawsonibm

misterdjules:
Can we add a test to make sure that `getLecagyCiphers` throws/doesn't throw as expected?

jasnell:
Done

misterdjules:
@jasnell Do we want to throw an error on a call to `getLegacyCiphers()` when the first parameter passed is not a string (undefined, null or anything else that is not a string)? The error message would be different, instead of `'Error: Unknown legacy cipher list'`, it would be something like `'Error: getLegacyCiphers requires a string as first parameter"`.

Also, I think we miss the "full" documentation for the `getLegacyCiphers` function in the API docs, including what the parameter is, its type, etc. Currently it is only briefly mentioned in the `Modifying the Default Cipher Suite` section.

jasnell:
@misterdjules  ... PR updated to address feedback.

jasnell:

misterdjules:
@jasnell Thank you, I think that's a significant improvement! :+1:

mdawsonibm:

test/simple/test-tls-cipher-list.js
```diff
@@ -44,6 +44,21 @@ function doTest(checklist, env, useswitch) {
...
+    case 5:
+      // Test --enable-legacy-cipher-list takes precedence over --cipher-list
+      args.unshift('--cipher-list=XYZ');
+      args.unshift('--enable-legacy-cipher-list=' + env);
+      break;
+    case 6:
+      // Test NODE_LEGACY_CIPHER_LIST takes precendence over --cipher-list
+      args.unshift('--cipher-list=XYZ');
+      if (env) options = {env:{"NODE_LEGACY_CIPHER_LIST": env}};
+      break;
```
misterdjules: We would probably need to have more tests than that to test the whole matrix of command line switches and environment variables. I would do that in a separate PR though, as this one is about `ls: throw an error on getLegacyCipher`.

misterdjules:
@jasnell I think @mhdawson meant that we should check for the type of the error thrown by `tls.getLegacyCiphers`. I didn't see anything related to that in the latest two commits.

I would not add a lot to this PR so that we can reason about it easily and land something reasonably quickly. We can work on other aspects of ciphers list related changes in other PRs.

Thank you again very much for your work!

jasnell:
@misterdjules after talking with @mhdawson today, I realize now that I missed an important aspect of your discussion with him regarding the order of precedence on the command line switches and environment variables, as well as the error type check in test test case. I'm going to make those additional changes and squash these PR's down this evening with the goal of landing them by Friday.

jasnell:
@misterdjules @mhdawson ... ok the changes discussed have been made and the multiple commits have been squashed down into two distinct commits.

lib/tls.js
```diff
@@ -1338,7 +1352,8 @@ exports.connect = function(/* [port, host], options, cb */) {
   var defaults = {
     rejectUnauthorized: '0' !== process.env.NODE_TLS_REJECT_UNAUTHORIZED
   };
-  if (DEFAULT_CIPHERS != _crypto.getLegacyCiphers('v0.10.38')) {
+  if (!usingV1038Ciphers()) {
```
misterdjules: Maybe add a comment to clarify why there's an exception for v0.10.38 here? Otherwise I'm concerned it's going to be difficult to understand a few months/years from now.

doc/api/tls.markdown
```diff
@@ -164,6 +164,35 @@ Reverting back to the defaults used by older releases can weaken the security
...
+This behavior has been changed in v0.10.39 and the default cipher list is now
+used by both the server and client when using TLS. However, when using
+`--enable-legacy-cipher-list=v0.10.38`, Node.js is reverted back to the
+v0.10.38 behavior of only using the default cipher list on the server.
+
+### Cipher List Precedence
+
+Note that the `--enable-legacy-cipher-list`, `NODE_LEGACY_CIPHER_LIST`,
+`--cipher-list` and `NODE_CIPHER_LIST` options are mutually exclusive. Only
+_one_ should be used at a time.
```
misterdjules: I would suggest removing `Only _one_ should be used at a time`, since the support for having some options/env vars override others is intentional and corresponds to valid use cases.

test/simple/test-tls-cipher-list.js
```diff
@@ -60,11 +45,185 @@ function doTest(checklist, env, useswitch) {
       });
 }
 
+// test that the command line switchs takes precedence
```
misterdjules: switchs -> switches.

test/simple/test-tls-cipher-list.js
```diff
@@ -60,11 +45,185 @@ function doTest(checklist, env, useswitch) {
...
+  doTest('ABC',
+         ['--cipher-list=ABC'],
+         {'NODE_LEGACY_CIPHER_LIST': 'v0.10.38'});
+
+
+  // test that --enable-legacy-cipher-list takes precence over both envars
+  doTest(V1038Ciphers,
+         ['--enable-legacy-cipher-list=v0.10.38'],
+         {
+           'NODE_LEGACY_CIPHER_LIST': 'v0.10.38',
```
misterdjules: Don't we want another value for `NODE_LEGACY_CIPHER_LIST` (e.g `ABC`), otherwise how do we know that the value for legacy cipher list was taken from the command line switch and not from the environment variable?

test/simple/test-tls-cipher-list.js
```diff
@@ -60,11 +45,185 @@ function doTest(checklist, env, useswitch) {
...
+   doTest('XYZ',
+          [
+            '--enable-legacy-cipher-list=v0.10.38',
+            '--cipher-list=XYZ'
+          ]);
+
+    // test the right-most command line option takes precedence
+    doTest('XYZ',
+           [
+             '--cipher-list=XYZ',
```
misterdjules: Do we want another value for the first `--cipher-list` command line switch (e.g `ABC`)? Otherwise how do we know that the value for ciphers list was taken from the second `--cipher-list` and not the first one?

test/simple/test-tls-cipher-list.js
```diff
@@ -60,11 +45,185 @@ function doTest(checklist, env, useswitch) {
...
+      rejectUnauthorized: false
+    }, function() {
+      socket.end();
+    });
+  }
+).toString();
+
+var test_count = 0;
+
+function doDefaultCipherTest(additional_args, env, failexpected) {
```
misterdjules: Minor: instead of passing booleans, I find passing options objects easier to read. In this specific case:
```
doDefaultCipherTest([], {}, test_uses_default_cipher_list);
```
turns into
```
doDefaultCipherTest([], {},  { failExpected: true });
```

test/simple/test-tls-cipher-list.js
```diff
@@ -60,11 +45,185 @@ function doTest(checklist, env, useswitch) {
...
+// created that intercepts the call to createCredentials and checks the output.
+// The following server code was adopted from test-tls-connect-simple.
+
+// note that the following function is written out to a string and
+// passed in as an argument to a child node instance.
+var script = (
+  function() {
+    var tls = require('tls');
+    var orig_createCredentials = require('crypto').createCredentials;
+    require('crypto').createCredentials = function(options) {
```
misterdjules: Instead of monkey patching `crypto.createCredentials`, what do you think about starting a server that sets its ciphers list to `RC4-MD5` only?

Since it's explicitly excluded from the ciphers list after node v0.10.38, the test could make sure that the client connection can only be established when passing `--enable-legacy-cipher-list=v0.10.38` or setting equivalent environment variables.

I believe it would make the implementation of this test much simpler.

misterdjules:
@jasnell Added a few comments, but it's definitely getting better, thank you!

test/simple/test-tls-cipher-list.js
jasnell: That's not quite what this is testing. This tests to make sure the default
ciphers are not set on the options when v0.10.38 legacy is set. The monkey
patch allows us to intercept the call and verify that the ciphers property
is indeed not set.

test/simple/test-tls-cipher-list.js
jasnell: In v0.10.39 there is only one legal value for this. Passing in anything but
v0.10.38 throws, even if it's overridden by the command line switch. For
v0.12.3 and master I specify different values.

test/simple/test-tls-cipher-list.js
jasnell: Yeah, actually that's just a typo... it should have been different

test/simple/test-tls-cipher-list.js
misterdjules: OK, thank you for the clarification. I would suggest also asserting that the monkey-patched version of `crypto.createCredentials` was called to avoid missing potential future regressions if `crypto.createCredentials` ends up being refactored and is not called anymore (like in v0.12).

Also, do we really need to spawn a server to run that test? If I understand correctly, `crypto.createCredentials` is called before any connection is initiated, so I would expect that we can perform the same test without starting a server.

test/simple/test-tls-cipher-list.js
```diff
@@ -19,36 +19,21 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-var spawn = require('child_process').spawn;
+var spawn  = require('child_process').spawn;
```
misterdjules: Style: the current style used does not align the `=` operator for assignment statements.

test/simple/test-tls-cipher-list.js
```diff
@@ -60,11 +45,192 @@ function doTest(checklist, env, useswitch) {
...
+  // over NODE_CIPHER_LIST
+  doTest(V1038Ciphers,
+         ['--enable-legacy-cipher-list=v0.10.38'],
+         {'NODE_CIPHER_LIST': 'XYZ'});
+
+  // test that --cipher-list takes precedence over NODE_LEGACY_CIPHER_LIST
+  doTest('ABC',
+         ['--cipher-list=ABC'],
+         {'NODE_LEGACY_CIPHER_LIST': 'v0.10.38'});
+
```
misterdjules: Style: unnecessary blank line.

test/simple/test-tls-cipher-list.js
```diff
@@ -60,11 +45,192 @@ function doTest(checklist, env, useswitch) {
...
+    doTest('XYZ',
+           [
+             '--cipher-list=ABC',
+             '--enable-legacy-cipher-list=v0.10.38',
+             '--cipher-list=XYZ'
+           ]);
+
+    // test that NODE_LEGACY_CIPHER_LIST takes precedence over
+    // NODE_CIPHER_LIST
+
```
misterdjules: Style: unnecessary blank line.

test/simple/test-tls-cipher-list.js
jasnell: I spawn the server purposefully to do a full roundtrip to make sure that the connection still works without those defaults being set on the client side.

jasnell:
@misterdjules @mhdawson

misterdjules:
@jasnell Thank you, I'll take a look on Sunday!

src/node_crypto.cc
```diff
@@ -4199,12 +4199,18 @@ const char* ToCString(const node::Utf8Value& value) {
 
 Handle<Value> DefaultCiphers(const Arguments& args) {
   HandleScope scope;
+  unsigned int len = args.Length();
+  if (len != 1 || !args[0]->IsString()) {
+    return ThrowException(Exception::TypeError(String::New("A single string parameter is required")));
```
misterdjules: Style: wrap at 80 columns.

test/simple/test-tls-cipher-list.js
```diff
@@ -60,11 +45,194 @@ function doTest(checklist, env, useswitch) {
...
+    '-e', require('util').format('(%s)()', script).replace(
+      'port: 0', 'port: ' + common.PORT)
+  ]);
+  var child = spawn(process.execPath, args, options);
+  child.stdout.
+    on('data', function(data) {
+      out += data;
+    }).
+    on('end', function() {
+      if (opts.failExpected && err === '') {
```
misterdjules: Why is this test done in the listener for the `'end'` event on the child process' standard output? The child seems to only write to the standard _error_ output.

test/simple/test-tls-cipher-list.js
```diff
@@ -60,11 +45,194 @@ function doTest(checklist, env, useswitch) {
...
+        console.error(options.ciphers);
+        process.exit(1);
+      }
+      return orig_createCredentials(options);
+    };
+    var socket = tls.connect({
+      port: 0,
+      rejectUnauthorized: false
+    }, function() {
+      socket.end();
```
misterdjules: If the client cannot connect, an `'error'` event would be emitted and node would write to stderr. The parent process would check that some data was read from the child's stderr, and it would make the test pass, but it shouldn't.

test/simple/test-tls-cipher-list.js
misterdjules: @jasnell OK, I understand now. Generally, testing a single characteristic/behavior in a test helps make it clearer and avoid false negatives.

In this case, the test doesn't make a difference between the child process writing to stderr because the client could not connect or because the `options.ciphers` value passed to `crypto.createCredentials` was not expected. See my comment here: https://github.com/joyent/node/pull/14572/files#r30068923.

If I understand correctly, `doDefaultCipherTest` seems to test _more than one thing_. It tests that `crypto.createCredentials` is called with the proper value for its `options.ciphers` argument, _and_ that a client that uses `--enable-legacy-cipher-list` can connect to a `tls.Server` instance started with the default options.

I think splitting that in two different simpler tests would make the tests more reliable and easier to understand.

test/simple/test-tls-cipher-list.js
```diff
@@ -60,11 +45,194 @@ function doTest(checklist, env, useswitch) {
...
+).toString();
+
+var test_count = 0;
+
+function doDefaultCipherTest(additional_args, env, opts) {
+  var options = {};
+  if (env) options.env = env;
+  var out = '', err = '';
+  additional_args = additional_args || [];
+  var args = additional_args.concat([
```
misterdjules: Another way to achieve the same thing would be to use `fork` and to pass an argument to the child process that allows to easily identify what code to run in the parent and in the child processes. Something similar to what is done in https://github.com/joyent/node/blob/v0.10/test/simple/test-domain-top-level-error-handler-throw.js. A search with `grep -rHn "'child'" test/simple` will show you many tests where this pattern is used.

That would require to have this test in a separate file, and I think this would make things simpler and cleaner overall.

jasnell:
@misterdjules ... I'll refactor that last test to split things out more. No problem there. Will see if I can get that in this morning. Beyond that, are there any other remaining issues on this?

jasnell:
@misterdjules

misterdjules:
@jasnell I would like to come up with a good solution for https://github.com/joyent/node/pull/23947 too, and add support for the new command line switches in `test/external/ssl-options`. I can add that since I'm familiar with how this test works.

We don't have to rush it though. As we discussed during the latest weekly meeting, if we want to release v0.12.3 before these changes are ready, we can revert the work in progress in v0.12 and include them back in time for the next v0.12.4 release.

jasnell:
Yes, I'm going to look at that one tonight again

misterdjules:
Moving to milestone v0.10.40, as v0.10.39 was released today.

misterdjules:
Closing in favor of #25603.

PR #14573, opened 2015-04-08

jasnell:

### DIFF
--- a/doc/api/tls.markdown
+++ b/doc/api/tls.markdown
@@ -204,6 +204,34 @@ Reverting back to the defaults used by older releases can weaken the security
 of your applications. The legacy cipher suites should only be used if absolutely
 necessary.
 
+NOTE: Due to an error in Node.js v0.10.38, the default cipher list only applied
+to servers using TLS. The default cipher list would _not_ be used by clients.
+This behavior has been changed in v0.10.39 and v0.12.x and the default cipher
+list is now used by both the server and client when using TLS. However, when
+using `--enable-legacy-cipher-list=v0.10.38`, Node.js is reverted back to the
+v0.10.38 behavior of only using the default cipher list on the server.
+
+### Cipher List Precedence
+
+Note that the `--enable-legacy-cipher-list`, `NODE_LEGACY_CIPHER_LIST`,
+`--cipher-list` and `NODE_CIPHER_LIST` options are mutually exclusive.
+
+If the `NODE_CIPHER_LIST` and `NODE_LEGACY_CIPHER_LIST` environment variables
+are both specified, the `NODE_LEGACY_CIPHER_LIST` setting will take precedence.
+
+The `--cipher-list` and `--enable-legacy-cipher-list` command line options
+will override the environment variables. If both happen to be specified, the
+right-most (second one specified) will take precedence. For instance, in the
+example:
+
+    node --cipher-list=ABC --enable-legacy-cipher-list=v0.10.38
+
+The v0.10.38 default cipher list will be used.
+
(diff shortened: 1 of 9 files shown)


jasnell:
@misterdjules @mdawsonibm

misterdjules:
Can we add a test to make sure that getLecagyCiphers throws/doesn't throw as expected?

jasnell:
Done

jasnell:

mdawsonibm:

jasnell:
@misterdjules @mhdawson ... ok, squashed the commits and made the additional changes discussed

doc/api/tls.markdown
```diff
@@ -215,10 +244,16 @@ Example:
 
 ## tls.getLegacyCiphers(version)
 
-Returns the legacy default cipher suite for the specified Node.js release.
+Returns a default cipher list used in a previous version of Node.js. The
+version parameter must be a string whose value identifies previous Node.js
+release version. The only values currently supported are `v0.10.38`,
+`v0.10.39`, and `v0.12.2`.
 
-Example:
```
misterdjules: Keeping an example (with code formatting) would be nice.

test/simple/test-tls-cipher-list.js
```diff
@@ -43,6 +43,54 @@ function doTest(checklist, env, useswitch) {
       // Test NODE_LEGACY_CIPHER_LIST
       if (env) options = {env:{"NODE_LEGACY_CIPHER_LIST": env}};
       break;
+    case 4:
```
misterdjules: Having arbitrary numbers determine the behavior of tests seems very confusing to me, because it's not clear when I read `doTest(V1038Ciphers, 'v0.10.38', 8);` that we test that `NODE_LEGACY_CIPHER_LIST takes precendence over NODE_CIPHER_LIST`.

Can we refactor that to make it clearer? Maybe we could have separate function clearly named for each test we want to perform? Parameters for each test would be local to these functions and make it clear what each test does.

lib/_tls_wrap.js
```diff
@@ -834,10 +834,13 @@ exports.connect = function(/* [port, host], options, cb */) {
 
   var defaults = {
     rejectUnauthorized: '0' !== process.env.NODE_TLS_REJECT_UNAUTHORIZED,
-    ciphers: tls.DEFAULT_CIPHERS,
     checkServerIdentity: tls.checkServerIdentity
   };
 
+  if (tls.DEFAULT_CIPHERS != tls.getLegacyCiphers('v0.10.38')) {
```
misterdjules: If one passes `--cipher-list='ECDHE-RSA-AES128-SHA256:AES128-GCM-SHA256:RC4:HIGH:!MD5:!aNULL:!EDH'`, my understanding is that `tls.DEFAULT_CIPHERS` will === `tls.getLegacyCiphers('v0.10.38'))`, and `default.ciphers` won't be set. I haven't tested it, so it may be a wrong assumption.

So I guess that here what is meant is more `if (user didn't pass --enable-legacy-cipher-list=v0.10.38) { ... }`, which is not equivalent to `if (tls.DEFAULT_CIPHERS != tls.getLegacyCiphers('v0.10.38'))`.

Thoughts?

misterdjules: Also, in general please use strict equality unless not possible. In this occurence, please use `!==` instead of `!=`.

jasnell:
@misterdjules @mhdawson ... ok, the PR has been rebased following the v0.12.3 cut and squashed down to just two commits against the v0.12 branch. @misterdjules please review and let me know if you feel this is ready to land.

jasnell:
Closing this. Taking a different approach.

PR #14574, opened 2015-04-08

jasnell:

### DIFF
--- a/doc/api/tls.markdown
+++ b/doc/api/tls.markdown
@@ -212,6 +212,35 @@ Reverting back to the defaults used by older releases can weaken the security
 of your applications. The legacy cipher suites should only be used if absolutely
 necessary.
 
+NOTE: Due to an error in Node.js v0.10.38, the default cipher list only applied
+to servers using TLS. The default cipher list would _not_ be used by clients.
+This behavior has been changed in v0.10.39 and v0.12.x and the default cipher
+list is now used by both the server and client when using TLS. However, when
+using `--enable-legacy-cipher-list=v0.10.38`, Node.js is reverted back to the
+v0.10.38 behavior of only using the default cipher list on the server.
+
+### Cipher List Precedence
+
+Note that the `--enable-legacy-cipher-list`, `NODE_LEGACY_CIPHER_LIST`,
+`--cipher-list` and `NODE_CIPHER_LIST` options are mutually exclusive. Only
+_one_ should be used at a time.
+
+If the `NODE_CIPHER_LIST` and `NODE_LEGACY_CIPHER_LIST` environment variables
+are both specified, the `NODE_LEGACY_CIPHER_LIST` setting will take precedence.
+
+The `--cipher-list` and `--enable-legacy-cipher-list` command line options
+will override the environment variables. If both happened to be specified, the
+right-most (second one specified) will take precedence. For instance, in the
+example:
+
+    node --cipher-list=ABC --enable-legacy-cipher-list=v0.10.38
+
+The v0.10.38 default cipher list will be used.
+
+    node --enable-legacy-cipher-list=v0.10.38 --cipher-list=ABC
+
+The custom cipher list will be used.
+
 ## tls.getCiphers()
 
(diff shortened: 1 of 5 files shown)


jasnell:
@misterdjules @mdawsonibm

misterdjules:
Can we add a test to make sure that `getLecagyCiphers` throws/doesn't throw as expected?

jasnell:
Done

jasnell:

mdawsonibm:

mdawsonibm:

jasnell:
@mhdawson @misterdjules ok, made the additional edits discussed and squashed the commits. Please verify.

lib/_tls_wrap.js
```diff
@@ -836,10 +836,13 @@ exports.connect = function(/* [port, host], options, cb */) {
 
   var defaults = {
     rejectUnauthorized: '0' !== process.env.NODE_TLS_REJECT_UNAUTHORIZED,
-    ciphers: tls.DEFAULT_CIPHERS,
     checkServerIdentity: tls.checkServerIdentity
   };
 
+  if (tls.DEFAULT_CIPHERS != tls.getLegacyCiphers('v0.10.38')) {
```
misterdjules: If one passes `--cipher-list='ECDHE-RSA-AES128-SHA256:AES128-GCM-SHA256:RC4:HIGH:!MD5:!aNULL:!EDH'`, my understanding is that `tls.DEFAULT_CIPHERS` will === `tls.getLegacyCiphers('v0.10.38'))`, and `default.ciphers` won't be set. I haven't tested it, so it may be a wrong assumption.

So I guess that here what is meant is more `if (user didn't pass --enable-legacy-cipher-list=v0.10.38) { ... }`, which is not equivalent to `if (tls.DEFAULT_CIPHERS != tls.getLegacyCiphers('v0.10.38'))`.

Thoughts?

misterdjules: Also, in general please use strict equality unless not possible. In this occurence, please use `!==` instead of `!=`.

jasnell: Yes, that is the case. I personally don't see this as a significant problem. If someone passes in the exact default cipher list as v0.10.38 then the likely intent is to revert the behavior back to v0.10.38, which means the default.ciphers not being set. It's an acceptable edge case IMHO but I'm good either way. It's a bit of a larger change tho since this does not yet specifically track whether `--enable-legacy-cipher-list` is passed vs. `--cipher-list`

test/simple/test-tls-cipher-list.js
```diff
@@ -40,6 +40,54 @@ function doTest(checklist, env, useswitch) {
       // Test NODE_LEGACY_CIPHER_LIST
       if (env) options = {env:{"NODE_LEGACY_CIPHER_LIST": env}};
       break;
+    case 4:
```
misterdjules: Having arbitrary numbers determine the behavior of tests seems very confusing to me, because it's not clear when I read `doTest(V1038Ciphers, 'v0.10.38', 8);` that we test that `NODE_LEGACY_CIPHER_LIST takes precendence over NODE_CIPHER_LIST`.

Can we refactor that to make it clearer? Maybe we could have separate function clearly named for each test we want to perform? Parameters for each test would be local to these functions and make it clear what each test does.

doc/api/tls.markdown
```diff
@@ -223,10 +252,15 @@ Example:
 
 ## tls.getLegacyCiphers(version)
 
-Returns the legacy default cipher suite for the specified Node.js release.
+Returns a default cipher list used in a previous version of Node.js. The
+version parameter must be a string whose value identifies previous Node.js
+release version. The only values currently supported are `v0.10.38`, `v0.10.39`, `v0.12.2` and `v0.12.3`.
 
-Example:
-    var cipher_suite = tls.getLegacyCiphers('v0.10.38');
```
misterdjules: Keeping an example (with code formatting) would be nice.

jasnell:
Closing this as it cannot land here now. Will be refactoring this change and opening a new PR against master on nodejs/node.

PR #14819, opened 2015-04-14
src: fix backport for SIGINT crash fix on FreeBSD

### LINKED ISSUE #9326: Segfault on FreeBSD
I'm running this simple source:
```javascript
var http = require("http");

http.createServer(function(request, response) {
  response.writeHead(200, {"Content-Type": "text/plain"});
  response.write("Hello World");
  response.end();
}).listen(5000);
```

It works fine until I hit Ctrl+C: it segfaults.

Here's the backtrace:
```
#0  0x000000080226f469 in swapcontext () from /lib/libthr.so.3
#1  0x000000080226f062 in sigaction () from /lib/libthr.so.3
#2  <signal handler called>
#3  0x00000008024f344a in _umtx_op () from /lib/libc.so.7
#4  0x00000008024e2c24 in sem_timedwait () from /lib/libc.so.7
#5  0x0000000000b53ef8 in v8::base::Semaphore::Wait ()
#6  0x00000000008c7e84 in v8::internal::SweeperThread::Run ()
#7  0x0000000000b5594d in v8::base::ThreadEntry ()
#8  0x000000080226a4f5 in pthread_create () from /lib/libthr.so.3
#9  0x0000000000000000 in ?? ()
```

I'm running FreeBSD 10.1-RELEASE with node installed from the ports.

If I run a script that contains only a `console.log` it doesn't crash when it exits.

misterdjules:
61fe1fe21ba281b05b90a632e2dad29eefb14a0a backported
b64983d77cc3ed2e4b585f3bfc8ed23802389a52 from io.js, but failed to
change nullptr to NULL, which lead to a build break on FreeBSD since the
current build system doesn't enable support for C++11.

This change replaces nullptr by NULL, and has been tested on
FreeBSD 10.1-RELEASE-p8.

Fixes #9326.

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -2790,7 +2790,7 @@ static void SignalExit(int signo) {
   struct sigaction sa;
   memset(&sa, 0, sizeof(sa));
   sa.sa_handler = SIG_DFL;
-  CHECK_EQ(sigaction(signo, &sa, nullptr), 0);
+  CHECK_EQ(sigaction(signo, &sa, NULL), 0);
 #endif
   raise(signo);
 }


cjihrig:
LGTM

misterdjules:
Landed in f99eaef, thank you @cjihrig!

PR #15514, opened 2015-04-15
Update dgram.js fixes # 3

pothan:
something


cjihrig:
Thanks, for opening a PR. I don't really see the point of this change though. If you provide a better description for why you think this is necessary, then we can revisit this.

PR #14592, opened 2015-04-10
deps: upgrade npm to 2.7.6

othiym23:
There are three fixes in `npm@2.7.6` that are especially pertinent to downstream packagers:
- [`b747593`](https://github.com/npm/npm/commit/b7475936f473f029e6a027ba1b16277523747d0b)
  [#7630](https://github.com/npm/npm/issues/7630) Don't automatically log all git failures as errors. `maybeGithub` needs to be able to fail without logging to support its fallback logic. ([@othiym23](https://github.com/othiym23))
- [`78005eb`](https://github.com/npm/npm/commit/78005ebb6f4103c20f077669c3929b7ea46a4c0d)
  [#7743](https://github.com/npm/npm/issues/7743) Always quote arguments passed
  to `npm run-script`. This allows build systems and the like to safely escape
  glob patterns passed as arguments to `run-scripts` with `npm run-script
  <script> -- <arguments>`. This is a tricky change to test, and may be
  reverted or moved to `npm@3` if it turns out it breaks things for users.
  ([@mantoni](https://github.com/mantoni))
- [`da015ee`](https://github.com/npm/npm/commit/da015eee45f6daf384598151d06a9b57ffce136e)
  [#7074](https://github.com/npm/npm/issues/7074) `read-package-json@1.3.3`:
  `read-package-json` no longer caches `package.json` files, which trades a
  very small performance loss for the elimination of a large class of really
  annoying race conditions. See [#7074](https://github.com/npm/npm/issues/7074)
  for the grisly details. ([@othiym23](https://github.com/othiym23))

There are the usual other array of smaller tweaks and fixes, which can be read about in [the release notes](https://github.com/npm/npm/releases/tag/v2.7.6). This PR also includes the changes in `npm@2.7.5`, which have [their own release notes](https://github.com/npm/npm/releases/tag/v2.7.5), including two significant security fixes to npm's dependencies.

PR #15617, opened 2015-04-17
deps: upgrade npm to 2.8.3

othiym23:
The npm team has been very busy over the last week, partially working on getting issues related to scoped modules straightened out prior to the launch of private packages on the npm, Inc. registry, but mostly to get git working as robustly as possible with npm, so I could stop thinking about it for a while. Because there were many intermediate releases as we fixed bugs, the combined changelog is massive, but I've worked hard to make the release notes readable, so:
- https://github.com/npm/npm/releases/tag/v2.8.0
- https://github.com/npm/npm/releases/tag/v2.8.1
- https://github.com/npm/npm/releases/tag/v2.8.2
- https://github.com/npm/npm/releases/tag/v2.8.3

The reason for the semver-minor version bump, taken from the 2.8.0 release notes, is:
- **EXPERIMENTAL FEATURE**: Add support for `github:`, `gist:`, `bitbucket:`,
  and `gitlab:` shorthand prefixes. GitHub shortcuts will continue to be
  normalized to `org/repo` instead of being saved as `github:org/repo`, but
  `gitlab:`, `gist:`, and `bitbucket:` prefixes will be used on the command
  line and from `package.json`. BE CAREFUL WITH THIS. `package.json` files
  published with the new shorthand syntax can _only_ be read by `npm@2.8.0` and
  later, and this feature is mostly meant for playing around with it. If you
  want to save git dependencies in a form that older versions of npm can read,
  use `--save-exact`, which will save the git URL and resolved commit hash of
  the head of the branch in a manner simiilar to the way that `--save-exact`
  pins versions for registry dependencies.  This is documented (so check `npm
  help install` for details), but we're not going to make a lot of noise about
  it until it has a chance to bake in a little more.

A large portion of the diff is changes to the tests that I made in an effort to clean them up and to make it easier for contributors to write new tests without getting confused about how they should look.

This supersedes #14421 and #14592.


misterdjules:
Superseded by #20311. Thank you once again @othiym23.

Issue #15447, opened 2015-04-15
setTimeout can’t work well after the process is blocked


ilc-opensource:
node.js version: 4d9c81b7e2522c5d5d9d35058cbb0bce1228d360 (latest version), v0.10.33, node-v0.12.2
Platform: Ubuntu 14.04.1 LTS
Architecture: x86_64

    var block = require('bindings')('hello'); // c++ addon
    var a = function() {
      var timer = (new Date()).getTime();
      console.log('before block:'+timer);
    
      block.hello(); // will be block at some time
    
      timer = (new Date()).getTime();
      console.log('after block:'+timer);
    
      setTimeout(a, 5000);
    };
    a();

In the above script, function a will be executed very 5000 milliseconds if the block.hello is not blocked; But if the block.hello is blocked in c++ code, the time point which function a will be executed again is much larger than 5000 milliseconds after block.hello is unblocked. The following is the output of the above script:
1   before block:1429079335210
2   after block:1429079335210
3   before block:1429079340215
4   after block:1429079366741
5   before block:1429079398298
6   after block:1429079398299
7   before block:1429079403304
8   after block:1429079403304
Time interval between line2 and line 3 is 5005; 
**Time interval between line4 and line 5 is 31557;**
Time interval between line6 and line 7 is 5005;

block.hello is blocked and unblocked between line 3 and line 4.

## How to reproduce the issue?
1.	Download the tar file and decompress; https://www.dropbox.com/s/1y299n1y1uy69ij/node_js_debug.tar?dl=0
2.	cd node_js_debug;
3.	sh build.sh;
4.	node a.js;
5.	In another terminal, ./setFrontEndApp 123456; after a while, ./setFrontEndApp 0

## How to fix?
I find we should recalculate now each time when processing the pending time events.

diff --git a/lib/timers.js b/lib/timers.js
index 68f87d7..c6af7f9 100644
--- a/lib/timers.js
+++ b/lib/timers.js
@@ -80,11 +80,10 @@ function listOnTimeout() {

       debug('timeout callback %d', msecs);
    
    -  var now = Timer.now();
    -  debug('now: %s', now);
    -
       var diff, first, threw;
       while (first = L.peek(list)) {
    +    var now = Timer.now();
    +    debug('now: %s', now);
         diff = now - first._idleStart;
         if (diff < msecs) {
           list.start(msecs - diff, 0);

obastemur:
[...]

PR #17203, opened 2015-04-21
timers: fix timeout when added in timer's callback

### LINKED ISSUE #9333: Debugger does not respect timer
This was originally https://github.com/joyent/node/issues/9319, but I've reduced the case down more.
Note: this is happening with node 0.10.31 and above.

Given code `example.js`:

```javascript
setInterval(function() {  
  console.log(1)
  debugger
  console.log(2);
}, 10);
```

ok I enter into debugger now:

```
$ node debug ./example.js 
< Debugger listening on port 5858
connecting to port 5858... ok
break in example.js:1
> 1 setInterval(function() {  
  2   console.log(1)
  3   debugger
debug> c
< 1
break in example.js:3
  1 setInterval(function() {  
  2   console.log(1)
> 3   debugger
  4   console.log(2);
  5 }, 10);
debug> c    //*********now wait 1min before you hit enter******//
< 2 //********now it gets stuck here for 1min before it moves onto the next break point****//
```

My assumption is that the break points are messing up node's timer some how.

This may seem like a small issue in the debugger, but it's causing huge pains for Protractor users because all of protractor's code (which is using webdriver) is in a giant setInterval.

reference: 
https://github.com/angular/protractor/issues/1890
https://github.com/angular/protractor/issues/1899
https://github.com/angular/protractor/issues/1822
https://github.com/angular/protractor/issues/1890
https://github.com/angular/protractor/issues/1893

and the only workaround for that right now is to downgrade node to 0.10.30 or lower. 

Thanks!

misterdjules:
When a timer is added in another timer's callback, its underlying timer
handle will be started with a timeout that is actually incorrect.

The reason is that  the value that represents the current time is not
updated between the time the original callback is called and the time
the added timer is processed by timers.listOnTimeout. That leads the
logic in timers.listOnTimeout to do an incorrect computation that makes
the added timer fire with a timeout of scheduledTimeout +
timeSpentInCallback.

This change fixes that and make timers scheduled within other timers'
callbacks fire as expected.

Fixes #9333 and #15447.

### DIFF
--- a/lib/timers.js
+++ b/lib/timers.js
@@ -83,10 +83,19 @@ function listOnTimeout() {
   debug('timeout callback ' + msecs);
 
   var now = Timer.now();
-  debug('now: ' + now);
+  debug('now: %d', now);
 
   var first;
   while (first = L.peek(list)) {
+    // If the previous iteration caused a timer to be added,
+    // update the value of "now" so that timing computations are
+    // done correctly. See test/simple/test-timers-blocking-callback.js
+    // for more information.
+    if (now < first._monotonicStartTime) {
+      now = Timer.now();
+      debug('now: %d', now);
+    }
+
     var diff = now - first._monotonicStartTime;
     if (diff < msecs) {
       list.start(msecs - diff, 0);
--- a/test/common.js
+++ b/test/common.js
@@ -218,3 +218,11 @@ exports.hasMultiLocalhost = function hasMultiLocalhost() {
   t.close();
   return ret === 0;
 };
+
+exports.busyLoop = function busyLoop(time) {
+  var startTime = new Date().getTime();
+  var stopTime =  startTime + time;
+  while (new Date().getTime() < stopTime) {
+    ;
+  }
+};
--- a/test/simple/test-timers-blocking-callback.js
+++ b/test/simple/test-timers-blocking-callback.js
@@ -1,0 +1,99 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+/*
+ * This is a regression test for https://github.com/joyent/node/issues/15447
+ * and https://github.com/joyent/node/issues/9333.
+ *
+ * When a timer is added in another timer's callback, its underlying timer
+ * handle was started with a timeout that was actually incorrect.
+ *
+ * The reason was that the value that represents the current time was not
+ * updated between the time the original callback was called and the time
+ * the added timer was processed by timers.listOnTimeout. That lead the
+ * logic in timers.listOnTimeout to do an incorrect computation that made
+ * the added timer fire with a timeout of scheduledTimeout +
+ * timeSpentInCallback.
+ *
+ * This test makes sure that a timer added by another timer's callback
+ * fire with the expected timeout.
+ *
+ * It makes sure that it works when the timers list for a given timeout is
+ * empty (see testAddingTimerToEmptyTimersList) and when the timers list
+ * is not empty (see testAddingTimerToNonEmptyTimersList).
+ */
+
+var assert = require('assert');
+var common = require('../common');
+
+var TIMEOUT = 100;
+
+var nbBlockingCallbackCalls = 0;
+var latestDelay = 0;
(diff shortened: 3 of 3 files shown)


lib/timers.js
```diff
@@ -82,11 +82,11 @@ function listOnTimeout() {
 
   debug('timeout callback ' + msecs);
 
-  var now = Timer.now();
-  debug('now: ' + now);
-
   var first;
   while (first = L.peek(list)) {
+    var now = Timer.now();
```
indutny: Maybe we should do it only if the list was changed? I believe this will come with a performance loss for a many-timer case.

lib/timers.js
```diff
@@ -82,11 +82,11 @@ function listOnTimeout() {
...
+    debug('now: ' + now);
```
indutny: I'd rather do `debug('now: %d', now);` for performance reasons.

trevnorris: ditto.

lib/timers.js
misterdjules: Good point, I added a `_addedTimer` property on timers lists to determine when a timer has been added to the list. There are other ways to do that, but this one seemed to be the more lightweight and less intrusive.

lib/timers.js
misterdjules: Good point too! Updated the PR.

misterdjules:
@indutny @trevnorris  I updated the PR with your review comments. I also took the opportunity to slightly refactor the included test:
* `busyLoop` is now in `test/common.js`.
* More importantly, the test make sure that it covers both cases of the timers list being empty and non-empty when the timer is added after a long-running timer callback.

Please let me know what you think. Thank you!

indutny:
LGTM! Thank you!

obastemur:
@misterdjules  I don't think this is a better way in terms of performance. This implementation calls timer.now twice instead of one.

trevnorris:
@obastemur I believe this is a bug fix. If it's comparable in performance then that's fine.

indutny:
@obastemur I think it calls it only once...

obastemur:
@trevnorris I was comparing this to original suggested fix. This is a different implementation since it tries to call timer.now less. IMHO timer.now doesn't cost anything serious. I had made run a perf. test on a range of devices. The most scary one takes 0.001 ms on a mips32r2 board which is still nothing. Yet, this implementation calls at least +1. If there is only one timeout. It calls twice. 

@indutny it calls twice. First calls before the loop and second a timeout needs to be active to be there right?

indutny:
@obastemur it should not invoke `now()` unless new timer is added to the same bucket.

obastemur:
@indutny I don't think it's very hard to test and see a basic setTimeout case.

indutny:
@obastemur could you please share a test case with us?

obastemur:
```setTimeout(function(){ console.log("Hello!");}, 1);```

**Comment by indutny** [comment]
@obastemur 

```
NODE_DEBUG=timer ./node 1
timeout callback 1
now: %d
54512505
Hello!
1 list empty
```

As you can see `debug('now %d')` is invoked only once.

**Comment by obastemur** [comment]
@indutny My bad. After you insisting it won't. I've checked it again. Seems like the result was coming from the original fix. Somehow https://github.com/misterdjules/node/tree/8b00d748aee6c1cc3f06faea51339b74b4763d8c didn't build on the test machine and results mixed. 

This one calls at least 25% less (if system wasn't busy)

Sorry for the confusion.

**Comment by indutny** [comment]
No problem at all, @obastemur . I have seen this too, before @misterdjules fixed the PR.

**Comment by obastemur** [comment]
a minor detail. I've applied this patch to JXcore to run the tests on our test environment. After this fix, test/simple/test-timers-ordering is no longer a reliable test case. It can be very hard to reproduce on a single run but I'm running the test cases on multiple devices at once and it fails as expected. (bsd, osx, debian on mips..) This doesn't mean the fix is wrong though. As expected, if the system is busy, multiple timeouts may end-up inside the same cycle which this test case fails to measure.

**Comment by misterdjules** [comment]
Thank you @indutny @obastemur and @trevnorris for your feedback!

Ran the tests suite and found no regression on [Windows](http://jenkins.nodejs.org/job/node-review-windows-julien/46/) and on [UNIX](http://jenkins.nodejs.org/job/node-review-unix-julien/54/).

**Comment by misterdjules** [comment]
@joyent/node-coreteam Added to the 0.10.39 milestone as we may want to release that with the next release.

**Comment by misterdjules** [comment]
@obastemur @trevnorris @indutny As I was reviewing this change one last time before landing it, I thought that its implementation could and should be simpler. This resulted in https://github.com/misterdjules/node/commit/52c35bd5c0f06e8d52d686ac6d5f77cbf014a34a that builds on top of the previous commit in the same PR.

If that looks better to you too, I'll squash both commits and land the resulting new commit.

Thank you!

**Comment by misterdjules** [comment]
@indutny ping, did you have some time to read [my latest comment](https://github.com/joyent/node/pull/17203#commitcomment-10987685)?

File: lib/timers.js [hunk 3]
```diff
@@ -83,10 +83,19 @@ function listOnTimeout() {
   debug('timeout callback ' + msecs);
 
   var now = Timer.now();
-  debug('now: ' + now);
+  debug('now: %d', now);
```
saquibkhan: ```debug('now: %d', now);``` Is using %d appropriate? or should we use %s, now can also have double value right?

lib/timers.js
```diff
@@ -182,6 +191,8 @@ exports.active = function(item) {
       item._monotonicStartTime = Timer.now();
       L.append(list, item);
     }
+
+    list = lists[msecs];
```
saquibkhan: Can you explain why this line of code is needed?

indutny:
LGTM

lib/timers.js
misterdjules: @saquibkhan It is not needed, I forgot to remove it from the previous patch, good catch :+1: and thank you for taking the time to review this change :)

lib/timers.js
misterdjules: @saquibkhan `now` is a non negative 64 bits integer number that is stored as a double if its value is larger than a V8 SMI. Plus the `%d` format supports non-integer numbers since it uses the function `Number` to convert the parameter corresponding to this format to a number.

bootstraponline:
:+1:

misterdjules:
Still no regression on [UNIX](http://jenkins.nodejs.org/job/node-review-unix-julien/57/) or [Windows](http://jenkins.nodejs.org/job/node-review-windows-julien/49/), landing.

misterdjules:
@joyent/node-collaborators Rebased, finally landing in v0.10.

misterdjules:
@joyent/node-collaborators Landed in d38e865fce93c00cb128034202fb1e26987efdbc, thank you @obastemur, @indutny and @trevnorris!

PR #18204, opened 2015-04-21
net: do not set V4MAPPED on FreeBSD

### LINKED ISSUE #8540: http.get fails on Android and FreeBSD
Hi,

when making a simple http request I'm getting the following error:

<pre>
Error: getaddrinfo EAI_BADFLAGS
    at Object.exports._errnoException (util.js:742:11)
    at errnoException (dns.js:46:15)
    at Object.onlookup [as oncomplete] (dns.js:91:26)
</pre>

Did a few tests and tried to follow the error through the code, but at this point all I can say is that I'm lost. Any ideas? As usually, I'll be happy to help on fixing this.

Resulting object from calling http.get can be found [here](http://pastebin.com/VAYJXPaE)

misterdjules:
V4MAPPED is not supported on recent FreeBSD versions, at least on 10.1.
Thus, do not set this flag in net.connect on FreeBSD.

Fixes #8540.

### DIFF
--- a/doc/api/dns.markdown
+++ b/doc/api/dns.markdown
@@ -259,8 +259,9 @@ The following flags can be passed as hints to `dns.lookup`.
 of addresses supported by the current system. For example, IPv4 addresses
 are only returned if the current system has at least one IPv4 address
 configured. Loopback addresses are not considered.
-- `dns.V4MAPPED`: If the IPv6 family was specified, but no IPv6 addresses
-were found, then return IPv4 mapped IPv6 addresses.
+- `dns.V4MAPPED`: If the IPv6 family was specified, but no IPv6 addresses were
+found, then return IPv4 mapped IPv6 addresses. Note that it is not supported
+on some operating systems (e.g FreeBSD 10.1).
 
 ## Implementation considerations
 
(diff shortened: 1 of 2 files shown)


misterdjules:
Tested on FreeBSD 10.1-RELEASE-p6.

misterdjules:
I didn't add a regression test for this specific problem since `net.connect` is used in lots of different places in the current simple tests suite.

cjihrig:
LGTM if the CI is happy. It might be worth adding this to the docs.

indutny:
LGTM

misterdjules:
@cjihrig @indutny Thanks! @cjihrig I don't think we currently mention the usage of V4MAPPED in the API docs, could you please elaborate on what you had in mind?

cjihrig:
@misterdjules maybe just mention [here](https://nodejs.org/api/dns.html#dns_supported_getaddrinfo_flags) that `V4MAPPED` does not work on FreeBSD.

misterdjules:
@cjihrig Good catch, will do and thank you!

misterdjules:
@cjihrig Updated.

cjihrig:
LGTM

misterdjules:
Thank you, landed in 0e392f3.

PR #18456, opened 2015-04-22
fix a bug when ./configure --without-ssl

ddnionio:
When ./configure --without-ssl and make, it will report no definition of SSL2_ENABLE and SSL3
_ENABLE. Because they are declared in "node_crypto.h", and "node_crypto.h" will not be included here when without ssl

#if HAVE_OPENSSL
#include "node_crypto.h"
#endif


jasnell:
@ddnionio ... Fixing the bug looks fine but the #6ad26b2 commit is wonky here. Can you fix this PR up so that it's just the actual fix.

ddnionio:
@jasnell, sorry about the mistake in pulling this request.
I have already fixed it and re-pulled as #21853.
Thanks

jasnell:
Closed as duplicate

PR #18957, opened 2015-04-22
Add require() lines to child.stdio example

enaqx:


cjihrig:
Would you mind updating the rest of this example to use single quotes, as that seems to be the style throughout the rest of this file? Other than that, LGTM.

jasnell:
Closing this here. The fix already landed in nodejs/node. Can cherry pick over to v0.12 if necessary.

PR #21853, opened 2015-04-26
fix the bug when ./configure --without-ssl (fix the pr #18456)

ddnionio:
When ./configure --without-ssl and make, it will report no definition of SSL2_ENABLE and SSL3
_ENABLE. Because they are declared in "node_crypto.h", and "node_crypto.h" will not be included here when without ssl

#if HAVE_OPENSSL
#include "node_crypto.h"
#endif

fix the pr #18456


jasnell:
Closing as duplicate of https://github.com/joyent/node/pull/25862

PR #18206, opened 2015-04-22
V8 lateral 3.28 update

misterdjules:
This continues the work done in https://github.com/joyent/node/pull/9185. It's basically the same PR with some added floating patches and the hidden files put back in.

Tested on [UNIX](http://jenkins.nodejs.org/job/node-review-unix-julien/53/) and [Windows](http://jenkins.nodejs.org/job/node-review-windows-julien/45/). The one test failing on Windows might be a flakey test, we'll need to make sure of that before landing this PR.

Fixes https://github.com/joyent/node/issues/9241.

Issue #14411, opened 2015-04-02
V8 Lateral upgrade - float V8 patch to address infinite loop


mhdawson:
With the lateral upgrade to V8 discussed here: https://github.com/joyent/node/pull/9185

The following 
   node --harmony --use-strict -e 'for (let i = 0; i < 3; ++i) { if (i == 1) { continue; } }'

results in an infinite loop.

This is an instance of the problem reported in V8 Issue 3683 - https://code.google.com/p/v8/issues/detail?can=2&q=3683&colspec=ID%20Type%20Status%20Priority%20Owner%20Summary%20HW%20OS%20Area%20Stars&id=3683    

The corresponding fix is: https://chromium.googlesource.com/v8/v8.git/+/b17eaaa5755e625493c5fe537f42b58838923c52

Cherry-picking that commit on top of 3.28.71.19 resolves the issue.

In V8 releases, the fix was rolled into 3.31.6:

commit 8b0d1187aaea32f767f0daf152434b3681c192ca
Author: v8-autoroll <v8-autoroll@chromium.org>
Date:   Fri Nov 14 18:01:38 2014 -0800

    Version 3.31.6 (based on 4a4158f3632ba2b18f6525a132891d480ea40bad)

    Fix desugaring of let bindings in for loops to handle continue properly (issue 3683).

    Performance and stability improvements on all platforms.

    Cr-Commit-Position: refs/heads/candidates@{#25280}

Opening this issue to track floating this change once the lateral upgrade is in. 

Along with cherry-picking the commit across we also need to add a test.

misterdjules:
@mhdawson Just to make sure we're tracking this issue correctly, this floating patch would fix #9113 right?

misterdjules:
Fixed in 2a5f4bd7ce40b9537c5d77558687858c7246d896 and cb55a05f528f57e350063f618274b5a182261387

PR #23948, opened 2015-04-28
v8: fix let bindings in for loops

### LINKED ISSUE #9113: Bug: for loop becomes endless loop in 0.11.15

misterdjules:
This PR:
- Adds [the floating patch that fixes let bindings and continue statements in for loops](https://github.com/joyent/node/commit/072460265226c047369558b23e9ff2748965bf6c).
- Adds [a test](https://github.com/joyent/node/commit/00ca880a7013d75820bdf087698930d6fc11cb92) that makes sure that this patch is floated in future V8 upgrades if needed.

They are in two separate commits so that every time we float the patch, we don't float the test too.

This PR is based off [another PR that downgrades V8 to the actual proper stable version for node v0.12](https://github.com/joyent/node/pull/18206), which is the reason why there are more commits than the two highlighted above.

Once #18206 lands, this PR will be rebased on top of v0.12 and will only include these two commits.

Fixes #9113 and #14411.

/cc @joyent/node-coreteam  and especially @mhdawson.

### DIFF
--- a/deps/v8/src/parser.cc
+++ b/deps/v8/src/parser.cc
@@ -2846,29 +2846,33 @@ Statement* Parser::DesugarLetBindingsInForStatement(
   //
   // We rewrite a for statement of the form
   //
-  //  for (let x = i; cond; next) body
+  //  labels: for (let x = i; cond; next) body
   //
   // into
   //
   //  {
-  //     let x = i;
-  //     temp_x = x;
-  //     flag = 1;
-  //     for (;;) {
-  //        let x = temp_x;
-  //        if (flag == 1) {
-  //          flag = 0;
-  //        } else {
-  //          next;
-  //        }
+  //    let x = i;
+  //    temp_x = x;
+  //    first = 1;
+  //    outer: for (;;) {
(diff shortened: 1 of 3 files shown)


cjihrig:
The test LGTM. The floating patch appears to be correct.

misterdjules:
Thank you @cjihrig!

misterdjules:
Rebased on top of v0.12, landing soon.

misterdjules:
Landed in 2a5f4bd7ce40b9537c5d77558687858c7246d896 and cb55a05f528f57e350063f618274b5a182261387.

PR #24405, opened 2015-04-29
Add dns.resolveURI(), support for DNS URI records

anthonykirby:
This patch adds  support for URI records, as described at:
 http://tools.ietf.org/html/draft-faltstrom-uri
An RR value (= 256) was assigned by IANA in 2011, and URI records are
 supported by recent versions of BIND.

The change includes modification to the c-ares dependency.  I've also
 submitted an identical patch to the c-ares project (http://c-ares.haxx.se)
 but it doesn't seem very active, nor does node seem to be pulling all their
 changes, so I suggest it's appropriate to patch node/deps/cares directly.

### DIFF
--- a/deps/cares/cares.gyp
+++ b/deps/cares/cares.gyp
@@ -71,6 +71,7 @@
         'src/ares_parse_soa_reply.c',
         'src/ares_parse_srv_reply.c',
         'src/ares_parse_txt_reply.c',
+        'src/ares_parse_uri_reply.c',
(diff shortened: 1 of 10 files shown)


jasnell:
@anthonykirby .. thank you for the PR. Because of the transition to http://github.com/nodejs/node, if you'd like to pursue this further, this would need to be opened as a new PR over there. It's not going to be able to land here.

PR #25102, opened 2015-05-01
don't  need to repeat.

kuldipem:


jasnell:
Closing this here. Opened a new PR with other bundled doc updates in http://github.com/nodejs/node, wll land this with that.

PR #25109, opened 2015-05-05
Add missing `options` argument to createSecurePair.

socketpair:
Also helps in implementation of #6204, where some ptions passed to
createSecurePair and ignored before this patch.

These options are very helpful if someone wants to pass
options.servername or options.SNICallback to securepair.

### DIFF
--- a/doc/api/tls.markdown
+++ b/doc/api/tls.markdown
@@ -563,7 +563,7 @@ publicly trusted list of CAs as given in
 <http://mxr.mozilla.org/mozilla/source/security/nss/lib/ckfw/builtins/certdata.txt>.
 
 
-## tls.createSecurePair([context][, isServer][, requestCert][, rejectUnauthorized])
+## tls.createSecurePair([context][, isServer][, requestCert][, rejectUnauthorized][, options])
 
 Creates a new secure pair object with two streams, one of which reads/writes
 encrypted data, and one reads/writes cleartext data.
@@ -581,6 +581,8 @@ and the cleartext one is used as a replacement for the initial encrypted stream.
  - `rejectUnauthorized`: A boolean indicating whether a server should
    automatically reject clients with invalid certificates. Only applies to
    servers with `requestCert` enabled.
+
+ - `options`: An object with common SSL options. See [new tls.TLSSocket][].
 
 `tls.createSecurePair()` returns a SecurePair object with `cleartext` and
 `encrypted` stream properties.
--- a/test/simple/test-tls-securepair-fiftharg.js
+++ b/test/simple/test-tls-securepair-fiftharg.js
@@ -1,0 +1,69 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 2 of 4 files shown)


indutny:
Great! May I ask you to add a test to not let this slip away again?

socketpair:
I do not know how to write test for that case...Also I think this bug can be merged without test :). + I updated documentation (not sure if syntax is correct)

socketpair:
Added tests, but do not know if working. You may comment sources. Now, I'm trying to build and run tests

socketpair:
Does not work...seems not that bug:


```
=== release test-tls-securepair-fiftharg ===                    
Path: pummel/test-tls-securepair-fiftharg
node: ../src/async-wrap-inl.h:98: v8::Handle<v8::Value> node::AsyncWrap::MakeCallback(v8::Handle<v8::String>, int, v8::Handle<v8::Value>*): Assertion `cb_v->IsFunction()' failed.
Command: out/Release/node ********/node/test/pummel/test-tls-securepair-fiftharg.js
```

indutny:
@socketpair there is a bug that should be fixed by https://github.com/nodejs/io.js/pull/1720

indutny:
cc @misterdjules please backport this if it sounds good to you! ;)

indutny:
Btw, why put this test into pummel folder?

socketpair:
I don't know. I just see some tests about tls in that folder, and put my test in same folder. I can change PR if you say what to fix.

indutny:
I'd go with `test/simple` ;) Btw, you could probably try to rebase your branch on top of nodejs/io.js#1720 to include the fix commit in this PR, and to make sure that it builds correctly.

socketpair:
I have cherry-picked commit from IOjs that fixed bug in C++

socketpair:
I have moed test from `pummel` to `simple` dir

socketpair:
Now all tests passed! yay! Please merge!

doc/api/tls.markdown
```diff
@@ -582,6 +582,8 @@ and the cleartext one is used as a replacement for the initial encrypted stream.
    automatically reject clients with invalid certificates. Only applies to
    servers with `requestCert` enabled.
 
+ - `options`: An object with common SSL options. See [new tls.TLSSocket][].
```
socketpair: do not know if markdown syntax is correct here

socketpair:
How to speedup merging? who is responsible?

test/simple/test-tls-securepair-fiftharg.js
```diff
@@ -0,0 +1,63 @@
...
+    var catched_servername = null;
+    var pair = tls.createSecurePair(sslcontext, true, false, false, {
+        SNICallback: function(servername, cb) {
+            catched_servername = servername;
+        }
+    });
+
+    // captured trafic from browser's request to https://www.google.com
+    var ssl_hello_with_SNI = new Buffer([
+        22,3,1,2,0,1,0,1,252,3,3,123,36,32,232,194,245,125,208,188,197,238,164,123,101,37,21,189,4,223,219,197,179,76,
```
indutny: Sorry, but we use 80 column limit everywhere. I may only suggest to convert this to hex and split it before it gets to 80 column. It will be much more compact this way.

indutny:
Sorry for delay! Just one minor nit, otherwise LGTM!

socketpair:
Hex encoding takess too much space. Just reformat decimal representation.

test/simple/test-tls-securepair-fiftharg.js
socketpair: fixed

indutny:
@socketpair what do you think about reopening this PR for io.js master branch? I have just landed that fix commit that your PR depended upon. ;)

socketpair:
I'm confused about everything? and forgot what is happening... :( There are nodejs, iojs and two patches in both projects.

I want all fixes to be in both projects. What I should do so?

indutny:
@socketpair it can't go into v0.12, and so the next release of node is io.js :) I would suggest to open PR for io.js master ;)

indutny:
Don't forget to cc me there ;)

jasnell:
@socketpair .. just to reiterate what @indutny said. This is not going to be able to land here. The right thing to do here would be to open a new PR against master on http://github.com/nodejs/node in order to get this landed.

socketpair:
@jasnell thanks for detailed explanation what should be done.
@indutny CC to you :)

https://github.com/nodejs/node/pull/2441


Should I close this pull request ?

jasnell:
@socketpair .. no problem. I'll go ahead and close the PR here. Thanks!

PR #25138, opened 2015-05-06
test: Extend timeout due to failure on AIX

mhdawson:
The loopback on AIX is slower by default than on other platforms
and we've seen a number of tests fail on AIX for this reason. This
looks to be another instance.  Changing the test to bind to the
host ip instead of the loopback makes it pass reliably.

This change extends the timeout so that it passes reliably on AIX
even with the slower loopback behaviour

```
modified:   test/simple/test-tls-wrap-timeout.js
```


cjihrig:
LGTM

mhdawson:
Landed as fcbf226a944ab89400bfbc336b4ee4d28852189b

PR #25139, opened 2015-05-06
Fixing doc location for tls.connect option checkServerIdentity

tlivings:


jasnell:
Close this here. This has already been fixed in nodejs/node. Can cherrypick that fix back to v0.12 if necessary

PR #25141, opened 2015-05-06
deps: update libuv to 1.5.0

saghul:
R=@trevnorris, @misterdjules


misterdjules:
@joyent/node-coreteam Adding to the 0.12.3 milestone, it would be great to have a few members of the core team reviewing this PR to determine if we want to include that in the upcoming v0.12.3 release. Thank you!

misterdjules:
Thank you very much @saghul! Tests running on [Windows](http://jenkins.nodejs.org/job/node-review-windows-julien/48/) and [UNIX](http://jenkins.nodejs.org/job/node-review-unix-julien/56/).

deps/uv/include/uv-version.h
```diff
@@ -31,9 +31,9 @@
  */
 
 #define UV_VERSION_MAJOR 1
-#define UV_VERSION_MINOR 4
-#define UV_VERSION_PATCH 2
-#define UV_VERSION_IS_RELEASE 0
-#define UV_VERSION_SUFFIX "node1"
```
misterdjules: Just so that we don't forget about that, if we need to float additional patches, we'll need to put back this version suffix.

misterdjules:
All tests pass on [UNIX](http://jenkins.nodejs.org/job/node-review-unix-julien/56/) and on [Windows](http://jenkins.nodejs.org/job/node-review-windows-julien/48/).

cjihrig:
I'm +1 to including it.

trevnorris:
LGTM

misterdjules:
@saghul The Node.js Jenkins CI has some tests failing on [SmartOS](http://jenkins.nodejs.org/job/libuv-v1.x/54/label=smartos/tapTestReport/) and [OSX](http://jenkins.nodejs.org/job/libuv-v1.x/54/label=osx-build/tapTestReport/). Does that look like genuine failures, flakey tests or anything to worry about to you?

saghul:
@misterdjules those are all known failures on those platforms. :disappointed:

misterdjules:
@saghul Thank you again :) LGTM.

saghul:
> Just FYI, we should be able to remove that floating patch by using the RHEL developer toolset on our CentOS Jenkins agent. Hopefully someone will be able to do that work soon.

Feel free to drop it when merging if need be :-)

misterdjules:
@saghul That won't be before landing this PR, see https://github.com/joyent/node-infra/issues/4 for more context. Also, why bumping the minor version number? There doesn't seem to be any addition to the public API, unless I missed something.

saghul:
@misterdjules Ah, I see. 

Initially the minor number bump was due to the new APIs for customizing the memory allocator, but I ended up reverting those. Nevertheless, I decided to bump the minor since we ended up packing many bugfixes, some resulting in "sort of new features" such as the ability to swap stdout and stderr on child processes. A bit of a stretch, yeah, but numbers are free! :-)

misterdjules:
Landing with http://jenkins.nodejs.org/job/node-accept-pull-request/82/.

misterdjules:
@joyent/node-coreteam The node-accept-pull-request Jenkins job keeps failing because `test/simple/test-tls-server-verify.js` takes more than 60 seconds to run on our Windows 2012 server Jenkins agent. It doesn't fail on that machine, it takes around 80 seconds and thus times out.

I made the decision to move forward with this and land this PR anyway since it was the only test failing, and I was able to make sure the test actually passes on two different Windows systems (2012 server and Windows 7).

As a result, landed in 84f1ab611480ab1ee1142b5fb0b39f57faa7dd3f and e1cc263ce8bfc14681cb277863f7d19a807bf551.

@orangemocha Would you mind taking a look at https://github.com/joyent/node/issues/25279?

Issue #17202, opened 2015-04-20
TLS Peer certificate exponent listed as "10001"


iamthechad:
The exponent should probably be stored in the peer certificate object as the number `65537` instead of the binary`10001` (unless there's a reason for the binary representation).

misterdjules:
@iamthechad `10001` is an hexadecimal representation, not a binary representation. The exponent is not stored in any specific human readable representation, `10001` is just the way [it's converted by node's native layer](https://github.com/joyent/node/blob/v0.12.2-release/src/node_crypto.cc#L1222-L1223).

What is the problem with using this specific representation?

iamthechad:
There's no problem per se. It's just inconsistent with tools that show the decimal value by default. (Although, if I can't tell binary and hex apart I have worse issues. Admittedly, other tools show it as `0x10001` which makes it more obvious.)

misterdjules:
@iamthechad I agree it is definitely confusing, and some clearer output would be beneficial. 

@joyent/node-coreteam I wonder if that would be considered an API change. Even though the API documentation doesn't mention the exponent, some users could be relying on its format to parse `getPeerCertificate`'s output.

I would suggest considering it's an API change and submitting a PR against master to have both the decimal and hexadecimal forms of the exponent displayed, or at least to prefix the hexadecimal form with `0x`.

Thoughts?

indutny:
This is certainly would be a breaking change. What do you think about just documenting it?

misterdjules:
Documenting in v0.10/v0.12 and making it clearer by changing the output in v0.13 sounds good to me.

indutny:
Ok, makes sense :)

misterdjules:
@iamthechad Do you have some time to make the changes as described [in my previous comment](https://github.com/joyent/node/issues/17202#issuecomment-98885573)?

iamthechad:
I'd be happy to take a stab at it. Where should I make the changes - against the `v0.10` and `v0.12` tags?

misterdjules:
@iamthechad For the documentation changes, that would be against v0.10. For the code changes, that would be against master. Thank you :+1:

iamthechad:
I've submitted doc changes. Are you also asking if I could make the code changes? I can certainly take a stab at it. Is there any documentation around for setting up a dev environment for Node? I'd love to be able to load things into a decent IDE so I can at least navigate easier.

misterdjules:
[...]

PR #25142, opened 2015-05-07
Clarify exponent encoding

iamthechad:
Refs #17202

### DIFF
--- a/doc/api/tls.markdown
+++ b/doc/api/tls.markdown
@@ -618,9 +618,13 @@ Example:
          O: 'node.js',
          OU: 'Test TLS Certificate',
          CN: 'localhost' },
+      modulus: '...hex representation of modulus...',
+      exponent: '10001',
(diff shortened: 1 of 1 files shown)


doc/api/tls.markdown
```diff
@@ -618,10 +618,14 @@ Example:
          O: 'node.js',
          OU: 'Test TLS Certificate',
          CN: 'localhost' },
+      modulus: '...hex representation of modulus...',
+      exponent: '10001',
```
rlidwka: Inline comment would be more clear:

```js
exponent: '10001', // 0x10001
```

jasnell:
@iamthechad ... can you squash these commits and make sure the change is consistent with what has landed in nodejs/node master? once squashed, I'll land

iamthechad:
I can squash, but don't we want this to reflect what's in 0.10 and not what's coming in a future release?

PR #25269, opened 2015-05-08
documentation: emitter.setMaxListeners(Infinity)

### LINKED ISSUE #22987: events: use `Infinity` for setMaxListeners instead of 0
Is there a sound reason for choosing `0` as the magic value that disables the max listeners? It's essentially using `0` as a special value to represent `unlimited` but JS already has a perfectly valid number to represent infinity: `Infinity`. It would simplify the memory-leak checking as the 0 special case would no longer need checked. Any number would naturally return true for `n < Infinity`.

jasonkarns:
Instead of recommending `0` as the magic value to set max listeners to unlimited, recommend `Infinity`.

closes #22987

### DIFF
--- a/doc/api/events.markdown
+++ b/doc/api/events.markdown
@@ -88,7 +88,7 @@ Returns emitter, so calls can be chained.
 By default EventEmitters will print a warning if more than 10 listeners are
 added for a particular event. This is a useful default which helps finding
 memory leaks. Obviously not all Emitters should be limited to 10. This function
-allows that to be increased. Set to zero for unlimited.
+allows that to be increased. Set to `Infinity` for unlimited.
 
 Returns emitter, so calls can be chained.
 


jasnell:
@misterdjules ... I'm inclined to close this as the documentation does accurately reflect the implementation and conventional use. Wanted to check with you before I did, however.

jasonkarns:
It reflects the implementation, but I argue that the implementation is poor. Intending `0` to represent Infinity when there is already the `Infinity` constant is nonsense. Of course, it doesn't make sense at this point to change the implementation; but changing the documentation to reflect a more meaningful value (that works functionally identical) can pave the way for the implementation to be improved in the future.

jasnell:
That may very well be the case. Currently, the right place to pursue a
change would be in the new repo. Perhaps open a new issue or PR against
http://github.com/nodejs/node?

jasonkarns:
closing in favor of https://github.com/nodejs/node/pull/2559

Issue #25271, opened 2015-05-08
Documentation error for TLS.connect()


iamthechad:
While working on code that let to issue #25270, I found that the docs for `tls.connect()` are incorrect.

It's not specified in the docs, but `secureContext` can be specified on the `options` object.

dnakamura:
Setting secureContext is not supported, see comments on #25270

iamthechad:
Setting `secureContext` is not supported, but setting `ciphers` is. The ability to specify `ciphers` in the `options` object is also missing from the docs. Can we repurpose this issue to address that, or should I open a new one?

mhdawson:
I think its ok to re-purpose this one.

iamthechad:
Fixed via #25325

Issue #25270, opened 2015-05-08
Cannot set ciphers for TLS client


iamthechad:
Setting ciphers for TLS client connections does not seem to work. See the below code.

    var tls = require('tls'),
        net = require('net');

    var options = {
        isServer: false
    };

    var context = tls.createSecureContext({ciphers: 'THIS_IS_NOT_A_CIPHER'});
    var tcpSocket = new net.Socket();
    tcpSocket.connect(443, "www.google.com");
    options.secureContext = context;
    options.socket = tcpSocket;
    var socket = tls.connect(options, function() {
        console.log(socket.getPeerCertificate());
        socket.end();
    });
    socket.setEncoding('utf8');
    socket.on('error', function(error) {
        console.log(error);
    });

This code prints out the peer certificate information for Google instead of raising an error, since `THIS_IS_NOT_A_CIPHER` is clearly not a valid cipher.

Even when I use valid ciphers, I get the same results. If I inspect the SSL handshake using Wireshark, I can see that the Client Hello is providing ~60 ciphers instead of just the one I specified.

I see the expected result in Wireshark if I use openssl directly:

`openssl s_client -cipher AES128-SHA -connect www.google.com:443 -no_ssl2 -no_ssl3`

Given that there is no error for an invalid cipher and the wrong value being used in the SSL handshake for a valid cipher, I would guess there's some issue in either parsing or setting the user-specified `ciphers`.

iamthechad:
After digging through the source code, I've found that I can specify `ciphers` directly when calling `tls.connect()`. This option isn't documented, which is why I went through the whole route of creating a `SecureContext` then a `net.Socket`. If I specify `ciphers` directly into `tls.connect()`, I get the expected behavior.

jasnell:
@iamthechad ... just for reference, which node.js version?

iamthechad:
@jasnell D'oh! This is on 0.12.x, which seems to have had a pretty major reworking of the TLS API.

dnakamura:
As far as I can tell, setting SecureContext is not supported.  It won't throw any errors, but its ignored all the same.

iamthechad:
Setting `secureContext` doesn't work, but setting `ciphers` does. The docs for `tls.connect()` should be updated to reflect this. Can we reuse this issue for that change, or should I open a new one?

iamthechad:
This was fixed via #25325

PR #25272, opened 2015-05-08
Add secureContext to tls.connect doc

iamthechad:
Refs #25271


iamthechad:
*sigh* I guess this is what happens when I update from upstream after submitting a PR. I don't know how to get rid of those extra commits short of closing this PR and opening a new one.

iamthechad:
Per comments on #25270, `secureContext` is not actually supported. This PR should be rejected.

PR #25278, opened 2015-05-12
test: fix test-process-active-wraps for IPv4

mhdawson:
test-process-active-wraps fails if the machine it runs on does not have
IPv6 enabled. This is due to the fact that Node first tries to use IPv6
and then falls back to using IPv4 if it is not enabled, as it can be
seen in lib/net.js lines 1100-1105. Even though binding to an IPv6
address fails, there is an extra handle created in the process.

This fixes the test by checking if IPv6 is supported and then binding
either to the IPv6 loopback if it is, or the IPv4 loopback if it is
not. By binding to a specific address we avoid the fall back
behavior and the test runs consistently regardless of whether IPv6
is enabled or not.


mhdawson:
This is a replacement for https://github.com/joyent/node/pull/9448

misterdjules:
Thank you @mhdawson! Two comments:

1. We'll need to change that part of the commit message:
  ```
  This fixes the test by adding different assertions for when only IPv4
  is available.
  ```
  to something that better reflects what the change actually does (listening on a different address     depending on the configuration of the host).

2. We also don't mention modified files in commit messages, this is provided by Git as part of the commit's metadata.

mhdawson:
Good point, I had just moved over the commit message from the earlier PR which no longer is accurate.  Updated.

misterdjules:
LGTM, thank you @mhdawson :+1:

mhdawson:
The accept job failed on windows, turns out the test only passed on linux because it was not running properly but still passed the checks.  Going to close this one and open new PR to delete the test.  See new PR for explanation of why

mhdawson:
Replaced with https://github.com/joyent/node/pull/25326

PR #25282, opened 2015-05-12
Python 3 compatibility for `configure`

blueyed:
This makes the configure script compatible with Python 3 - at least
there are no syntax errors anymore.

This PR also removes an unused variable in the script (`byteorder`).

It will still fail to run proper:

```
Traceback (most recent call last):
  File "./configure", line 16, in <module>
    from gyp.common import GetFlavor
  File "./tools/gyp/pylib/gyp/__init__.py", line 37
    print '%s:%s:%d:%s %s' % (mode.upper(), os.path.basename(ctx[0]),
                         ^
SyntaxError: invalid syntax
```

But the files in `tools` are out of scope of node, if I understood it correctly.

### DIFF
--- a/configure
+++ b/configure
@@ -354,12 +354,12 @@ def cc_macros():
                          stdout=subprocess.PIPE,
                          stderr=subprocess.PIPE)
   except OSError:
-    print '''Node.js configure error: No acceptable C compiler found!
+    print('''Node.js configure error: No acceptable C compiler found!
 
         Please make sure you have a C compiler installed on your system and/or
         consider adjusting the CC environment variable if you installed
         it in a non-standard prefix.
-        '''
+        ''')
     sys.exit()
 
   p.stdin.write('\n')
@@ -419,12 +419,12 @@ def is_arm_hard_float_abi():
         not '__VFP_FP__' in cc_macros()):
     return False
   else:
-    print '''Node.js configure error: Your version of GCC does not report
+    print('''Node.js configure error: Your version of GCC does not report
       the Floating-Point ABI to compile for your hardware
 
       Please manually specify which floating-point ABI to use with the
       --with-arm-float-abi option.
-      '''
+      ''')
     sys.exit()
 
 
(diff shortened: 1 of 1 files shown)


Fishrock123:
The problem is more that gyp and other dependancies require python 2(.?) anyways, so even if we fix the configure, the build isn't guaranteed to work.

blueyed:
Yes, I've noticed this.

But it's a start, isn't it? :)

jasnell:
Incremental improvement is good and using the print function form is just cleaner IMHO. However, even tho the second commit is minimal, I'd prefer splitting into separate PRs.

blueyed:
Ok, I'll split it.

misterdjules:
@jasnell @blueyed I haven't had the time to look at the changes, and I haven't used Python for a very long time, but we need to make sure that __any change is still compatible with Python 2.6__, [as it is listed as supported in the README](https://github.com/joyent/node/blob/master/README.md#to-build).

blueyed:
It should be compatible.

Is it covered by tests / CI?

I've stayed away from using `"{}".format(...)`, which might have been an issue - where Python 2.6 requires a positional arg always, i.e. `"{0}".format(...)`.

misterdjules:
@blueyed Yes, that would be covered by our current CI.

jasnell:
@blueyed @misterdjules ... how do you want to proceed on this one?

rud:
Looks like a good and simple change

jbergstroem:
@rud as long as gyp still requires python 2 I don't see a reason to land this change.

blueyed:
..but it also does no harm and will be useful in the future.

bnoordhuis:
Python 2-isms would sneak in before long, nullifying this PR.  I agree with @jbergstroem, there's no point in landing it.  I'll close.

rud:
Seems the nodejs build requires 2.x Python for now, so can probably be skipped. Does not change the fact that this PR cleans up the code a tiny bit.

Issue #25279, opened 2015-05-12
test/simple/test-tls-server-verify.js times out on our Windows 2012 Jenkins agent


misterdjules:
This currently prevents us from using the node-accept-pull-request job to land changes. I did some additional tests, and `test/simple/test-tls-server-verify.js` takes around 80 seconds to run on our Windows 2012 server Jenkins agent. It takes less time on my local Windows 7 VMWare VM (under 60 seconds).

@orangemocha Do you have some time to look into this?

jbergstroem:
See https://github.com/iojs/io.js/issues/1461 for more info.

misterdjules:
@jbergstroem Thank you for the heads up, very much appreciated :+1:

orangemocha:
Investigating... we'll probably have to mark it as flaky to unblock the CI.

orangemocha:
The test is slow on Windows because of two 1 second delays in the Windows version of openssl-cli.exe. We have reported the issue to openssl: http://rt.openssl.org/Ticket/Display.html?user=guest&pass=guest&id=3849

Now working on a workaround, modifying the test to let all those test cases run in parallel instead of serialized, so that it finishes quickly in spite of the openssl-cli delay.

orangemocha:
PR ready: https://github.com/joyent/node/pull/25368

orangemocha:
Fixed in https://github.com/joyent/node/commit/10349829f2f462e57cf7db2ef4c3a387d23e4826, which also unmarks the test as flaky.

PR #25284, opened 2015-05-12
test,win: disable test-tls-server-verify for CI

orangemocha:
test-tls-server-verify takes a lont time to execute and times
out on the Jenkins machines.

### DIFF
--- a/test/simple/simple.status
+++ b/test/simple/simple.status
@@ -6,6 +6,7 @@ test-cluster-basic                : PASS,FLAKY
 
 [$system==win32]
 test-timers-first-fire            : PASS,FLAKY
+test-tls-server-verify            : PASS,FLAKY
 
 [$system==linux]
 test-fs-readfile-error            : PASS,FLAKY


orangemocha:
cc @joyent/node-coreteam : let's disable this test to unblock the CI. I am investigating the issue with P-1: https://github.com/joyent/node/issues/25279

jasnell:
+1

indutny:
-1 from me. This test is passing like a charm on io.js.

orangemocha:
@indutny this PR doesn't actually disable the test in general, it just lets node-accept-pull-request ignore its failures.

orangemocha:
..that is, you need to pass --flaky-tests=dontcare to the test runner to see any difference. We only do that in node-accept-pull-request

indutny:
It is kind of the similar to disabling it, right? ;) what's the point if the test should be run by CI anyway?

orangemocha:
it's only disabling it in the context of the automated merge job. The alternative is to block accepting all other pull requests until this test is fixed.

orangemocha:
btw, the same problem seem to affect io.js: https://github.com/iojs/io.js/issues/1461. This test is slow only on Windows.

indutny:
@orangemocha this is absolutely correct, but nevertheless it does not seem to be a reason to disable it, at least for me.

orangemocha:
@indutny does that mean you think we should stop accepting pull requests until this test is fixed? It has been slow on Windows forever, it just so happens that now it's crossing the 60 seconds mark.

jasnell:
If the test is timing out because of a known system specific issue but we can verify that the test actually does pass independent of the automated run, then there should be no problem here. Omitting it from the automated pass should not be a problem. However, there should be a plan to address the issue and get the test re-enabled so we need to make sure that this is revisited.

Fishrock123:
So just a note: io.js prefers to mentally "ignore" errors rather than have a "green" ci run with ignored tests.

I have a feeling this error was fixed on our end by a recent jenkins upgrade, although I may be mistaken.

jasnell:
It's entirely possible. I've seen these kinds of timeout errors magically disappear across jenkins versions before. In this case, the jenkins job is not able to simple ignore the error that occurs.

misterdjules:
@jasnell @Fishrock123 See [my comment](https://github.com/joyent/node/issues/25279#issue-75412663):

> I did some additional tests, and test/simple/test-tls-server-verify.js takes around 80 seconds to run on our Windows 2012 server Jenkins agent.

Fishrock123:
Right, I'm not saying this isn't slow, but I think the entire CI-must-be-100%-"green" / not-"blocked" mantra may be unsavoury / detrimental.

misterdjules:
I just wanted to clarify that the issue is probably not due to Jenkins, since the test takes 80 seconds to complete when I log in with remote desktop on the machine and I run this single test manually.

misterdjules:
@Fishrick123 Yes, I agree that it's still not entirely clear what is more practical. However I think [what @jasnell said previously](https://github.com/joyent/node/pull/25284#issuecomment-101343887) would work well until this issue is fixed:

> If the test is timing out because of a known system specific issue but we can verify that the test actually > does pass independent of the automated run, then there should be no problem here. Omitting it from
> the automated pass should not be a problem. However, there should be a plan to address the issue
> and get the test re-enabled so we need to make sure that this is revisited.

With an emphasis on: _but we can verify that the test actually does pass independent of the automated run_.

orangemocha:
> Right, I'm not saying this isn't slow, but I think the entire CI-must-be-100%-"green" / not-"blocked" mantra may be unsavoury / detrimental.

@Fishrock123 , the point is not for the entire CI to be green. All other CI jobs except for node-accept-pull-request will show *red* when this test fails. Even node-accept-pull-request will not be green. It will be yellow, which gives us a clue that not everything is ok. 

The test is marked as "not ok", but there's a TODO next to it, which tells node-accept-pull-request to continue with the merge.
See here for an example http://jenkins.nodejs.org/job/node-test-commit-windows/74/DESTCPU=ia32,label=windows/tapTestReport/

orangemocha:
Anyway, this PR already landed after James' +1 because I thought it was urgent to unblock the CI. Since it ended up being a bit controversial, we can discuss it at the next core team meeting, and decide then if it should stay or be reverted.

orangemocha:
Landed in https://github.com/joyent/node/commit/0dbaee4a9ba3bfcf1d8d1d776eec59f6228196b3

misterdjules:
Thank you @orangemocha :+1: I think that with some extra care until the test is fixed, this is the best solution for now.

PR #25296, opened 2015-05-13
tls,crypto: revert recent cipher lists changes

misterdjules:
Revert "disable RC4, add --cipher-list command line switch"

Revert "disable RC4, add --cipher-list command line switch" and
"tls: make --enable-legacy-cipher-list=val less verbose"

This reverts commit f9291a94498c6f9d70a74a1e402f333383286120 and
b5737bb97775ff461266d28245df104dde0e6986.

There is still some work to be done to guarantee secure defaults and a
smooth upgrade path for v0.12.x users. Before this work is finished, we
want to be able to release new versions of v0.12.x. So instead of
waiting for these changes to be ready to ship, revert them and integrate
them when they're ready to be shipped.

Conflicts:
    src/node.cc


jasnell:
@misterdjules LGTM. I'll prepare the new PR that pulls these changes + the additional bits back in via a single commit once this immediate release is done.

misterdjules:
Landed in 72188d87d090d0f0f0bfac8ef069bfb01787e69d.

Issue #25295, opened 2015-05-13
Deadlock in V8 when CPU profiling is active


dmelikyan:
In current version of V8 3.28.73 a deadlock causes the process to hang and consume 100% CPU when CPU profiling is active and the application is serving requests. The reason is this line of code https://github.com/joyent/node/blob/v0.12.2/deps/v8/src/isolate.cc#L50, which leads to pthread_mutex_lock deadlock. It is removed in version 3.29.79.

To reproduce start the CPU profiling using V8 API (e.g. with v8-profiler) and run some load agains the server. My sample app would also do some requests to DBs to simulate real world apps. When it hangs, you can use gdb to verify it hangs because of pthread_mutex_lock.

Looks like the patching would be dangerous since the V8 commit fixing the issue has multiple changes and was actually nothing to do with this issue. So the solution is to upgrade V8.

Are there any plans to upgrade V8 in the 0.12 at some point? The issue makes it impossible to profile applications under load test and in production.

misterdjules:
It's the same issue as https://github.com/joyent/node/issues/14576. There is no plan to upgrade V8 in v0.12, as it would be a breaking change. We would need to find a proper way to fix this issue for V8 included in v0.12 and float a patch.

misterdjules:
Closing as a duplicate of #14576, thank you very much for filing an issue with a clear and detailed description, it is very much appreciated :+1:

misterdjules:
@mhdawson Does anyone on your team familiar with V8 internals has the time to investigate coming up with a floating patch for this?

PR #25309, opened 2015-05-14
V8: avoid deadlock when profiling is active

dmelikyan:
A deadlock happens when sampler initiated by SIGPROF tries to lock
the thread and the thread is already locked by the same thread. As
a result, other thread involved in sampling process hangs. The
patch adds a check for thread lock before continuing sampler
operation.

The fix has been tested on a sample app under load with and without
profiling turned on.

Fixes issue #14576 and specifically the duplicate issue #25295

### DIFF
--- a/deps/v8/src/isolate.h
+++ b/deps/v8/src/isolate.h
@@ -384,6 +384,11 @@ class Isolate {
  public:
   ~Isolate();
 
+  // ISSUE-14576: allow to access process_wide_mutex_ from sampler.cc
+  static base::Mutex* GetProcessWideMutexPointer() {
+    return process_wide_mutex_.Pointer();
+  }
+
   // A thread has a PerIsolateThreadData instance for each isolate that it has
   // entered. That instance is allocated when the isolate is initially entered
   // and reused on subsequent entries.
--- a/deps/v8/src/sampler.cc
+++ b/deps/v8/src/sampler.cc
@@ -342,6 +342,14 @@ void SignalHandler::HandleProfilerSignal(int signal, siginfo_t* info,
     // We require a fully initialized and entered isolate.
     return;
   }
+
+  // ISSUE-14576: To avoid deadlock, return if there is a thread lock
+  if (Isolate::GetProcessWideMutexPointer()->TryLock()) {
+    Isolate::GetProcessWideMutexPointer()->Unlock();
+  } else {
+    return;
+  }
+
   if (v8::Locker::IsActive() &&
       !isolate->thread_manager()->IsLockedByCurrentThread()) {
     return;
--- a/test/simple/test-regress-GH-14576.js
+++ b/test/simple/test-regress-GH-14576.js
@@ -1,0 +1,57 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+/*
+ * Tests for sampler deadlock regression. 
+ * Issue https://github.com/joyent/node/issues/14576
+ */
+
+var assert = require('assert');
+var cp = require('child_process');
+
+var child = undefined;
+
+if (process.platform !== 'win32') {
+  process.on('SIGTERM', function() {
+    if(child) {
+      process.kill(child.pid, 'SIGKILL');
+    }
+
+    process.exit();
+  });
+}
+
+var testScript = "var i = 0; function r() { if(++i > 25) return; " +
+                 "setTimeout(r, 1); }; r();"
+var nodeCmd = process.execPath + 
+              ' --prof --nologfile_per_isolate' + 
+              ' -e "' + testScript + '"';
+var runs = 0;
+
+function runTestScript() {
+  child = cp.exec(nodeCmd, function(err, stdout, stderr) {
+    if (++runs > 50) return;
+
+    runTestScript();
+  });
+}
+
+runTestScript();


JacksonTian:
The issue was reported multi time recently. cc @misterdjules

mhdawson:
FYI @ tunniclm is validating this change against issues we've seen as well

tunniclm:
**Background**
I ran across this hang as well while testing some code using the CpuProfiler API. My test would hang about 1 in 25 runs. To enable continued testing of that code, I made some rough changes in an attempt to make my local V8 stable. I explored two different approaches to the one in this pr:

1. Make the process_wide_mutex_ recursive -- this did not resolve the hang but did narrow the window significantly (it would hang in about 1 in 300 runs); the problem would now only happen when pthread_mutex_lock() was interrupted at a critical point
2. Block the SIGPROF signal using pthread_sigmask() while acquiring and releasing the process_wide_mutex_ -- this worked (no hangs in 3000 runs)

In preparing to share this with the community I found @dmelikyan 's work. I wanted to check the fixes with my tests just in case they could find a similar window as (1).

**Result**
I have run my testcases 3000 times on Linux x86 and Linux PPC64 and there were no hangs, so this fix looks good to me. :)

deps/v8/src/sampler.cc
```diff
@@ -342,6 +342,15 @@ void SignalHandler::HandleProfilerSignal(int signal, siginfo_t* info,
     // We require a fully initialized and entered isolate.
     return;
   }
+
+  // ISSUE-14576: To avoid deadlock, return if there is a thread lock
+  if (Isolate::GetProcessWideMutex().Pointer()->TryLock()) {
+    Isolate::GetProcessWideMutex().Pointer()->Unlock();
+  }
```
misterdjules: Style: I'm not familiar with V8's code base's style guidelines, but I think this line should be `} else {` (curly braces on the same line as the `else` keyword).

deps/v8/src/isolate.h
```diff
@@ -384,6 +384,11 @@ class Isolate {
  public:
   ~Isolate();
 
+  // ISSUE-14576: allow to access process_wide_mutex_ from sampler.cc
+  static base::LazyMutex GetProcessWideMutex() {
+    return process_wide_mutex_;
```
misterdjules: Wouldn't that make a copy of `process_wide_mutex_` every time `GetProcessWideMutex()` is called, and thus `Mutex`'s constructor would be called, which in turn would call `InitializeNativeHandle` and finally `pthread_mutex_init`? If so, is that a problem and should it return a reference instead?

misterdjules:
@tunniclm Thank you for the tests report!

@dmelikyan Besides my two comments, I think it's worth it to try to come up with a regression test, is this something you have some time to work on? Thank you very much again for your work and your patience!

deps/v8/src/isolate.h
misterdjules: Nevermind, I was just not understanding the `LazyStaticInstance` type correctly. That looks good to me.

dmelikyan:
@misterdjules I've added a regression test. Not sure if "simple" if is the right place for it though. Also fixed the style problem.

dmelikyan:
@misterdjules Switched to using mutex pointer instead of object just in case related to your comments.

test/simple/test-regress-GH-14576.js
```diff
@@ -0,0 +1,47 @@
...
+ * Tests for sampler deadlock regression. 
+ * Issue https://github.com/joyent/node/issues/14576
+ */
+
+var assert = require('assert');
+var cp = require('child_process');
+
+var child = undefined;
+
+var deadlockTimer = setTimeout(function() {
```
misterdjules: The tests suite driver [has a default timeout of 60 seconds](https://github.com/joyent/node/blob/master/tools/test.py#L1208-L1209), and we usually don't setup timeouts in the tests themselves, so I would suggest removing that mechanism. If the test times out, it will eventually be killed by the tests suite driver and the failure will be reported as a timeout.

test/simple/test-regress-GH-14576.js
```diff
@@ -0,0 +1,47 @@
...
+  if(child) {
+    process.kill(child.pid, 'SIGKILL');
+  }
+
+  assert(false, 'Node.js process hung');
+}, 10000);
+
+var testScript = "var i = 0; function r() { if(i++ > 2000) return; setTimeout(r, 1); }; r();";
```
misterdjules: If I understand correctly, this is not sufficient as a regression test. The deadlock problem does not happen all the time and we would need to invoke node with `--prof` a certain number of times. I think 100 times is a safe assumption, as it's what I used to reproduce this issue and I was able to reproduce it within 100 runs all the time.

I also just validated that:
```
var i = 0; function r() { if(i++ > 2000) return; setTimeout(r, 1); }; r();
```
doesn't deadlock all the time with node v0.12.4:
```
➜  v0.12 git:(v0.12.4-release) ✗ ./node --version                                                                      
v0.12.4
➜  v0.12 git:(v0.12.4-release) ✗ time ./node -e 'var i = 0; function r() { if(i++ > 2000) return; setTimeout(r, 1); }; r();'
./node -e   0.12s user 0.07s system 7% cpu 2.580 total
➜  v0.12 git:(v0.12.4-release) ✗
```

deps/v8/src/sampler.cc
```diff
@@ -342,6 +342,14 @@ void SignalHandler::HandleProfilerSignal(int signal, siginfo_t* info,
     // We require a fully initialized and entered isolate.
     return;
   }
+
+  // ISSUE-14576: To avoid deadlock, return if there is a thread lock
+  if (Isolate::GetProcessWideMutexPointer()->TryLock()) {
```
misterdjules: Should we check for a NULL pointer here? I guess that it's very unlikely to happen, but it seems safer to have the check than not have it.

misterdjules:
@dmelikyan `test/simple` is the right place if the test runs very quickly (under a couple seconds) and doesn't depend on any external dependency (which is the case for your test).

deps/v8/src/sampler.cc
misterdjules: @dmelikyan @tunniclm Also, I'm still not sure I properly this fix. What is already locking the process wide mutex in the profiler thread when `isolate->thread_manager()->IsLockedByCurrentThread()` is called? Also, why is `Isolate::GetProcessWideMutexPointer()->TryLock()` returning true in case the process wide mutex is already locked? Shouldn't it return false since `man pthread_mutex_try_lock` says that `If the mutex is already locked, pthread_mutex_trylock() will not block waiting for the mutex, but will return an error condition`?

dmelikyan: TryLock returns false if it wasn't able to acquire lock. In this case we should return, otherwise a deadlock may happen. This is the point of the fix. The code on the next line, IsLockedByCurrentThread(), which ends up calling EnsureInitialized(), does lock, which results in a deadlock (as described in the pull request comment) and so I made sure it doesn't get there. Starting V8 version 3.29.79 the EnsureInitialized() call was removed from GetCurrentThreadId(), which fixes the deadlock problem automatically and this fix is not required anymore.

Concerning NULL pointer, it shouldn't be the case, otherwise IsLockedByCurrentThread() would also throw NULL pointer.

test/simple/test-regress-GH-14576.js
dmelikyan: The spawned and hanging node processes aren't killed by the test suite, because they do not die when parent process dies (at least on Mac). They should be killed explicitly. I've changed the way it's done so that it is compliant with the timeout option. Also running the test script multiple times now.

deps/v8/src/sampler.cc
misterdjules: @dmelikyan Right, I understand the rationale behind your fix now, thank you for the clarification. 

What is already locking the process wide mutex in the profiler thread when `SignalHandler::HandleProfilerSignal` is called? Possibly related question: is `SIGPROF` blocked while running `SignalHandler::HandleProfilerSignal`? Otherwise I imagine there would still be a race condition.

test/simple/test-regress-GH-14576.js
misterdjules: @dmelikyan That's great, thank you!

test/simple/test-regress-GH-14576.js
```diff
@@ -0,0 +1,53 @@
...
+  process.on('SIGTERM', function() {
+    if(child) {
+      process.kill(child.pid, 'SIGKILL');
+    }
+
+    process.exit();
+  });
+}
+
+var testScript = "var i = 0; function r() { if(++i > 25) return; setTimeout(r, 1); }; r();";
```
misterdjules: Style: wrap at 80 columns by concatenating the string.

test/simple/test-regress-GH-14576.js
```diff
@@ -0,0 +1,53 @@
...
+var runs = 0;
+
+function runTestScript() {
+	child = cp.exec(process.execPath + ' --prof -e "' + testScript + '"', function(err, stdout, stderr) {
```
misterdjules: Style: wrap at 80 columns and use two spaces indentation.

test/simple/test-regress-GH-14576.js
```diff
@@ -0,0 +1,53 @@
...
+ * Tests for sampler deadlock regression. 
+ * Issue https://github.com/joyent/node/issues/14576
+ */
+
+var assert = require('assert');
+var cp = require('child_process');
+
+var child = undefined;
+
+if(process.platform !== 'win32') {
```
misterdjules: Style, space between `if` and opening parenthesis `(`.

misterdjules:
@dmelikyan Thank you very much for making the changes according to my comments :+1: I think we're making some good progress to get this landed soon. I just have [one remaining question](https://github.com/joyent/node/pull/25309#discussion_r32047071) and some styling comments.

deps/v8/src/sampler.cc
dmelikyan: There are different places where the lock is acquired by the same thread, I've seen different stacks with gdb.

Not sure how/if SIGPROF is blocked, I think not, because it is fired once every millisecond (every 10ms on Android).

misterdjules: @dmelikyan According to different man pages (like http://man7.org/linux/man-pages/man2/sigaction.2.html, but also on OSX, I haven't looked on other platforms) and other sources, including "Advanced Programming In the Unix Environment" from Richard Stevens, it seems that `sigaction`:

> guarantee[s] that whenever we are processing a given signal, another occurrence of that same signal is blocked until we're finished processing the first occurrence. 

(that's an excerpt from the book mentioned above).

So I would think that your approach is safe.

However @tunniclm mentioned in [one of his comments](https://github.com/joyent/node/pull/25309#issuecomment-109214872) that he investigated:

> Block[ing] the SIGPROF signal using pthread_sigmask() while acquiring and releasing the process_wide_mutex_

which suggests that `SIGPROF` is not blocked when `SignalHandler::HandleProfilerSignal` runs. @tunniclm Are we missing something?

dmelikyan:
@misterdjules See latest commit for style changes. Other thing, running with --prof creates files, e.g. isolate-0x101004c00-v8.log in the node directory. Should we worry about it? I'll try to find a way to remove them or disable if possible at all.

misterdjules:
@dmelikyan That's a good question, I think it's worth it to investigate if we can remove them when the test completes.

deps/v8/src/sampler.cc
tunniclm: @misterdjules This was instead of, rather than in addition to the pthread_mutex_trylock() approach. Blocking was not added inside the signal handler, but rather around the code in the rest of the system that called pthread_mutex_lock() and pthread_mutex_unlock() (on the process_wide_mutex_) so that when the signal occurs it it will never interrupt those calls to ...lock()/..unlock().

dmelikyan: By blocking SIGPROF we're altering V8's original logic to some extent. I wouldn't do it and keep the patch only fixing the actual deadlock problem.

misterdjules: @tunniclm Thanks for the clarification, that makes sense :+1:

@dmelikyan Agreed, thanks!

misterdjules:
@dmelikyan @tunniclm So I would think that this should be ready to land when after we determine what can be done with the log files generated by the test. Thank you for the great work so far!

dmelikyan:
@misterdjules There is a V8 option for disabling per isolate log files. See my last commit. No log files are generated now.

misterdjules:
@dmelikyan Great! Now let's just squash the commits into one and we'll be ready to land it.

dmelikyan:
@misterdjules squashed.

misterdjules:
LGTM.

misterdjules:
Landed in b81a643f9ae341b0c23cecc54daccdc8d7bc746a, thank you all :+1:

misterdjules:
Added a reference to this change to [the list of floating patches on top of V8 in v0.12](https://github.com/joyent/node/wiki/V8-upgrade-process#v012x).

PR #25325, opened 2015-05-14
Document ciphers as part of tls.connect() options

iamthechad:
Refs #25270,#25271

### DIFF
--- a/doc/api/tls.markdown
+++ b/doc/api/tls.markdown
@@ -257,7 +257,7 @@ automatically set as a listener for the [secureConnection][] event.  The
     on the format.
 
     `ECDHE-RSA-AES128-SHA256`, `DHE-RSA-AES128-SHA256` and
-    `AES128-GCM-SHA256` are TLS v1.2 ciphers and used when node.js is
+    `AES128-GCM-SHA256` are TLS v1.2 ciphers and used when Node.js is
     linked against OpenSSL 1.0.1 or newer, such as the bundled version
     of OpenSSL.  Note that it is still possible for a TLS v1.2 client
     to negotiate a weaker cipher unless `honorCipherOrder` is enabled.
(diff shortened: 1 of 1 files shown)


dnakamura:
Might also be worth mentioning mention crypto.getCiphers() and/or http://www.openssl.org/docs/apps/ciphers.html#CIPHER-LIST-FORMAT.

iamthechad:
I added a link to the `getCiphers` function. There was already a link to the cipher list format.

I don't think it's worth adding anything about `honorCipherOrder` because this setting doesn't make sense when connecting as a client.

mhdawson:
Thanks for taking the time to submit this PR as adding the doc will help people trying to figure out how you can pass ciphers.

I still want to take one last look after generating the doc with the change applied which I plan to do next week (its a long weekend here) but a few comments you can either fix up now or wait until I've done that:

1) node.js should be Node.js.

2) The first line of the commit comment should start with "doc: " and the total length should be under 50 chars.

3) The commits will need to be squashed into a single commit. You can do this with a git rebase -i HEAD~2 and selecting squash for one of the two, followed by a push with a --force

mhdawson:
Looks good except that as a minor nit there is extra white space on 2 of the blank lines (I'm guessing an extra tab)  above:

  - `ciphers`: A string describing the ciphers to use or exclude.

and

       The full list of available ciphers can be obtained via [tls.getCiphers][].

Can you remove the extra white space, squash again to one commit and then re-push and then I"ll pull in

I also checked and passing the ciphers in the options for tls.connect is used in at least one of the existing test cases (simple/test-tls-honorcipherorder.js)

mhdawson:
Landed as a294aeef0315817fa230bb4e048564d2aea5cdf7

Please close the associated issues if you believe they are now resolved

Issue #8061, opened 2014-08-02
Failed test: simple/test-process-active-wraps.js 


jbergstroem:
For some reason, my system (node 37c2a528336b41b31abe4826b308071896e42423)  seems to count two ticks after starting the server:
```bash
./node test/simple/test-process-active-wraps.js 

assert.js:98
  throw new assert.AssertionError({
        ^
AssertionError: 2 == 1
    at expect (/home/jbergstroem/node/test/simple/test-process-active-wraps.js:28:10)
    at /home/jbergstroem/node/test/simple/test-process-active-wraps.js:37:3
    at Object.<anonymous> (/home/jbergstroem/node/test/simple/test-process-active-wraps.js:41:3)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:124:16)
    at node.js:811:3
```

built with: `./configure --prefix=/usr --shared-libuv --shared-http-parser --shared-zlib --shared-openssl --without-dtrace`, External libraries match the same versions as in deps (besides openssl since that is patched in deps).

hefangshi:
got same test failed in v0.12, any information?

mhdawson:
Just found this see https://github.com/joyent/node/pull/25326

mhdawson:
I'll close this once #25236 goes in

mhdawson:
25236 has landed in V0.12 head, closing

PR #25326, opened 2015-05-14
test: delete simple/test-process-active-wraps

mhdawson:
This test currently fails when run on machines without
IPv6 enabled. Futher it was delete in io.js under
3143d732f6efd82da76e9c53ad192ac14071bf70 as the test
was known to have problems across platforms and
releases.

I could not cherry-pick the original commit across because 
the test had been moved in io.js

The existing test was hard to understand so I wrote
a new version but then found exactly what was
reported in the io.js pull request.  Behaviour varies
across platforms such that writing a solid test would
either be infeasible or test so little that
it does not seem to make sense to keep it.


mhdawson:
Also mentioned in this open issue: https://github.com/joyent/node/issues/8061

cjihrig:
LGTM

mhdawson:
Landed as 97d4706e8163631b8bc5ae64a33f56e2c489e9e3

PR #25328, opened 2015-05-15
deps: revert v8 Array.prototype.values() removal

### LINKED ISSUE #25324: Breaking v8 change in v0.12.3
https://twitter.com/ljharb/status/598917863433342976

`Array.prototype.values` no longer exists in `v0.12.3`; it exists in `v0.12.2`. This will break any code relying on it.

cjihrig:
The Node 0.12 line was initially released with a version of v8 that included `Array.prototype.values()`. In https://github.com/joyent/node/pull/18206, v8 was updated to a version that dropped support for `values()`. https://codereview.chromium.org/647703003 removed this method because it causes problems with some versions of Outlook Web Access. This commit reverts the removal of `Array.prototype.values()`.

Closes #25324.


misterdjules:
I would suggest maybe adding a regression test similar to:
```
/*
 * Regression test for https://github.com/joyent/node/issues/25324.
 */
 
var array = ['foo', 'bar', 'baz'];
var iterator = array.values();

var nbValues = array.length;
var nbIterated = 0;

for (var letter of iterator) {
  ++nbIterated;
}

assert(nbIterated === nbValues);
```
so that we avoid breaking this in the future in v0.12.

Otherwise LGTM. 

@ljharb I imagine you found out about https://github.com/joyent/node/issues/25324 by running some code from es6-shim? If so, could you please run that code against that PR and check if that fixes it for you?

Thank you!

cjihrig:
@misterdjules added a test.

ljharb:
Yes, I clone the repo and run `npm run test-native`. I'm not sure how to install node easily from a non-release though.

deps/v8/test/mjsunit/es6/typed-array-iterator.js
```diff
@@ -21,9 +21,9 @@ function TestTypedArrayPrototype(constructor) {
   assertFalse(constructor.prototype.propertyIsEnumerable(Symbol.iterator));
 
   assertEquals(Array.prototype.entries, constructor.prototype.entries);
-  assertEquals(Array.prototype[Symbol.iterator], constructor.prototype.values);
+  assertEquals(Array.prototype.values, constructor.prototype.values);
```
ljharb: These tests should still be here.

cjihrig: I simply did a revert on the commit that removed `Array.prototype.values()`. What the v8 team does with their tests is their own business :-)

misterdjules:
@cjihrig Thank you! Do you have some time to run [the tests that @ljharb mentioned](https://github.com/joyent/node/pull/25328#issuecomment-102246946) and make sure they pass with these changes?

I would also put the node test in a separate commit, but that's a minor nit, I'm ok with it like that.

cjihrig:
Ran the tests. None of the failures appeared to be related to `Array.prototype.values()`.

cjihrig:
@misterdjules split the Node test into a separate commit as you requested.

DomT4:
What's the status of this? Is 0.12.4 likely to land shortly after this is merged?

ljharb:
+1, I think this needs to go out ASAP since it was a breaking change in a patch version

misterdjules:
@DomT4 @ljharb Yes, current plan is to ship this as v0.12.4 as soon as recent issues with Windows XP are resolved.

The relevant Windows XP issues are https://github.com/joyent/node/issues/25347 and https://github.com/joyent/node/issues/25348. The current status on these issues is that we managed to reproduce the problem, we tested that the suggested fix fixes the problem, we just want to go the extra mile and understand what caused the regression before we close these issues.

Hopefully that won't be too long.

misterdjules:
LGTM. Thank you @cjihrig :+1:

misterdjules:
Landed in 6157697bd5bdbf58b19b9d3177249a7bbb467f79 and e23450fa8128b42c958761d39fb0256024339bba. Thank you!

PR #25329, opened 2015-05-15
configure changes for portability and better interop

ensonic:
We are using these two changes for the chromeos packages of nodejs to make the build work in a cross compilation environment with custom toolchains.

### DIFF
--- a/configure
+++ b/configure
@@ -466,6 +466,13 @@ def host_arch_win():
   return matchup.get(arch, 'ia32')
 
 
+def int_lenient(n):
+  try:
+    return int(n)
+  except ValueError:
+    return 0
+
+
 def compiler_version():
   try:
     proc = subprocess.Popen(shlex.split(CC) + ['--version'],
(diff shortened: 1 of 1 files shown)


jasnell:
@ensonic ... at this point, because of the transition over to http://github.com/nodejs/node, it would likely be best to retarget this at nodejs/node or open a discussion at nodejs/build.

ensonic:
Hmm, I don't get it. This PR is for nodejs/node. Atleast when I opened the PR.

jasnell:
The PR was opened originally against the joyent/node repo, which is now archived. We would need this opened against the new nodejs/node repo

ensonic:
+jasnell, not your fault, but honestly this sucks. This is a not a great way to attract new contributors. The patches don't apply to upstream anymore and from cursory checking, they won't be needed anymore. At least I can't see any compiler version checks in the handrolled (yeah) configure script. And pkgconfig is resolved via the envvar too.

PR #25334, opened 2015-05-15
path: Remove dead code in favor of unit tests

nwoltman:
This PR removes dead code paths that are created by assertions that will never trigger. They may only trigger if either the `splitDeviceRe` or `splitPathRe` regular expressions or the `[win32|posix]SplitPath` functions are modified. If at some point they are modified (and for some reason the author doesn't update the code that uses those regexes/functions), current unit tests will catch most of the resulting errors and this PR adds extra tests to catch the remaining errors.

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -90,7 +90,7 @@ function win32SplitPath(filename) {
   // Separate device+slash from tail
   var result = splitDeviceRe.exec(filename),
       device = (result[1] || '') + (result[2] || ''),
-      tail = result[3] || '';
+      tail = result[3];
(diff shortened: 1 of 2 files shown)


jasnell:
Builds and tests pass on macosx. LGTM except for the linting errors.

woollybogger:
Fixed linting errors.

woollybogger:
This was landed in nodejs/node@2db57bd so this PR can probably be closed.

jasnell:
Awesome! Thanks!

PR #25337, opened 2015-05-16
net: setKeepAlive now throws an exception on error

### LINKED ISSUE #21080: setKeepAlive silently ignores invalid values
While testing #21079, I also discovered that TCP sockets' setKeepAlive() function silently ignores invalid values.  I tried passing 5000ms for the delay, which gets translated to 5s.  On my system, the minimum is 10s, and the setsockopt returns with EINVAL, but Node neither throws an exception nor emits an Error on the socket.  Here's the test program:

``` javascript
var sock = require('net').connect({
    'host': '172.25.10.4',
    'port': 12345
});
sock.on('connect', function () {
        console.log('setKeepAlive returns: ', sock.setKeepAlive(true, 5000));
});
```

and here's the output while tracing setsockopt():

```
$ truss -t setsockopt node tcpkatest1.js 
/1:     setsockopt(11, SOL_SOCKET, SO_KEEPALIVE, 0xFFFFFD7FFFDF8F70, 4, SOV_DEFAULT) = 0
/1:     setsockopt(11, tcp, 0x22, 0xFFFFFD7FFFDF8F7C, 4, SOV_DEFAULT) Err#22 EINVAL
setKeepAlive returns:  undefined
```

(On this system, 0x22 is TCP_KEEPIDLE.  I checked that the value being passed in was "5", and that's less than the minimum allowed on this system.)

emmapowers:
socket.setKeepAlive now throws an error if the operation fails.
For example setting the keep alive time too small, or calling
the function on a closed socket.

Fixes #21080

### DIFF
--- a/lib/net.js
+++ b/lib/net.js
@@ -359,8 +359,11 @@ Socket.prototype.setNoDelay = function(enable) {
 
 
 Socket.prototype.setKeepAlive = function(setting, msecs) {
-  if (this._handle && this._handle.setKeepAlive)
-    this._handle.setKeepAlive(setting, ~~(msecs / 1000));
+  if (this._handle && this._handle.setKeepAlive) {
+    var err = this._handle.setKeepAlive(setting, ~~(msecs / 1000));
+    if (err)
+      throw errnoException(err, 'setKeepAlive');
+  }
 };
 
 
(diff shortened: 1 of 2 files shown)


jasnell:
cc @joyent/node-coreteam ... any API compatibility issue here?

brendanashworth:
I'm aware that @misterdjules [advocated for throwing an exception](https://github.com/joyent/node/issues/21080#issuecomment-99638726), but I'd like to ask for extra review (with convergence in mind) on that decision because in io.js, `socket.setKeepAlive()` can be deferred to other ticks. [take a look here](https://github.com/nodejs/io.js/commit/85d99830096a48b7d50cc3b0e5c3fba4172c2d02)

It may be a programmer error, but is throwing on a later event loop cycle still the best way to go?

jasnell:
@joyent/node-tsc ... ping

PR #25342, opened 2015-05-16
module: fix column offset on first line stack trace

### LINKED ISSUE #9445: Stack trace column number incorrect if error is thrown on line 1
It seems that node internally prepends some code to wrap a script into a module. However, it fails to correct for this in the stack traces of any errors that occur on line 1 of a script.  This manifests itself e.g. in uglified files where EVERYTHING is on line 1 of the script. This renders sourcemaps useless. 

Example:
```javascript
throw new Error("bla");
```

Output (note the column number of the first stack frame is larger than the line in the code):
```
c:\issue.js:1
(function (exports, require, module, __filename, __dirname) { throw new Error(
                                                                    ^
Error: bla
    at Object.<anonymous> (c:\issue.js:1:69)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:935:3
```

emmapowers:
When modules are loaded by require(), they are wrapped in
a function. This added text causes errors on the first line
of a file to be offset by the length of the function
declaration. For example, an error on the first character
of a module would appear as if it were on column 63.

To fix this problem I added line and column offset support
to the ContextifyScript object and added options to the
runInContext(and friends) function. I then added a column
offset to module._compile that compensates for the function
wrapper.

Fixes #9445


src/node_contextify.cc
```diff
@@ -658,6 +660,51 @@ class ContextifyScript : public BaseObject {
...
+                                      const int i) {
+    Local<Integer> defaultLineOffset = Integer::New(args.GetIsolate(), 0);
+
+    if (args[i]->IsUndefined()) {
+      return defaultLineOffset;
+    }
+    if (args[i]->IsInt32()) {
+      return args[i].As<Integer>();
+    }
+    if (!args[i]->IsObject()) {
```
cjihrig: I think this is the only check you actually need here. The `IsUndefined()` and `IsInt32()` checks above can go. Same comment applies for your `GetColumnOffsetArg()` function.

test/simple/test-module-loading.js
```diff
@@ -306,3 +306,18 @@ process.on('exit', function() {
 // #1440 Loading files with a byte order marker.
 assert.equal(42, require('../fixtures/utf8-bom.js'));
 assert.equal(42, require('../fixtures/utf8-bom.json'));
+
+// #9445 Error on first line of a module have the correct column number.
```
cjihrig: This can just be (might need to split this across lines if it's more than 80 characters):

`// Error on the first line of a module should have the correct line and column number`

test/simple/test-module-loading.js
```diff
@@ -306,3 +306,18 @@ process.on('exit', function() {
...
+var gh9445Exception;
```
cjihrig: Can you just name this `err`?

test/simple/test-module-loading.js
```diff
@@ -306,3 +306,18 @@ process.on('exit', function() {
...
+try {
+  var err_js = require('../fixtures/module-err.js');
```
cjihrig: No need to assign this to a variable since you don't reference it anywhere.

test/simple/test-module-loading.js
```diff
@@ -306,3 +306,18 @@ process.on('exit', function() {
...
+}
+catch (e) {
```
cjihrig: Can you move this to the previous line like `} catch (e) {`

cjihrig:
@joyent/node-coreteam what do you think of this? It does seem to solve the problem. If we take this, I'm not sure we should necessarily document it as part of the public API.

Also, if we go this route, an alternative would be to change the module wrapper from:

`'(function (exports, require, module, __filename, __dirname) { '`

to

`'(function (exports, require, module, __filename, __dirname) {\n'`

And just use a `lineOffset` of -1. Would that work on Windows too? It honestly scares me a little to mess with the module wrapper though.

lib/module.js
```diff
@@ -441,8 +441,10 @@ Module.prototype._compile = function(content, filename) {
 
   // create wrapper function
   var wrapper = Module.wrap(content);
+  var wrapperOffset = Module.wrapper[0].length;
```
cjihrig: This can be pulled out of the function to avoid calculating every time.

lib/module.js
```diff
@@ -441,8 +441,10 @@ Module.prototype._compile = function(content, filename) {
...
 
-  var compiledWrapper = runInThisContext(wrapper, { filename: filename });
+  var compiledWrapper = runInThisContext(wrapper,
+      { filename: filename, columnOffset: -wrapperOffset });
```
cjihrig: Does this pass `make lint`?

test/simple/test-vm-context.js
```diff
@@ -61,6 +61,22 @@ catch (e) {
 assert.ok(gh1140Exception,
           'expected exception from runInContext signature test');
 
+// Issue GH-9445:
+console.error('test runInContext offset');
+var gh9445Exception;
```
cjihrig: Just use `err`.

test/simple/test-vm-context.js
```diff
@@ -61,6 +61,22 @@ catch (e) {
...
+try {
+  vm.runInContext('throw new Error()', context, { filename: 'expected-filename.js',
```
cjihrig: Leave the opening curly brace, `filename` should go on the next line.

test/simple/test-vm-context.js
```diff
@@ -61,6 +61,22 @@ catch (e) {
...
 
+// Issue GH-9445:
+console.error('test runInContext offset');
+var gh9445Exception;
+try {
+  vm.runInContext('throw new Error()', context, { filename: 'expected-filename.js', 
+                                                  lineOffset: 32,
+                                                  columnOffset: 123});
+}
+catch (e) {
```
cjihrig: Move this up to the previous line.

jasnell:
@brendan0powers ... how would you like to proceed on this? @cjihrig raised a number of issues that would need to be addressed. If you'd like to pursue, updating the PR and re-opening it against http://github.com/nodejs/node would likely be the best approach.

brendanashworth:
It looks like this is being followed up in https://github.com/nodejs/node/pull/2867. I'll close this for now, thanks!

Issue #25353, opened 2015-05-19
url.parse() fails if auth contain a colon


ben-page:
If the user or the password component of auth contains colons, url.parse() fails.
```javascript
var url = require('url');

var user = encodeURIComponent('us:er');
var password = encodeURIComponent('pass:word');
var uri = 'http://' + user + ':' + password + '@localhost/';
console.log(uri); // http://us%3Aer:pass%3Aword@localhost/

var parsed = url.parse(uri);
console.log(parsed.auth); // us:er:pass:word
```

The output of parse() is not understood by format().
```javascript
console.log(url.format(parsed)); // http://us:er%3Apass%3Aword@localhost/
```

I think auth needs to be split into two fields (username and password).

ben-page:
Browsers handled it this way:
```javascript
var parser = document.createElement('a');
parser.href = "http://us%3Aer:pass%3Aword@example.com:3000/path%3Ename/?search=some%20phrase#hash";

console.log(parser.hash); // #hash
console.log(parser.host); // example.com:3000
console.log(parser.hostname); // example.com
console.log(parser.href); // http://us%3Aer:pass%3Aword@example.com:3000/path%3Ename/?search=some%20phrase#hash
console.log(parser.origin); // http://example.com:3000
console.log(parser.password); // pass%3Aword
console.log(parser.pathname); // /path%3Ename/
console.log(parser.port); // 3000
console.log(parser.protocol); // http:
console.log(parser.search); // ?search=some%20phrase
console.log(parser.username); // us%3Aer
```
I think is much better than the way url.parse() currently work. Though, I don't see any harm in decoding the components after the URL has been parsed. It might be nice to have a argument to disable automatic URL decoding.

dnakamura:
Just to be clear, this is not so much url.parse() failing (its still technically doing what its supposed to),  but rather a limitation of the api (it's impossible to separate the username/password from the auth field). 
PS. that's just a colon, not a semicolon

jasnell:
Hmm.. there appears to be a couple of failures here. Given the test case, if you look at the `parsed` object, you'll see:

```
> parsed
{ protocol: 'http:',
  slashes: true,
  auth: 'us:er:pass:word',
  host: 'localhost',
  port: null,
  hostname: 'localhost',
  hash: null,
  search: null,
  query: null,
  pathname: '/',
  path: '/',
  href: 'http://us:er%3Apass%3Aword@localhost/' }
> 
```

[...]

PR #25359, opened 2015-05-20
url: add username and password fields to parse() output

ben-page:
url.parse() fails if the auth field contains a colon(:) or at-sign(@).

The solution is to add two new fields, username and password. Additionally, url.format() is updated to prefer these new fields over the auth field. The tests and docs have been updated accordingly. This should not be breaking as the behavior of the auth field was not modified.

Here's the issue: https://github.com/joyent/node/issues/25353

### DIFF
--- a/lib/url.js
+++ b/lib/url.js
@@ -42,6 +42,9 @@ function Url() {
   this.pathname = null;
   this.path = null;
   this.href = null;
+  this.href = null;
+  this.username = null;
+  this.password = null;
 }
 
 // Reference: RFC 3986, RFC 1808, RFC 2396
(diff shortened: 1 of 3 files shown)


ben-page:
https://github.com/nodejs/io.js/issues/1802

lib/url.js
```diff
@@ -42,6 +42,9 @@ function Url() {
   this.pathname = null;
   this.path = null;
   this.href = null;
+  this.href = null;
```
DavidTPate: Redundant variable

lib/url.js
```diff
@@ -212,6 +215,17 @@ Url.prototype.parse = function(url, parseQueryString, slashesDenoteHost) {
     if (atSign !== -1) {
       auth = rest.slice(0, atSign);
       rest = rest.slice(atSign + 1);
+
+      colon = auth.lastIndexOf(':');
```
DavidTPate: The colon is actually optional. I would do this instead:

```js
var authParts = auth.split(':');
this.username = authParts[0];

// A username can be provided without a password (no `:`) or with a blank password (`:` followed directly by `@`)
if (authParts.length === 2) {
  this.password = authParts[1];
}
```

lib/url.js
ben-page: Thanks

jasnell:
@ben-page ... given that this is an api change (even if a small one) it's not likely something that we'd want to land in v0.12. This would need to land in nodejs/node master first then be cherry picked back if appropriate. Best thing to do would be to close this PR here and open a new one against nodejs/node master if there isn't one there already.

PR #25362, opened 2015-05-21
tls: properly handle multi-value RDNs in certificate subject/issuer

PaulSD:
Without this, checkServerIdentity will never succeed if a server
certificate's CN is part of a multi-value RDN.  A multi-value RDN is
sometimes used by CAs in this manner to list an emailAddress, UID,
serialNumber, or other identifier in the DN.

### DIFF
--- a/lib/tls.js
+++ b/lib/tls.js
@@ -220,10 +220,10 @@ exports.checkServerIdentity = function checkServerIdentity(host, cert) {
 };
 
 // Example:
-// C=US\nST=CA\nL=SF\nO=Joyent\nOU=Node.js\nCN=ca1\nemailAddress=ry@clouds.org
+// C=US\nST=CA\nL=SF\nO=Joyent\nOU=Node.js\nCN=ca1 + emailAddress=ry@clouds.org
 exports.parseCertString = function parseCertString(s) {
   var out = {};
-  var parts = s.split('\n');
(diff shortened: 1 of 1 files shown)


jasnell:
@indutny ... thoughts?

indutny:
@jasnell I have several:

* This needs a test case certificate, it might be self-signed or better signed by the ca1 in the `test/fixtures/keys` (needs to be generated using `text/fixtures/Makefile`, so we will be able to update it in future)
* I think we should change the way we parse the CN string. I feel like account for spaces this way may lead to some problems in future. What if we will add a new field to the X509 cert, which will be an array of some subject's CN RDNs?

Thanks!

PaulSD:
I'll add a test cert when I get a chance (probably in a couple days).

I agree that it would be better to parse the CN differently.  Apache and cURL both use something like:
subj = X509_get_subject_name(x509);
i = -1;
while ((i = X509_NAME_get_index_by_NID(subj, NID_commonName, i)) != -1) {
    names.push(X509_NAME_ENTRY_get_data(X509_NAME_get_entry(name, i)));
}
The existing subjectaltname parsing should be updated, as it also depends on whitespace:
https://github.com/joyent/node/blob/master/lib/tls.js#L130
I'll also work on this when I have some time...

PR #25363, opened 2015-05-21
tls: Use OpenSSL default trusted CA list and allow configuration of the trusted CA path

PaulSD:
Please see the commit messages for additional explanation.  These are actually two separate changes - I would have split this into two separate pull requests except that the changes are adjacent in the source code, and the documentation for the changes is interdependent, so I couldn't split this into two pull requests without causing merge conflicts between them.

### DIFF
--- a/src/node_crypto.cc
+++ b/src/node_crypto.cc
@@ -699,6 +699,9 @@ void SecureContext::AddRootCerts(const FunctionCallbackInfo<Value>& args) {
 
   assert(sc->ca_store_ == NULL);
 
+  if (SSL_CTX_set_default_verify_paths(sc->ctx_))
+    return;
+
   if (!root_cert_store) {
     root_cert_store = X509_STORE_new();
 
--- a/src/node_crypto.cc
+++ b/src/node_crypto.cc
@@ -699,6 +699,29 @@ void SecureContext::AddRootCerts(const FunctionCallbackInfo<Value>& args) {
 
   assert(sc->ca_store_ == NULL);
 
+  if (args.Length() == 2) {
+    char* caFile = NULL;
+    char* caPath = NULL;
+    if (args[0]->IsString()) {
+      String::Utf8Value file(args[0]);
+      caFile = *file;
+    } else if (!args[0]->IsUndefined() && !args[0]->IsNull()) {
+      return sc->env()->ThrowTypeError("Bad parameter");
+    }
+    if (args[1]->IsString()) {
+      String::Utf8Value path(args[1]);
+      caPath = *path;
+    } else if (!args[1]->IsUndefined() && !args[1]->IsNull()) {
+      return sc->env()->ThrowTypeError("Bad parameter");
+    }
+    if (!SSL_CTX_load_verify_locations(sc->ctx_, caFile, caPath))
+      return sc->env()->ThrowTypeError("Error loading CA certificates from caFile or caPath");
+    return;
+  }
+
+  if (args.Length() != 0)
+    return sc->env()->ThrowTypeError("Bad parameter");
+
   if (SSL_CTX_set_default_verify_paths(sc->ctx_))
     return;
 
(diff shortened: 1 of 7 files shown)


jasnell:
@indutny ... thoughts on this one?

indutny:
I'm -1 for using default CA list, and +1 for introducing new API ;)

PaulSD:
Why -1 for the default CA list?  Would it be better if I kept the existing behavior as the default, and added some options to allow the user to switch to the OpenSSL CA list instead? (Perhaps a compile-time option to enable the OpenSSL CA list, and/or an environment variable which causes the OpenSSL CA list to be used at runtime?)  I'd like to have some ability to change the CA list without changing code...

indutny:
I like the idea of having an option to use it, but using it by default doesn't seem a good idea to me. Many systems has old list of CAs, and we are trying to keep it up to date in node.js using Mozilla's list. cc @bnoordhuis

PaulSD:
Ok, thanks.  I'll update the patch to make this an option rather than the default...

src/node_crypto.cc
```diff
@@ -699,6 +699,32 @@ void SecureContext::AddRootCerts(const FunctionCallbackInfo<Value>& args) {
 
   assert(sc->ca_store_ == NULL);
 
+  if (args.Length() == 2) {
+    char* caFile = NULL;
+    char* caPath = NULL;
```
bnoordhuis: Can you use `nullptr` here and snake_case the variable names?

src/node_crypto.cc
```diff
@@ -699,6 +699,32 @@ void SecureContext::AddRootCerts(const FunctionCallbackInfo<Value>& args) {
...
+    if (args[0]->IsString()) {
+      String::Utf8Value file(args[0]);
+      caFile = *file;
```
bnoordhuis: This is unsafe; you're assigning a pointer to memory that goes out of scope immediately afterwards.  Happens a few lines below as well.

src/node_crypto.cc
```diff
@@ -699,6 +699,32 @@ void SecureContext::AddRootCerts(const FunctionCallbackInfo<Value>& args) {
...
+    } else if (!args[0]->IsUndefined() && !args[0]->IsNull()) {
+      return sc->env()->ThrowTypeError("Bad parameter");
```
bnoordhuis: I'd move the type checking to JS land and just `CHECK(args[0]->IsString())` here.

src/node_crypto.cc
```diff
@@ -699,6 +699,32 @@ void SecureContext::AddRootCerts(const FunctionCallbackInfo<Value>& args) {
...
+    }
+    if (args[1]->IsString()) {
+      String::Utf8Value path(args[1]);
+      caPath = *path;
+    } else if (!args[1]->IsUndefined() && !args[1]->IsNull()) {
+      return sc->env()->ThrowTypeError("Bad parameter");
+    }
+    if (!SSL_CTX_load_verify_locations(sc->ctx_, caFile, caPath))
+      return sc->env()->ThrowTypeError("Error loading CA certificates from caFile or caPath");
```
bnoordhuis: Long line.

src/node_crypto.cc
```diff
@@ -699,6 +699,32 @@ void SecureContext::AddRootCerts(const FunctionCallbackInfo<Value>& args) {
...
+    return;
+  }
+
+  if (args.Length() != 0)
+    return sc->env()->ThrowTypeError("Bad parameter");
```
bnoordhuis: I'd drop this and put a `CHECK(args.Length() == 0 || args.Length() == 2)` at the top.

src/node_crypto.cc
```diff
@@ -699,6 +699,32 @@ void SecureContext::AddRootCerts(const FunctionCallbackInfo<Value>& args) {
...
+
+  if (SSL_CTX_set_default_verify_paths(sc->ctx_))
+    return;
```
bnoordhuis: I understand how this works but it's very implicit.  Can you add a comment explaining it and maybe move it into the `if` block a few lines up?

bnoordhuis:
One caveat with this change: openssl lazy-loads the certificate files blocking I/O.  If the files are located on a slow file system (like NFS), it's going to block the node process for some time.

Issue #25348, opened 2015-05-18
0.12.3 on Windows XP give not valid win32 application error


deliverymanager:
I tested on a Windows XP platform.
a) 0.10.33 works fine.
b) 0.12.2 works fine
c) 0.12.3 after installing without a problem, I try to open the Node.exe and it gives me an error "node.exe is not a valid win32 application"

I have also tested the same win32 x86 0.12.3 version on a windows 7 platform and it works fine.
So, probably, this is related to a windows XP problem.

jasnell:
duplicate of https://github.com/joyent/node/issues/25347

misterdjules:
@orangemocha Do you have some time to investigate?

orangemocha:
Not a dupe of #25347, as that one was reported on Windows 7, and this issue is specific to Windows XP.

Most likely would need to backport https://github.com/nodejs/io.js/commit/e8d08503c7821e8c92e9fa236ed7328e9bdfe62a

jasnell:
@orangemocha ah... ok.. missed that distinction entirely.

JimHolloman:
I have the same issue on XP, 32-bit.  Both the installation (.msi) package and the stand-alone node.exe file (I downloaded both) provide the message "node.exe is not a valid win32 application". The npm provides the same error with the program named changed to npm.exe.

I wonder if someone packaged the 64-bit code as a 32-bit install (.msi) and stand-alone program (node.exe).

orangemocha:
Ok, the problem is the gyp/project files don't explicitly specify a subsystem version, which results in the default being used. The default changed from VS 2010 to VS 2012 and later. So this means that one of the release machines still has VS 2010. (@misterdjules we should probably fix that or at least try to document it in an asset sheet). https://github.com/nodejs/io.js/commit/e8d08503c7821e8c92e9fa236ed7328e9bdfe62a fixes the issue. Opening a PR shortly.

orangemocha:
Had a chat with @misterdjules and he told me that the release machines have both VS 2010 and 2013. So the change that probably introduced the regression is https://github.com/joyent/node/commit/a58b17456a7fa4ff846c4883b96d6dee7e46a62f, where we select a different toolset version.
Since that change is not in v0.10, the v0.10 release still works in XP. I will cherry pick https://github.com/joyent/node/pull/25367 into v0.10 later.

misterdjules:
[...]

PR #25367, opened 2015-05-22
win: bring back xp/2k3 support

orangemocha:
Issue: https://github.com/joyent/node/issues/25348
The gyp/project files don't explicitly specify a subsystem version,
which results in the default being used. The default differs between
linked versions, and it changed from VS 2010 to VS 2012 and later.

Backport e8d08503c7821e8c92e9fa236ed7328e9bdfe62a from io.js.
Original commit message follows:

Chrome still runs on Windows XP, so there is no reason that iojs
couldn't.

PR: https://github.com/iojs/io.js/pull/512
Reviewed-by: Ben Noordhuis info@bnoordhuis.nl
(cherry picked from commit e8d08503c7821e8c92e9fa236ed7328e9bdfe62a)

### DIFF
--- a/common.gypi
+++ b/common.gypi
@@ -135,8 +135,21 @@
       },
       'VCLinkerTool': {
         'conditions': [
+          ['target_arch=="ia32"', {
+            'TargetMachine' : 1, # /MACHINE:X86
+            'target_conditions': [
+              ['_type=="executable"', {
+                'AdditionalOptions': [ '/SubSystem:Console,"5.01"' ],
+              }],
+            ],
+          }],
           ['target_arch=="x64"', {
-            'TargetMachine' : 17 # /MACHINE:X64
+            'TargetMachine' : 17, # /MACHINE:AMD64
+            'target_conditions': [
+              ['_type=="executable"', {
+                'AdditionalOptions': [ '/SubSystem:Console,"5.02"' ],
+              }],
+            ],
           }],
         ],
         'GenerateDebugInformation': 'true',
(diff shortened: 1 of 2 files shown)


orangemocha:
cc @joyent/node-coreteam , @piscisaureus

misterdjules:
LGTM, thank you @orangemocha :+1:

indutny:
As far as I know, libuv does not support XP anymore. Are we sure that we could provide support for this platform?

indutny:
cc @saghul

misterdjules:
@indutny By fixing this issue, we want to fix a hard regression (not even being able to run node on XP). This is not a commitment for full support of Windows XP at the same level of any other "supported" platform.

As far as I know, there's no formal support commitment from the project to any platform anyway, which is I think one thing that @orangemocha wanted to clarify sooner than later, precisely to set expectations for users when they encounter these types of issues (and hopefully to prevent them from happening).

misterdjules:
My understanding is that including this change in the next release (v0.12.4) can't hurt and fixes a regression for all XP users. Thus, I landed this PR in 3122052890e838220b49794de924b323675b3b18 and I'm closing this issue but please feel free to add more comments and continue the discussion around supported platforms.

Thank you!

indutny:
If we will ever decide on merging this into io.js, I suggest we should have a TSC conversation. It doesn't look like that was a proper discussion of this here.

piscisaureus:
> If we will ever decide on merging this into io.js, I suggest we should have a TSC conversation. It doesn't look like that was a proper discussion of this here.

@indutny 
Are you sure you meant to put this comment on this PR?
The patch was cherry-picked *from* io.js.

indutny:
Ah, so it was discussed... Sorry for this then :)

saghul:
FWIW, in libuv we kind of promised not to break XP/2k3 support in the v1.x
series. What happens later, we'll see :-)

PR #25368, opened 2015-05-22
test,win: speedup tls-server-verify

orangemocha:
Addresses https://github.com/joyent/node/issues/25279 (and the related https://github.com/nodejs/io.js/issues/1461 once ported there).

The test now takes half as long as it did before.

### DIFF
--- a/deps/openssl/openssl/apps/app_rand.c
+++ b/deps/openssl/openssl/apps/app_rand.c
@@ -124,10 +124,16 @@ int app_RAND_load_file(const char *file, BIO *bio_e, int dont_warn)
     char buffer[200];
 
 #ifdef OPENSSL_SYS_WINDOWS
-    BIO_printf(bio_e, "Loading 'screen' into random state -");
-    BIO_flush(bio_e);
-    RAND_screen();
-    BIO_printf(bio_e, " done\n");
+    /*
+     * allocate 2 to dont_warn not to use RAND_screen() via
+     * -no_rand_screen option in s_client
+     */
+    if (dont_warn != 2) {
+      BIO_printf(bio_e, "Loading 'screen' into random state -");
+      BIO_flush(bio_e);
+      RAND_screen();
+      BIO_printf(bio_e, " done\n");
+    }
 #endif
 
     if (file == NULL)
--- a/deps/openssl/openssl/apps/s_client.c
+++ b/deps/openssl/openssl/apps/s_client.c
@@ -233,6 +233,7 @@ static int ocsp_resp_cb(SSL *s, void *arg);
 static BIO *bio_c_out = NULL;
 static int c_quiet = 0;
 static int c_ign_eof = 0;
+static int c_no_rand_screen = 0;
 
 #ifndef OPENSSL_NO_PSK
 /* Default PSK identity and key */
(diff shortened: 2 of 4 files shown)


orangemocha:
@indutny , I noticed that in the test case that has renegotiate:true only connection with the first client is renegotiated, and the test depends on that. That's why we didn't parallelize clients if renegotiate is true. Was that intentional / mean to test that specific sequence? If not, we could refactor that test case to speed up things a bit more.

orangemocha:
With the -no-rand-screen changes, the test is now much much faster (less than 1s) on my machine. I am not sure however if speeding up a test is a good justification for adding a floating patch in openssl. I added all the commits in here to get more feedback.

joaocgreis:
I updated the commits as suggested in https://github.com/nodejs/io.js/pull/1836 .

orangemocha:
@joyent/node-coreteam : these changes already landed in io.js with https://github.com/nodejs/io.js/pull/1836. Can you give a quick +1 so I can land this PR here?

Please note that the commits about -no_rand_screen are floated on top of openssl.

jasnell:
LGTM

orangemocha:
Thank you, landed in https://github.com/joyent/node/commit/10349829f2f462e57cf7db2ef4c3a387d23e4826 and adjacent commits.

misterdjules:
@orangemocha Thanks! Could you please add the floating patches to the list here: https://github.com/joyent/node/wiki/OpenSSL-upgrade-process#floating-patches? Also, is there a chance that these patches are landed in OpenSSL upstream?

orangemocha:
> Could you please add the floating patches to the list here: https://github.com/joyent/node/wiki/OpenSSL-upgrade-process#floating-patches?

Updated. Thanks for reminding me @misterdjules !

> Also, is there a chance that these patches are landed in OpenSSL upstream?

Quoting @shigeki : 
> The patch of -no-rand-screen would not be accepted to the upstream because it can not be applied in a general use. This fix was made carefully to be as small as minimum and limited to have changes of only openssl-cli.exe binary so that libopenssl.a linked to node is nothing to do with it.

misterdjules:
@orangemocha @shigeki Thank you and great work on that one again! Out of curiosity, have we tried to actually upstream the `s_client` patch? Maybe a compromise could be reached that would avoid floating another patch?

shigeki:
@misterdjules I think that this patch is hard to be accepted to the upstream. The code of `RAND_screen` is really old code since SSLeay about more than 10 years ago and I cannot insist my confidence that we have enough entropy to be secure without this in a general use case. As for floating patch to openssl-clie.exe in node repo for using unit test, another fix of keypress have already been introduced in https://github.com/joyent/node/commits/master/deps/openssl/openssl/apps/s_client.c . We have two floating patches as a result.

misterdjules:
@shigeki I put up [a list of current floating patches on top of OpenSSL in joyent/node](https://github.com/joyent/node/wiki/OpenSSL-upgrade-process#floating-patches), and we are already at 5 floating patches. I wouldn't be surprised if I missed a few. My main concern is having this list grow over time to something that becomes difficult to manage.

If it doesn't take a lot of effort for you (and if you have time) to send a patch upstream and get some feedback from the OpenSSL project, I would still think that's worth it just to try and see how that goes.

PR #25372, opened 2015-05-23
disable modp1 by default and deprecate modp2 and modp5 in getDiffieHellman

### LINKED ISSUE #25366: Remove (or discourage) small DH groups in crypto.getDiffieHellman
Hi,
the set of predefined RFC DH groups offered by getDiffieHellman should be reconsidered. Due to the new Logjam revelations (https://weakdh.org/sysadmin.html, search for OpenSSH), groups of preset prime and of size smaller than 2048 bits should be considered easily breakable (well, it's not Logjam that showed it, but only put the problem into the public's attention).

According to https://tools.ietf.org/html/rfc4253#section-8.1 Oakley 2 Group is a MUST in SSH protocol implementations (there is at least one that exists: https://github.com/mscdex/ssh2-streams), so we cannot simply drop it. However we could do either:
     1) drop modp1 group (which has 768 bits) and, possibly, modp5 as well (1536 bits)
     2) deprecate all groups < 2048 bits in the docs
I can prepare a patch for any of these options, but please comment.

Tomasz

thinred:
Fixes #25366 by deprecating modp1, modp2 and modp5 groups in the docs. Some time should be given to the users to update their codebases.

### DIFF
--- a/doc/api/crypto.markdown
+++ b/doc/api/crypto.markdown
@@ -425,21 +425,29 @@ expected.
 ## crypto.getDiffieHellman(group_name)
 
 Creates a predefined Diffie-Hellman key exchange object.  The
-supported groups are: `'modp1'`, `'modp2'`, `'modp5'` (defined in [RFC
-2412][]) and `'modp14'`, `'modp15'`, `'modp16'`, `'modp17'`,
-`'modp18'` (defined in [RFC 3526][]).  The returned object mimics the
-interface of objects created by [crypto.createDiffieHellman()][]
-above, but will not allow to change the keys (with
-[diffieHellman.setPublicKey()][] for example).  The advantage of using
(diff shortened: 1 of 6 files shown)


mhdawson:
I think that since this would apply to 0.10.X stream we'd want a PR for merging the change into v0.10 and then we'd merge into 0.12.X.

The other question is whether we should add the check to crypto.createDiffieHellman(prime_length) as well.  ie reject a prime_length smaller than 768

thinred:
1st question: Ok, I'll rebase it.
2nd question: I'd say that no, the issue primarily applies to *well-known* groups (but, sure, using a small prime would be an issue, but it was an issue even before).

thinred:
I created a new PR on top of 0.10 in #25564.

PR #25382, opened 2015-05-27
repl: fix tab completion for a non-global context

sixmen:
a context is not an instance of 'Context' from v0.11.7


cjihrig:
LGTM. I think this should go into the 0.12 branch. I haven't tested, but this is likely an issue on io.js as well.

I think this has been slipping through the cracks because the failed assertions are not being properly reported. To see what I'm talking about, check out https://github.com/cjihrig/node/commit/42e7963cd6c3f2cc9da0c07ae8e7bc50221b9f6d.

jasnell:
LGTM. Landing in v0.12 makes sense.

PR #25456, opened 2015-05-28
http: fix crash when setting a header value to a non-string with a newline

Jimbly:
From conversation [here](https://groups.google.com/forum/#!topic/nodejs/S5aV7V_PRqw), it seems storeHeader() happily converts any value to a string, unless the value, when converted to a string, has a newline in it, in which case is crashes.  This should fix that to also convert to a string when stripping newlines.


jasnell:
can you please add a test case along with this?

Jimbly:
Test added, verified that it fails before my fix.

PR #25460, opened 2015-05-29
Fixed Typo in events.js

webpapaya:
In lib/events.js there is a typo error in the comments.

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -208,7 +208,7 @@ EventEmitter.prototype.once = function once(type, listener) {
   return this;
 };
 
-// emits a 'removeListener' event iff the listener was removed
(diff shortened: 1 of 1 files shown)


KenanY:
Could be shorthand for "if and only if".

gtomitsuka:
@KenanY Meh, then the author should just write "if and only if". The original typo(https://github.com/joyent/node/commit/84221fd1d690070877adacd365ac6429b92bd838) author is @bnoordhuis. @bnoordhuis Is that a typo or a not?

bnoordhuis:
Not a typo, see https://github.com/nodejs/io.js/pull/1832.

Issue #25466, opened 2015-06-01
events: shared object references are emitted to listeners


samhmills:
When emitting to multiple listeners for the same event type, the for 
loop iteration over these listeners shares the same arguments. This 
means (for objects), the copy is only shallow, and changes made by one 
listener are overwritten on the object for the next listener.

For example, if we pass an object with arrays: as below

``` js
EventEmitter.emit("event", {
  data: [{
      users: [9237895]
  }]
});
```

And the first listener sets the users array as null like so: 

``` js
EventEmitter.on('event', function(packet) {
    var data = packet.data[0];
    data.users = null;
});
```

The second event listener will ALWAYS show the array as null. This is
because only a shallow copy of the arguments is made and not a deep
copy.

Here is the full code snippet that will reproduce the issue predictably:

``` js
var EventEmitter = require('events').EventEmitter;
var EE = new EventEmitter();

console.log("starting")

EE.on('event', function(packet) {
    //console.log(packet); //Log pre to see that it is not null
    var data = packet.data[0];
        data.users = null;  
});

EE.on('event', function(packet) {
    console.log(packet.data[0]); //packet.data[0].users will equal null.
});


setInterval(function(){
    console.log("emitting");
    EE.emit("event", {
        data: [
            {
                users: [9237895] //This outputs as null on the second event listener;
            }
        ]
    });
}, 5000)
```

jasnell:
Yes, I've run into this a few times myself. It's definitely not ideal behavior. I'll schedule some time to review the PR but changing this behavior is, unfortunately, an API change that needs to be handled with a bit of care. 

cc @joyent/node-coreteam

Fishrock123:
Oh joy. I bet at least some people are using this in convoluted ways.

cjihrig:
I'm -1 on this. While I sympathize with the current implementation's problems, performing deep copies can be expensive. Isn't it JavaScript convention to pass objects around by reference anyway?

Fishrock123:
-1 for the same reasons as @cjihrig though.

jasnell:
@cjihrig ... generally yes, but this behavior should, at the very least, be documented and as far as I can tell it is not. I'm just as concerned about the deep copying tho. Forcing a deep copy for every listener is going to be extremely expensive and likely completely unnecessary. However, it may be worthwhile providing a utility that performs a deep copy on a per listener basis... something along the lines of:

```
var listener = function(arg) { ... };
[...]

PR #25467, opened 2015-06-01
events: fix: shared object references

samhmills:
Fixes the issue: https://github.com/joyent/node/issues/25466

When emitting to multiple listeners for the same event type, the for
loop iteration over these listeners shares the same arguments. This
means (for objects), the copy is only shallow, and changes made by one
listener are overwritten on the object for the next listener.

For example, if we pass an object with arrays: as below

``` js
EventEmitter.emit("event", {
  data: [{
      users: [9237895]
  }]
});
```

And the first listener sets the users array as null like so:

``` js
EventEmitter.on('event', function(packet) {
    var data = packet.data[0];
    data.users = null;
});
```

The second event listener will ALWAYS show the array as null. This is
because only a shallow copy of the arguments is made and not a deep
copy.

This fixes that issue, by enforcing a deep copy and removing all
references from the new arguments array passed to each listener.

Link to code used to reproduce this issue:
https://ghostbin.com/paste/v4ova

Link to issue:
https://github.com/joyent/node/issues/25466

### DIFF
--- a/doc/api/events.markdown
+++ b/doc/api/events.markdown
@@ -118,6 +118,10 @@ Execute each of the listeners in order with the supplied arguments.
 
 Returns `true` if event had listeners, `false` otherwise.
 
+**Note:** Objects are passed by reference, meaning that the same object can be 
+passed to multiple listeners. If a listener needs to modify a shared object, a 
+copy of the object should be made.
+
 
 ### Class Method: EventEmitter.listenerCount(emitter, event)
 


jasnell:
@Hunchmun ... you should be able to squash these commits down easily by doing a `git reset --soft HEAD~3` then redoing the final commit followed by a `git push -f` to force the update on your branch. Makes things a bit cleaner when landing.

Hunchmun:
@jasnell Thanks for your help, completed.

doc/api/events.markdown
```diff
@@ -44,6 +44,11 @@ calls passing the same combination of `event` and `listener` will result in the
 
 Returns emitter, so calls can be chained.
 
+**Note:** Due to Pass by Reference, any objects passed to event listeners will be
```
cjihrig: How about this?

> **Note:** Objects are passed by reference, meaning that the same object can be passed to multiple listeners. If a listener needs to modify a shared object, a copy of the object should be made.

cjihrig: I wonder if this note should go on `emit()` instead of `addListener()`.

jasnell: Likely. and +1 on the revision.

jasnell:
@Hunchmun ... did you mean to close this?

Hunchmun:
@jasnell nope, my apologies, this was as a result of resetting the branch, forgot to reopen, 

Thanks,

jasnell:
Thanks @Hunchmun! ... LGTM

jasnell:
Landed in a24db43

PR #25469, opened 2015-06-01
fs: Fix StatWatcher to handle error codes passed in status

### LINKED ISSUE #25345: fs.watchFile regression in 0.12
When `fs.watchFile(path, callback)` is called on an non-existing file, the `0.12` behavior is different than `0.10`.

In `0.10.38`, the callback is never called.

In `0.12.2`, the callback is called immediately with this object as `curr` and `prev`:
```js
{ dev: 0,
  mode: 0,
  nlink: 0,
  uid: 0,
  gid: 0,
  rdev: 0,
  blksize: 0,
  ino: 0,
  size: 0,
  blocks: 0,
  atime: Thu Jan 01 1970 01:00:00 GMT+0100 (CET),
  mtime: Thu Jan 01 1970 01:00:00 GMT+0100 (CET),
  ctime: Thu Jan 01 1970 01:00:00 GMT+0100 (CET),
  birthtime: Thu Jan 01 1970 01:00:00 GMT+0100 (CET) }
```

This new behavior breaks one of my scripts.

dnakamura:
fixes #25345

### DIFF
--- a/lib/fs.js
+++ b/lib/fs.js
@@ -1225,7 +1225,7 @@ function StatWatcher() {
 
   this._handle.onchange = function(current, previous, newStatus) {
     if (oldStatus === -1 &&
-        newStatus === -1 &&
+        newStatus < 0 &&
         current.nlink === previous.nlink) return;
 
     oldStatus = newStatus;
--- a/test/simple/test-fs-watchFile.js
+++ b/test/simple/test-fs-watchFile.js
@@ -1,0 +1,15 @@
+var assert = require('assert');
+var fs = require('fs');
+
+function callbackFail(){
+    assert(false, "Callback should not be called for non-existant files")
+}
+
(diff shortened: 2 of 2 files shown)


cjihrig:
Can you add a test?

dnakamura:
@cjihrig  Tests are added, can you take a look?

cjihrig:
It looks like io.js started to go down this path in https://github.com/nodejs/io.js/pull/2028, but decided against it. The current proposed fix is in https://github.com/nodejs/io.js/pull/2093. I think we should try to stay consistent with io.js in this case.

thefourtheye:
@cjihrig Strictly speaking, https://github.com/nodejs/io.js/pull/2093/ is not a fix for this, it simply makes this as a documented feature :-)

PR #25472, opened 2015-06-03
_http_server.js: fix typo in comment

devonharvey:
Fix misspelling of 'response' on line 453 of _http_server.js

BEFORE:
// When we're finished writing the response, check if this is the last
 // respose, if so destroy the socket.
AFTER:
// When we're finished writing the response, check if this is the last
 // response, if so destroy the socket.

No other changes were made.

### DIFF
--- a/lib/_http_server.js
+++ b/lib/_http_server.js
@@ -450,7 +450,7 @@ function connectionListener(socket) {
     }
 
     // When we're finished writing the response, check if this is the last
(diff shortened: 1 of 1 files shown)


cjihrig:
LGTM

jasnell:
Landed in https://github.com/joyent/node/commit/016ff771905b3ac82e0269b46bb98c011fa8a282

PR #25475, opened 2015-06-04
[libuv] upgrade to 1.6.1

saghul:
R=@misterdjules

### DIFF
--- a/deps/uv/src/win/fs-event.c
+++ b/deps/uv/src/win/fs-event.c
@@ -72,13 +72,13 @@ static int uv_split_path(const WCHAR* filename, WCHAR** dir,
 
   if (i == 0) {
     if (dir) {
-      *dir = (WCHAR*)malloc((MAX_PATH + 1) * sizeof(WCHAR));
+      *dir = (WCHAR*)uv__malloc((MAX_PATH + 1) * sizeof(WCHAR));
       if (!*dir) {
-        uv_fatal_error(ERROR_OUTOFMEMORY, "malloc");
+        uv_fatal_error(ERROR_OUTOFMEMORY, "uv__malloc");
       }
 
       if (!GetCurrentDirectoryW(MAX_PATH, *dir)) {
-        free(*dir);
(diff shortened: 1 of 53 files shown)


saghul:
Make that 1.6.1 :-)

deps/uv/src/win/fs-event.c
```diff
@@ -225,11 +225,10 @@ int uv_fs_event_start(uv_fs_event_t* handle,
   }
 
   if (!handle->buffer) {
-    handle->buffer = (char*)_aligned_malloc(uv_directory_watcher_buffer_size,
-                                            sizeof(DWORD));
+    handle->buffer = (char*)uv__malloc(uv_directory_watcher_buffer_size);
```
misterdjules: @saghul IIUC, here the call to `_aligned_malloc` would be replaced by a call to `malloc`, is it intentional and what is the impact of such a change?

misterdjules:
@saghul [One question](https://github.com/joyent/node/pull/25475#discussion_r32255959) just to make sure I don't leave any stone unturned, otherwise that looks good. Will land asap unless the answer to that question means we need to change something.

Thank you once again @saghul :+1:

deps/uv/src/win/fs-event.c
saghul: That was reviewed back in the day and Bert pointed out that plain malloc
provided the necessary alignment in that case, that's why we dropped it.

misterdjules: @saghul Sounds good, thank you for the clarification!

misterdjules:
LGTM.

misterdjules:
@saghul @joyent/node-collaborators Landing with http://jenkins.nodejs.org/job/node-accept-pull-request/127/.

misterdjules:
Landed in 50f961596d5febb7c18dc01f28c257a423faf765 and 097607c782746ae0130f7e128ae44d5c4d146edc. Thank you very much once again @saghul :+1:

PR #25485, opened 2015-06-06
Fixed typo

nicolafio:


cjihrig:
Not a typo. Short for `if and only if`.

Issue #16525, opened 2015-04-19
CLI option to not print out the trailing newline character


redchair123:
The option is similar to echo's `-n` argument to suppress the newline character.

evanlucas:
Would be nice. As a workaround, one could use `process.stdout.write()`

bahamas10:
:+1: on this... i've used this in the past

    node -pe '5 * 5' | tr -d '\n'

but it's not perfect as it will strip ALL newlines

misterdjules:
That sounds like a reasonable feature request to me, thank you! /cc @joyent/node-coreteam.

@Niggler @bahamas10 @Niggler Would any of you be interested in submitting a PR for that?

jasnell:
+1 this would be useful

PR #25497, opened 2015-06-09
cli: add -n and --printf flags

zain:
This fixes issue #16525. It adds `-n` and `--printf` flags to the CLI to evaluate the script and print the result, similar to `-p` and `--print` but without a trailing newline.

This is my first contribution so feedback is welcome.

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -123,6 +123,7 @@ using v8::Value;
 using v8::kExternalUint32Array;
 
 static bool print_eval = false;
+static bool printf_eval = false;
 static bool force_repl = false;
 static bool trace_deprecation = false;
 static bool throw_deprecation = false;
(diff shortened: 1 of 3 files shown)


Fishrock123:
Arguments parsing will be (is in io.js) overhauled. What would `-pn` or `-np` do? (i.e. combined with regular print?)

zain:
@Fishrock123: as written, `-np` and `-pn` are invalid (`./node: bad option: -np`).

Fishrock123:
@zain Right, but I'm wondering for when we move to an option parser where that is valid. (because we will be.)

zain:
@Fishrock123 ah, got it. Correct behavior for `-np` could be one of three things:

1) print with a newline (and ignore `-n`)
2) print without a newline (and ignore `-p`)
3) error

I'm partial to the first option, because `-p` is more common than `-n`.

jasnell:
given the current status with nodejs/node, this would not land here. A new PR would need to be opened against master on nodejs/node. Going to go ahead and close this here.

PR #25506, opened 2015-06-11
tty: no readline when TERM==dumb

### LINKED ISSUE #5344: Need way to disable REPL ANSI escape codes

saljam:
To stop the REPL from spitting ANSI control codes on
terminals which don't support them. (Fixes #5344)

This is a rebase and style fix of #5602.

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -142,7 +142,8 @@
           useGlobal: true,
           ignoreUndefined: false
         };
-        if (parseInt(process.env['NODE_NO_READLINE'], 10)) {
+        if (parseInt(process.env['NODE_NO_READLINE'], 10) ||
+            process.env['TERM'] === 'dumb') {
           opts.terminal = false;
         }
         if (parseInt(process.env['NODE_DISABLE_COLORS'], 10)) {


samertm:
Any action on this? This would fix a major headache I have with the node repl.

thefourtheye:
Hi, since io.js and node.js are merging at nodejs/node repository, I would recommend submitting this patch there.

jasnell:
That said, this is one that we'd likely want to backport to v0.12...

saljam:
So, should I reopen this there?

jasnell:
Yes, if you would

ChALkeR:
This was reopened as https://github.com/nodejs/node/pull/2712.
Discussion should continue there.

PR #25513, opened 2015-06-12
Upgrade to openssl-1.0.1n to v0.10

shigeki:
This is an upgrade to openssl-1.0.1n to the v0.10 branch. No asm and conf files are changed.

I don't have an access to CI of joyent so `make test` was done on my Ubuntu(x64) and only test-debugger-client.js was failed. I think it is not related to this PR.

OpenSSL-1.0.1n includes alt chain support in https://github.com/openssl/openssl/commit/f7bf8e02dfcb2c02bc12a59276d0a3ba43e6c204 so that re-adding 1024-bit root certs of 1425ccd1b2baca267d3201bfc219e28bb84989fd can be reverted. But it is not included here yet.

For v0.12, `-no_rand_screen` patch and limiting DHParam size below 1024 bits for logjam attack would be needed in addition. I will submit another PR to agains v0.12 branch.


shigeki:
Note that v0.10 does not support DHE so node-v0.10.x is vulnerable only to CVE-2015-1788.

shigeki:
`-no_rand_screen` patch  is reserved when I backport this patches from v0.10 to v0.12. So I will submit a PR to limit DHParam for logjam.

shigeki:
move to #25523 for a new release of 1.0.1o.

Issue #25509, opened 2015-06-11
Update openssl to 1.0.1o


mhdawson:
List of changes: 
   https://github.com/openssl/openssl/blob/OpenSSL_1_0_1-stable/CHANGES

Skimming through them the one most likely to affect end users is:

 *) Reject DH handshakes with parameters shorter than 768 bits.
[Kurt Roeckx and Emilia Kasper]

There are also some related changes we may want to make at the same time:

From io.js:

 https://github.com/nodejs/io.js/pull/1739/

   * Is this  still needed given the openssl update, possibly as it could provide better error info.

https://github.com/nodejs/io.js/pull/1831

  * Do we still need this one given the limitation in the new version of open ssl
  * patch likely applies ok to 0.12.X but 0.10.X would require refactoring

Joyent/node issue

 Drop at DH group modp1 from as suggsted in -https://github.com/joyent/node/issues/25366

 Some related docs:

  *   http://www.internet-computer-security.com/VPN-Guide/Diffie-Hellman.html
  *   https://weakdh.org/imperfect-forward-secrecy.pdf (From 25366)

mhdawson:
@shigeki  and @indutny  comments or suggestions ?

indutny:
It is not only this, but ECParameters thing too. It mainly affects the clients and node.js has `tls.connect()` that is vulnerable to this.

Overall, I'm +1 on upgrading it in v0.12

mhdawson:
I was assuming but probably should have stated that we need to update both 0.10.X and 0.12.X for openssl.

@indutny, any comments on whether we want/need to also try to pull in the changes in the other PR/issues mentioned ?

mhdawson:
Looking more closely at #25366.  This test case using modp1 passes with the openssl fix to limit the size to be >768 (https://github.com/openssl/openssl/commit/63830384e90d9b36d2793d4891501ec0248274330) so its seems changing getDiffieHellman() is separate from any impact from openssl changes.  This would mean we could chose to do this independently, and if we do in 0.10.X and 0.12X guard with an option to revert to the original behavior. 

<PRE>
var crypto = require('crypto');
var alice = crypto.getDiffieHellman('modp1');
var bob = crypto.getDiffieHellman('modp1');

alice.generateKeys();
bob.generateKeys();

var alice_secret = alice.computeSecret(bob.getPublicKey(), null, 'hex');
var bob_secret = bob.computeSecret(alice.getPublicKey(), null, 'hex');

[...]

PR #25514, opened 2015-06-12
tls: make server not use DHE in less than 1024bits

shigeki:
DHE key lengths less than 1024bits is already weaken as pointed out in https://weakdh.org/ . 1024 bits will not be safe in near future. We will extend this up to 2048 bits somedays later. Warning is shown via  console.trace() when its key size if less than 2048 bits. This fixes CVE-2015-4000 together with upgrading to OpenSSL-1.0.1n.

### DIFF
--- a/doc/api/tls.markdown
+++ b/doc/api/tls.markdown
@@ -201,9 +201,12 @@ automatically set as a listener for the [secureConnection][] event.  The
 
     Defaults to `prime256v1`. Consult [RFC 4492] for more details.
 
-  - `dhparam`: DH parameter file to use for DHE key agreement. Use
-    `openssl dhparam` command to create it. If the file is invalid to
-    load, it is silently discarded.
+  - `dhparam`: A string or `Buffer` containing Diffie Hellman parameters,
+    required for Perfect Forward Secrecy. Use `openssl dhparam` to create it.
+    Its key length should be greater than or equal to 1024 bits, otherwise
+    it throws an error. It is strongly recommended to use 2048 bits or
+    more for stronger security. If omitted or invalid, it is silently
+    discarded and DHE ciphers won't be available.
 
   - `handshakeTimeout`: Abort the connection if the SSL/TLS handshake does not
     finish in this many milliseconds. The default is 120 seconds.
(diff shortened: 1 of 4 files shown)


misterdjules:
@shigeki @joyent/node-collaborators Moved to 0.12.6 milestone according to https://github.com/joyent/node/issues/25509#issuecomment-112599898, but please feel free to suggest otherwise if you do not agree with that comment.

shigeki:
test-dhe.js has a 512 bits test. Without this, the test would be failed due to 768 key limit of openssl-1.0.1o. I will submit a new PR to change it with only 1024 and 2048 bits key to pass tests.

shigeki:
#25533 is submitted.

jasnell:
@shigeki ... what do you want to do with this one?

shigeki:
node-v0.12 has already 768 key limits owing to openssl upgrade. I think it has little benefits to raise it up to 1024 bits with introducing a new command line option. So close this now.

mhdawson:
@shigeki, I think the remaining issue was that the ssl upgrade only limits the key size on the client side.  It was still possible to create a server that would accept connections with the smaller key sizes smaller than 768 and that's why it made sense to backport.  Is that incorrect ?

mhdawson:
@shigeki could we we set up a time to discuss ?

shigeki:
@mhdawson Sorry, I missed this was reopened. Yes, you are right. I will update this PR soon. Thanks.

mhdawson:
I see you are back and working on a PR in master, does that mean you'll get to this one soon ?

shigeki:
I think we had better to move this issue to LTS project to decide what to do with a security enhancement with semver-major.

mhdawson:
Somehow I thought the fix to limit the server side was already in 4.X but from the comment on 1831 it sounds like that was not the case.

mhdawson:
I misread the comment, which indicated both client/server status. It does say it only affects the server in 0.12.X as the change is already in 4.X

Created a PR to cover this for 0.12.X https://github.com/nodejs/node/pull/3890

mhdawson:
PR landed in v0.12-staging in https://github.com/nodejs/node/commit/caa16b41d6820d86eeb3d272daa3cef25fa3ed9f.

PR #25518, opened 2015-06-12
remove LAST_ACCESS event from uv fs-event

rihteri:
Remove LAST_ACCESS event from fs.watch on windows.

Please review carefully, as this changes the behaviour of that API.

I think on linux this should work similarly, because there is no IN_ACCESS given to uv__inotify_add_watch.

An even better solution would be to add new event types in addition to 'rename' and 'change'.


rihteri:
And documentation on MSDN is here: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365465%28v=vs.85%29.aspx

cjihrig:
Thanks for opening this. However, this discussion belongs at https://github.com/libuv/libuv.

PR #25522, opened 2015-06-12
path.isAbsolute throws bit more informative message #25521

### LINKED ISSUE #25521: Bit odd error message if p in path.isAbsolute(p) call is not a string
Simple to reproduce

``` javascript
require('path').isAbsolute({})
```

produces
TypeError: undefined is not a function
    at Object.posix.isAbsolute (path.js:479:15)
    at repl:1:17
    at REPLServer.defaultEval (repl.js:132:27)
    at bound (domain.js:254:14)
    at REPLServer.runBound [as eval](domain.js:267:12)
    at REPLServer.<anonymous> (repl.js:279:12)
    at REPLServer.emit (events.js:107:17)
    at REPLServer.Interface._onLine (readline.js:214:10)
    at REPLServer.Interface._line (readline.js:553:8)
    at REPLServer.Interface._ttyWrite (readline.js:830:14)

Should produce something like 'not a string' error ...

veljkopopovic:
Should resolve #25521

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -224,6 +224,7 @@ win32.normalize = function(path) {
 
 
 win32.isAbsolute = function(path) {
+  if (!util.isString(path)) throw new TypeError('Arguments to path.isAbsolute must be a string');
   return win32StatPath(path).isAbsolute;
 };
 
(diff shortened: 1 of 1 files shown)


jasnell:
@veljkopopovic ... because of the current status of the merge with nodejs/node, this would need to be closed here and a new PR opened against master on nodejs/node if it hasn't already been handled.

PR #25533, opened 2015-06-17
test: remove 512 bits test in test-tls-dhe.js

shigeki:
Test of 512 bits key is failed after upgrading openssl-1.0.1o due to its limit of 768 bits key size. Remove it and start from 1024 bits test.


misterdjules:
@joyent/node-collaborators Adding to the 0.12.5 milestone, not to 0.10.39 as this test only exists in v0.12, not in v0.10.

mhdawson:
Landed as: https://github.com/joyent/node/commit/139dbe8bb70901c62c4a594d2e742f191a1dfb85

PR #25534, opened 2015-06-17
udp: Fix potential memory leak in UdpWrap::OnRecv

### LINKED ISSUE #25462: Clang helps to detect 3 errors in src /udp_wrap.cc in UDPWrap::OnRecv
`char* base = static_cast<char*>(realloc(buf->base, nread));`

1. `nread` may be zero. in that case undefined behaviour occur. (in linux, memory is just freed)
2. After that call, `buf->base` may become garbage, since realloc may return different pointer.
3. realloc() may return NULL

dnakamura:
Fixes #25462.  Handle case where realloc fails and returns a null ptr.

### DIFF
--- a/src/udp_wrap.cc
+++ b/src/udp_wrap.cc
@@ -445,7 +445,17 @@ void UDPWrap::OnRecv(uv_udp_t* handle,
     return;
   }
 
-  char* base = static_cast<char*>(realloc(buf->base, nread));
+  char* base = NULL;
+  if (nread != 0) {
+     base = static_cast<char*>(realloc(buf->base, nread));
+    // If realloc fails, fallback to old pointer
+    // safe since we are always shrinking the buffer
+    if(base == NULL)
+        base = buf->base;
+  } else {
+    free(buf->base);
+  }
+
   argv[2] = Buffer::Use(env, base, nread);
   argv[3] = AddressToJS(env, addr);
   wrap->MakeCallback(env->onmessage_string(), ARRAY_SIZE(argv), argv);


dnakamura:
In that case we should still behave properly.  We would be building a buffer from a bogus pointer, but since its size will be zero, it should never be de-referenced. That being said, it probably still makes sense to modify the if statement to ``` if(base == NULL && nread != 0) ```

src/udp_wrap.cc
```diff
@@ -446,6 +446,10 @@ void UDPWrap::OnRecv(uv_udp_t* handle,
   }
 
   char* base = static_cast<char*>(realloc(buf->base, nread));
+  // If realloc fails, fallback to old pointer
+  if(base == NULL && nread != 0)
+      base = buf->base;
```
sam-github: If realloc() failed then `buf->base its original size, which is *not* `nread`, so this "fix" means passing a buffer to `Buffer::Use()` that is smaller than it is claimed to be, which is a memory error.

What problem are you trying to solve?

Note that memory allocation failures should be fatal, see line 310 above. With modern O/S and their memory overcommitment, its pretty hard to get allocs of small memory areas to fail, btw.

socketpair: Also it is not specified if reallc will free memory if nread is zero. Man page said that this will be free()-abke pointer, but it may not be NULL. this may lead to memory leak. Why not to just free() memory if nread become zero?

dnakamura: @sam-github in this case realloc is shrinking the buffer, so using the old larger buffer is ok.  And while ive looked at some implementations of realloc, and they all seem to just return the original pointer if a shrinking realloc fails. However this isn't actually in the spec, and in theory there could be an implementation that does not
@socketpair  C90 does specify that memory will be freed, although C99 seems to have modified it. However you are right, I can't think why just freeing if ncount == 0 wouldnt make sense

dnakamura:
Updated code to reflect suggestions (created fixup commit to avoid destroying old comments, will squash when everyone is happy with the code)

src/udp_wrap.cc
```diff
@@ -445,7 +445,16 @@ void UDPWrap::OnRecv(uv_udp_t* handle,
...
-  char* base = static_cast<char*>(realloc(buf->base, nread));
+  char* base = NULL;
+  if (nread != 0) {
+     base = static_cast<char*>(realloc(buf->base, nread));
+    // If realloc fails, fallback to old pointer
+    // safe since we are always shrinking the buffer
+    if(base == NULL)
+        base = buf->base;
+  } else
+    free(buff->base);
```
socketpair: buf with one letter f

socketpair: Also forgot `buf->base=NULL` for second branch. I think it is more correct to write `buf->base = base` after both branches. I.e.:

```
if (nread) {
     base = static_cast<char*>(realloc(buf->base, nread));
     if (base == NULL)
          base = buf->base;
     else
          buf->base = base;
} else {
    free(buf->base);
    buf->base = base = NULL;
}

```

dnakamura: updating ```buf->base``` is unnecessary as its completely discarded after this function. Its a struct built up for the end use by the end consumer  of the read callback (ie us)

socketpair: OK, but this is not trivial for me.

mhdawson: Could you add { } around the else so we have:

} else {
    free(buff->base);
}
     else
          buf->base = base;

thefourtheye:
If possible, can you include a test?

mhdawson:
Devin can correct me if I'm wrong but I think its going to be very hard to craft a reliable test that generates an OnRecv() callback with length 0 across platforms. 

@thefourtheye  do you have something in mind in terms of how to create the test ?

dnakamura:
correct, it would be tricky. not to mention, testing the other part (realloc failing) isn't possible short of linking against a fake libc

thefourtheye:
@mhdawson Oh no idea. Actually I wanted to understand the impact of this change. So, I thought I could understand better if we had a test :-)

mhdawson:
landed as 5acbee6c9437b922751dcc3f37c0d8ca4933c6b4

PR #25535, opened 2015-06-17
stream_wrap: Fix issue with null file handles

### LINKED ISSUE #25365: Access violation in nodejs when restarting a listening server
Nodejs version 0.12.3
Windows 7 and Windows 2012 x64

For our application we need a https or spdy server which can be restarted using a path on the filesystem. On the first call to open() this functions OK, however on the second call it often crashes nodejs. The crash occurs in

https://github.com/joyent/node/blob/master/src/stream_wrap.cc

on line 492. The variable wrap = NULL.

We have tried this with https as well as with spdy.
Code:

``` javascript
function RestartableHttps() {
    var server = null;
    var sockets = [];
    var closing = false;
    var setServer=function(options,hostName) {
        server = spdy.createServer(options, app);
        server.on('connection', function (socket) {
            if (closing) {
                socket.destroy();
                return;
            }
            sockets.push(socket);
            socket.once('close', function () {
                console.log('socket closed');
                sockets.splice(sockets.indexOf(socket), 1);
            });
        });
        closing = false;
        server.listen(443,hostName,function() {
            console.log("Listening on 443");
        });
    }

    var createCertOptions = function (path, type) {
        var keyPath = path + '/Settings/' + type + '.key';
        var certPath = path + '/Settings/' + type + '.crt';
        var caCert = 'Webserver/ca.crt';
        if (!fs.existsSync(keyPath))
            return null;
        if (!fs.existsSync(certPath))
            return null;
        if (!fs.existsSync(caCert))
            return null;
        var options = {
            key: fs.readFileSync(keyPath),
            cert: fs.readFileSync(certPath),
            ca: [fs.readFileSync(caCert)]
        };
        return options;
    }

    var forcedClose = function (cb) {
        if (server === null) {
            cb();
            return;
        }
        closing = true;
        for (var i = 0; i < sockets.length; i++) {
            sockets[i].destroy();
        }
        sockets = [];
        server.close(cb);
    }


    this.open = function (path, certificateType,hostName) {
        var open = function () {
            var options = createCertOptions(path, certificateType);
            if (options === null)
                server = null;
            else
                setServer(options,hostName);
        }
        forcedClose(open);
    }    
}
```

dnakamura:
Fixes #25365. Essentially, a file I/O request is generated and queued, in a different thread the stream is closed and handle set to null. When a worker thread then tries to execute the request it gets a null handle

### DIFF
--- a/src/stream_wrap.cc
+++ b/src/stream_wrap.cc
@@ -224,6 +224,10 @@ void StreamWrap::WriteBuffer(const FunctionCallbackInfo<Value>& args) {
   Environment* env = Environment::GetCurrent(args.GetIsolate());
 
   StreamWrap* wrap = Unwrap<StreamWrap>(args.Holder());
+  if ( NULL == wrap) {
+    args.GetReturnValue().Set(UV_EBADF);
+    return;
+  }
 
   assert(args[0]->IsObject());
   assert(Buffer::HasInstance(args[1]));
@@ -282,6 +286,11 @@ void StreamWrap::WriteStringImpl(const FunctionCallbackInfo<Value>& args) {
   int err;
 
   StreamWrap* wrap = Unwrap<StreamWrap>(args.Holder());
+
+  if ( NULL == wrap) {
+    args.GetReturnValue().Set(UV_EBADF);
+    return;
+  }
 
   assert(args[0]->IsObject());
   assert(args[1]->IsString());
@@ -547,6 +556,10 @@ void StreamWrap::SetBlocking(const FunctionCallbackInfo<Value>& args) {
   HandleScope scope(env->isolate());
 
   StreamWrap* wrap = Unwrap<StreamWrap>(args.Holder());
+  if ( NULL == wrap) {
+    args.GetReturnValue().Set(UV_EBADF);
+    return;
+  }
 
   assert(args.Length() > 0);
   int err = uv_stream_set_blocking(wrap->stream(), args[0]->IsTrue());
--- a/src/stream_wrap.cc
+++ b/src/stream_wrap.cc
@@ -411,6 +411,14 @@ void StreamWrap::Writev(const FunctionCallbackInfo<Value>& args) {
 
   StreamWrap* wrap = Unwrap<StreamWrap>(args.Holder());
 
+  if ( NULL == wrap) {
+    // wrap is set to NULL when the handle for the request
+    // is closed. Since the handle is no longer valid return
+    // UV_EBADF to indicate this
+    args.GetReturnValue().Set(UV_EBADF);
+    return;
+  }
+
   assert(args[0]->IsObject());
   assert(args[1]->IsArray());
 


jasnell:
@trevnorris @cjihrig ... can either of you take a quick look at this as well?

jasnell:
@mhdawson @dnakamura ... we'd likely want to target this PR against v0.12 rather than master. 
@dnakamura ... when you get a chance, would you be able to check v0.10 and the latest io.js as well? We may want to backport this fix to v0.10 and get it landed in the converged repo as well.

src/stream_wrap.cc
```diff
@@ -224,6 +224,10 @@ void StreamWrap::WriteBuffer(const FunctionCallbackInfo<Value>& args) {
   Environment* env = Environment::GetCurrent(args.GetIsolate());
 
   StreamWrap* wrap = Unwrap<StreamWrap>(args.Holder());
+  if ( NULL == wrap) {
```
cjihrig: Can you remove the space between `(` and `NULL` in all of your `if` statements. I think core normally writes as `wrap == NULL` as well.

mhdawson:
In putting together a test case we discovered a few other problems as well.  Devin is working on capturing the test cases and looking at alternate changes to see if what was in the pull request still makes sense or something higher up would make more sense

dnakamura:
I gave some explanation to the underlying problem in #25365. Code for reproducing, which is run in the root directory of https://github.com/EricSmekens/testNodeJS  (which is just used for server certs)

client:
```
var net = require('net');
var tls = require('tls');
var sock = tls.connect(1234, {rejectUnauthorized: false});
sock.pipe(process.stdout);
```

server:
```
var tls = require('tls');
var path = require('path');
var fs = require('fs');

var createCertOptions = function (pth, type) {
        var keyPath = path.join(pth, 'Settings', type + '.key');
        var certPath = path.join(pth , 'Settings' , type + '.crt');
        var caCert = path.join('Webserver', 'ca.crt');
        if (!fs.existsSync(keyPath)){
            console.warn("no keypath ", keyPath);
            return null;
        }
        if (!fs.existsSync(certPath)){
            console.warn("no cert path");
            return null;
        }
        if (!fs.existsSync(caCert)){
            console.warn("no cacert");
            return null;
        }
        var options = {
            key: fs.readFileSync(keyPath),
            cert: fs.readFileSync(certPath),
            ca: [fs.readFileSync(caCert)]
        };
        return options;
}


var regSocket = null;
var dummyFunction = function(){};

function secureConnectionListener(secureSock){
    secureSock.on('close', function(){console.log("Tlssocket close event")});
    setTimeout(function(){
        secureSock.write("HELLO\n");
        regSocket.destroy();
        setTimeout(function(){
            secureSock.write("Never received");
        }, 100);
    }, 100);
}

function connectionListener(sock){
    regSocket = sock;
    sock.on('close', function(){console.log("Regular socket close event")});
}

var server = tls.createServer(createCertOptions('./certs', 'primary'));
server.on('connection', connectionListener);
server.on('secureConnection', secureConnectionListener);
server.listen(1234);
```
Also note that closing the secure socket after closing the normal socket results in unexpected behaviour.
```
function secureConnectionListener(secureSock){
    secureSock.on('close', function(){console.log("Tlssocket close event")});
    setTimeout(function(){
        secureSock.write("HELLO\n");
        regSocket.destroy();
        secureSock.destroy();
    }, 100);
}
```
gives the output:
``` 
node server.js
Regular socket close event
Regular socket close event
```
Changing the order (closing the secure socket before the regular socket) results in the close event on the secure socket being emitted twice

PR #25541, opened 2015-06-18
Reoncile v0.12 with new CI

orangemocha:
As part of CI reconciliation, these commits make v0.12 work with https://jenkins-iojs.nodesource.com/job/iojs+any-pr+multi/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,6 +127,9 @@ test-all-http1: test-build
 test-all-valgrind: test-build
 	$(PYTHON) tools/test.py --mode=debug,release --valgrind
 
+test-ci:
+	$(PYTHON) tools/test.py -ptap --logfile test.tap --mode=release --arch=$(DESTCPU) simple message internet
+
 test-release: test-build
 	$(PYTHON) tools/test.py --mode=release
 
@@ -417,11 +420,6 @@ bench-idle:
 	./node benchmark/idle_server.js &
 	sleep 1
 	./node benchmark/idle_clients.js &
-
-test-ci:
-	$(PYTHON) tools/test.py -ptap --logfile test.tap --mode=release --arch=$(DESTCPU) simple message internet
-	$(MAKE) jslint
-	$(MAKE) cpplint
 
 jslintfix:
 	PYTHONPATH=tools/closure_linter/ $(PYTHON) tools/closure_linter/closure_linter/fixjsstyle.py --strict --nojsdoc -r lib/ -r src/ --exclude_files lib/punycode.js
--- a/Makefile
+++ b/Makefile
@@ -306,11 +306,11 @@ $(PKG): release-only
 	rm -rf $(PKGDIR)
 	rm -rf out/deps out/Release
 	$(PYTHON) ./configure --download=all --with-intl=small-icu \
-		--without-snapshot --dest-cpu=ia32 --tag=$(TAG)
+		--dest-cpu=ia32 --tag=$(TAG)
 	$(MAKE) install V=$(V) DESTDIR=$(PKGDIR)/32
 	rm -rf out/deps out/Release
 	$(PYTHON) ./configure --download=all --with-intl=small-icu \
-		--without-snapshot --dest-cpu=x64 --tag=$(TAG)
+		--dest-cpu=x64 --tag=$(TAG)
 	$(MAKE) install V=$(V) DESTDIR=$(PKGDIR)
 	SIGN="$(APP_SIGN)" PKGDIR="$(PKGDIR)" bash tools/osx-codesign.sh
 	lipo $(PKGDIR)/32/usr/local/bin/node \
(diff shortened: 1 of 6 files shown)


orangemocha:
cc @rvagg & @jbergstroem , since I can't mention @nodejs\build.

orangemocha:
also cc @misterdjules

Makefile
```diff
@@ -306,11 +309,11 @@ $(PKG): release-only
 	rm -rf $(PKGDIR)
 	rm -rf out/deps out/Release
 	$(PYTHON) ./configure --download=all --with-intl=small-icu \
-		--without-snapshot --dest-cpu=ia32 --tag=$(TAG)
```
rvagg: do we know if snapshots are "fixed" in this version of v8? my recollection is that there's a particular version that makes snapshots more safe to ship release builds with

@fishrock123 do you remember the context of this?

Fishrock123: They are fixed in all io.js versions, but not 0.12, is my understanding. See this maybe: https://github.com/nodejs/io.js/pull/1663

orangemocha: Thanks for the heads up. https://github.com/nodejs/io.js/issues/1631#issuecomment-100101375 doesn't seem to mention a specific v8 version where this was fixed. It probably boils down to calling v8::V8::SetEntropySource() at startup? 
In any case, I think we can address the issue of re-enabling v8 snapshots in v0.12 separately (and we should).
In v0.12 we always build without snapshots, so this PR doesn't change that. It is meant to make no-snapshot the default behavior when simply invoking `python configure` as iojs+any-pr+multi does.

jasnell:
+1

Makefile
```diff
@@ -127,6 +127,9 @@ test-all-http1: test-build
 test-all-valgrind: test-build
 	$(PYTHON) tools/test.py --mode=debug,release --valgrind
 
+test-ci:
+	$(PYTHON) tools/test.py -ptap --logfile test.tap --mode=release --arch=$(DESTCPU) simple message internet
```
jbergstroem: Perhaps add an extra space to avoid confusion: `-p tap`

jbergstroem:
LGTM with minor nit.

Makefile
misterdjules: @orangemocha Should we pass `--flaky-tests=dontcare`?

vcbuild.bat
```diff
@@ -202,7 +202,11 @@ if "%test%"=="" goto exit
 if "%config%"=="Debug" set test_args=--mode=debug
 if "%config%"=="Release" set test_args=--mode=release
 
+set test_args=%test_args% --arch=%target_arch% 
+
+
 if "%test%"=="test" set test_args=%test_args% simple message
+if "%test%"=="test-ci"       set test_args=%test_args% -p tap --logfile test.tap simple message internet
```
misterdjules: @orangemocha Do we also want to pass `--flaky-tests=dontcare` here?

configure
```diff
@@ -292,11 +292,16 @@ parser.add_option('--without-perfctr',
     dest='without_perfctr',
     help='build without performance counters')
 
+parser.add_option('--with-snapshot',
+    action='store_true',
+    dest='with_snapshot',
+    help=optparse.SUPPRESS_HELP)
```
misterdjules: `optparse.SUPPRESS_HELP` means that `--with-snapshot` doesn't show when running `configure --help`, I think it would be helpful to add some help similar to the original help that was available for `--without-snapshot`.

Makefile
misterdjules: @orangemocha I'll play the devil's advocate here :)

I'm concerned about the impact of this change for Node.js embedders who might rely on the performance improvement of snapshots being enabled by default. If we change the default build behavior, it might lead a significant performance drop for them. So even though disabling snapshots by default makes it easier for the build process, technically I guess that is a breaking change.

Basically I just want us to be mindful of that use case while considering this change.

misterdjules:
@orangemocha Added a few comments, but this looks like it's moving in the right direction :+1:

Makefile
orangemocha: @jbergstroem : will fix

orangemocha: @misterdjules : I haven't gotten to that part yet. I was planning to make those changes in a separate PR, but I might as well leave this one open and some commits for the node-accept-pull-request job.

Makefile
orangemocha: @misterdjules : do you build releases with snapshots or without? I'd argue that the default should be in line with how releases are built. 

But if that doesn't work, we could add a configure-for-ci parameter to configure.py.

configure
orangemocha: will fix

Makefile
misterdjules: > @misterdjules : do you build releases with snapshots or without? I'd argue that the default should be in line with how releases are built.

No doubt about that, it's really the impact of changing the behavior of the default build for v0.10 and v0.12 users who build node binaries themselves that I'm wondering about.

> But if that doesn't work, we could add a configure-for-ci parameter to configure.py.

I would prefer this approach for v0.10 and v0.12, although I would pick a different option name. We use `build-release` for `vcbuild.bat`, maybe we could use that name consistently across platforms. It's not a great name though because both "build" and "release" are ambiguous terms.

Makefile
misterdjules: Sounds good, either works for me.

orangemocha:
Addressed review feedback in a new PR: https://github.com/joyent/node/pull/25653. Can you folks please take a look at that one?

orangemocha:
Continuing at https://github.com/joyent/node/pull/25653

Issue #25500, opened 2015-06-10
system-wide install instructions should be changed 


manfer:
It is not a good idea to mess up system folders with unnecesary files. The LICENSE, README.md and ChangeLog files has to be excluded when extracting. So it would be a good idea to change the system-wide instructions provided in README.md.

```
cd /usr/local && tar --strip-components 1 -xzf \
                    /path/to/node-<version>-<platform>-<arch>.tar.gz \
                    --exclude='node-*/ChangeLog' \
                    --exclude='node-*/LICENSE' \
                    --exclude='node-*/README.md'
```

sam-github:
Its surprisingly difficult to figure out how to install the linux tarballs on a system, I've regularly come into contact with customers who aren't installing node very well, the docs could definitely use some help on this. @manfer can you PR the README, and be prepared for some discussion?

/cc @rmg

rmg:
Ya, it's kind of unfortunate that there isn't a shorter way to do it.. As a result, I don't usually bother filtering them out since they happen to be harmless in _this_ case.

manfer:
@rmg even if it is harmless it is not a good idea. If all packages do the same, systems will finish with a lot of files in unwanted places. As you comment it is unfortunate not having a shorter way.

Another way would be having different builds for local install or system-wide install. On system-wide install if those files needs to be installed should go into one of the doc folders the same as it is done when building and installing with *make install*. Probably not a good solution.

@sam-github sorry, I can't understand "PR the README" meaning.

It is difficult to decide how to install node in linux. Personally I don't like any method that doesn't make use of the package manager. So that excludes building the package and installing with *make install*, and excludes installing from tarball binary too. Besides, directly unpacking the tarball binary to */usr/local* with the command above, and if I'm not wrong, we install the dev package too which is not necessary for most node users.

So, I was going to install from binary tarball when I opened this issue but finally I decided to create rpms myself. Something that was easy because there is a github project [nodejs-rpm](https://github.com/kazuhisya/nodejs-rpm) with the spec file needed to create rpms from node source which is up to date.

For distros based on enterprise linux there are repositories with versions of node but those are older versions than the current node version. The versions I found in epel and nodesource were 0.10.x. Don't recall the exact build number but it starts with 0.10.

rmg:
@manfer I wasn't disagreeing, I was just saying that in this case I've just decided to live with it on my systems.

[...]

PR #25547, opened 2015-06-19
Update README.md

manfer:
Edited linux system wide instructions to avoid installing the doc files inside /usr/local when extracting the binary tarball.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -266,6 +266,7 @@ endif
 TARBALL=$(TARNAME).tar.gz
 BINARYNAME=$(TARNAME)-$(PLATFORM)-$(ARCH)
 BINARYTAR=$(BINARYNAME).tar.gz
+BINARYDOC=$(BINARYNAME)/share/doc/$(TARNAME)
 PKG=out/$(TARNAME).pkg
 PACKAGEMAKER ?= /Developer/Applications/Utilities/PackageMaker.app/Contents/MacOS/PackageMaker
 
@@ -345,9 +346,10 @@ $(BINARYTAR): release-only
 	$(PYTHON) ./configure --prefix=/ --download=all --with-intl=small-icu \
 		--without-snapshot --dest-cpu=$(DESTCPU) --tag=$(TAG) $(CONFIG_FLAGS)
 	$(MAKE) install DESTDIR=$(BINARYNAME) V=$(V) PORTABLE=1
-	cp README.md $(BINARYNAME)
-	cp LICENSE $(BINARYNAME)
-	cp ChangeLog $(BINARYNAME)
+	mkdir -p $(BINARYDOC)
+	cp README.md $(BINARYDOC)
+	cp LICENSE $(BINARYDOC)
+	cp ChangeLog $(BINARYDOC)
 	tar -cf $(BINARYNAME).tar $(BINARYNAME)
 	rm -rf $(BINARYNAME)
 	gzip -f -9 $(BINARYNAME).tar
--- a/README.md
+++ b/README.md
@@ -54,7 +54,10 @@ Or system-wide with:
 
 ```sh
 cd /usr/local && tar --strip-components 1 -xzf \
-                    /path/to/node-<version>-<platform>-<arch>.tar.gz
+                    /path/to/node-<version>-<platform>-<arch>.tar.gz \
+                    --exclude='node-*/ChangeLog' \
+                    --exclude='node-*/LICENSE' \
+                    --exclude='node-*/README.md'
 ```
 
 ### To run the tests:
--- a/README.md
+++ b/README.md
@@ -53,11 +53,8 @@ tar xzf /path/to/node-<version>-<platform>-<arch>.tar.gz
 Or system-wide with:
 
 ```sh
-cd /usr/local && tar --strip-components 1 -xzf \
-                    /path/to/node-<version>-<platform>-<arch>.tar.gz \
-                    --exclude='node-*/ChangeLog' \
-                    --exclude='node-*/LICENSE' \
-                    --exclude='node-*/README.md'
+tar -xzC /usr/local --strip-components 1 \
+                    -f /path/to/node-<version>-<platform>-<arch>.tar.gz
 ```
 
 ### To run the tests:


rmg:
After some fiddling around, there is a shorter way to accomplish this, but it is probably not as intuitive:
```
cd /usr/local && tar --strip-components 1 -xzf \
                    /path/to/node-<version>-<platform>-<arch>.tar.gz \
                    --include='*/*/*' 
```

More realistically done as:
```
tar -C /usr/local --strip-components 1 --include='*/*/*' -xf node-0.12.4-linux-x86.tar.gz
```

manfer:
I like something like the second version. Sadly include option is not available on all tar versions. I can see it is available on FreeBSD, on OSX too, but for example but my linux distro hasn't that option.

Anyway it is not needed. We can specify at the end of the command the files we want to extract even using a pattern. So the --include is already implicit on tar command.

Though tar is smart enough to find out if the file is gzip compressed I would include the z option anyway.

So this would be my suggestion:

    tar -C /usr/local --strip-components 1 -xzf /path/to/node-<version>-<platform>-<arch>.tar.gz */*/*

rmg:
Good point!

One tweak: Need quotes around the `*/*/*` so the shell doesn't expand them
based on the contents of the current working directory. Probably single
quotes, just to be sure.

manfer:
So

    tar -C /usr/local --strip-components 1 -xzf /path/to/node-<version>-<platform>-<arch>.tar.gz '*/*/*'

would be a nice command for system-wide install of the binary package for people that need node and besides are involved in node development itself as it installs the dev package files too.

I'm trying to find a command for people that are not involved in node development that doesn't need the dev package files. That means excluding *..../include* and *..../share/systemtap*

I have tried so far

    tar -C /usr/local --strip-components 1 -xzf /path/to/node-<version>-<platform>-<arch>.tar.gz '*/[bls]*/[nm]*'

with no luck. It excludes *..../include* but not *..../share/systemtap*. Don't know why it matches systemtap.

As for now I have only been able to do it with two patterns at the end of the command. One for extracting *bin* and *lib* and other to extract *share/man*:

    tar -C /usr/local --strip-components 1 -xzf /path/to/node-<version>-<platform>-<arch>.tar.gz '*/[bl]*/[nm]*' '*/s*/m*'

manfer:
Thinking about it is probably better to use a more intuitive command for people that don't need the dev package files. So better a verbose command including the three folders. It is short enough.

    tar -C /usr/local --strip-components 1 -xzf /path/to/node-<version>-<platform>-<arch>.tar.gz '*/bin' '*/lib' '*/share/man'

jasnell:
@misterdjules ... how does this look to you?

misterdjules:
What's the use case for this? Is it that man pages and header files take too much space? On my system the ones shipped with Linux binaries tarballs take around 1MB of disk space:
```
➜  v0.12 git:(v0.12) du -sh /tmp/node-v0.*-linux-*/{share,include}
 20K	/tmp/node-v0.10.39-linux-x64/share
 24K	/tmp/node-v0.12.5-linux-x86/share
904K	/tmp/node-v0.10.39-linux-x64/include
1020K	/tmp/node-v0.12.5-linux-x86/include
➜  v0.12 git:(v0.12) 
```

manfer:
Just my honest opinion that's not a good reasoning.

If I were to install node from rpm or deb packages, or from a repository that already has nodejs packages, and if I'm not a nodejs developer but just nodejs user I would install nodejs and npm packages and not the dev package.

The same as I do with any other package on my system. Unless I have a good reason to install a dev package I won't. And I don't really care if the dev package takes a lot or few disk space.

For sure if I were installing all dev packages of all the packages installed on my system they wouldn't be just some bytes of disk space.

Anyway, the initial issue for this pull request is to [avoid unpacking LICENSE, README.md and ChangeLog inside /usr/local](https://github.com/joyent/node/issues/25500) which is solved changing the previous command into:

    tar -xzC /usr/local --strip-components 1 '*/*/*' \
      -f /path/to/node-<version>-<platform>-<arch>.tar.gz

which unpacks just files with at least a depth of 2 preventing those 3 files that has only depth 1 from being unpackaged.

misterdjules:
Thank you for the clarification @manfer :+1:

I agree that having the `README.md`, `LICENSE.md` and `ChangeLog` files at the root of `/usr/local` is confusing. However I can imagine some users would be interested in keeping them around and other projects keep similar files (like aclocal or autoconf).

So instead of documenting how to not have them, maybe we could move them to `share/node`?

rmg:
@misterdjules I think that's an excellent idea. I'm pretty sure that's what any RPM or deb re-packaging would be doing anyway.

manfer:
The problem is that the binary tarball is intended for local installs and for that reason LICENSE, README.md and ChangeLog are directly inside the first folder. For local installation that's correct because you want that as soon as people change to that folder after unpacking the tarball they see documentation files.

But for system-wide install that's a problem because you finish installing those three files in wrong place /usr/local.

The packages, deb packages or rpm packages, are correctly packaged to install those files in their correct folder /usr/share/doc

It is posible to change the command to install those three files in /usr/share/doc but it is going to be a more complex command for sure. The command would have to do two unpacks. One that extracts all files but those three. And other that unpacks only those three files to its correct place.

    tar -xzC /usr/local --strip-components 1 '*/*/*' \
      -f /path/to/node-<version>-<platform>-<arch>.tar.gz;
    tar -xzC /usr/share/doc --strip-components 1 --no-wildcards-match-slash \
      '*/ChangeLog' '*/LICENSE' '*/README.md' \      
      -f /path/to/node-<version>-<platform>-<arch>.tar.gz;

Another solution is that the binary package includes the LICENSE, README.md and ChangeLog files in both places */LICENSE and */usr/share/doc/LICENSE and same for the other two. And use for the extraction instructions just:

    tar -xzC /usr/local --strip-components 1 '*/*/*' \
      -f /path/to/node-<version>-<platform>-<arch>.tar.gz;

misterdjules:
@manfer 

> The problem is that the binary tarball is intended for local installs and for that reason LICENSE, README.md and ChangeLog are directly inside the first folder. For local installation that's correct because you want that as soon as people change to that folder after unpacking the tarball they see documentation files.

> But for system-wide install that's a problem because you finish installing those three files in wrong place /usr/local.

Right, I think we agree on the problem. In [my previous comment](https://github.com/joyent/node/pull/25547#issuecomment-116876000), I suggested to change `Makefile` to move these files to `share/node`, wouldn't that solve the problems you mentioned above?

manfer:
Yes that is probably good enough if you want the files intalled when using system-wide unpack instructions. Including the files inside `share/doc/node-<version>` or `share/doc/node` will unpack them into `/usr/local/share/doc`.

And system-wide unpack instructions command needs to be changed from actual:

    cd /usr/local && tar --strip-components 1 -xzf \
                        /path/to/node-<version>-<platform>-<arch>.tar.gz

to:

    tar -xzC /usr/local --strip-components 1 '*/*/*' \
      -f /path/to/node-<version>-<platform>-<arch>.tar.gz;

(if the files are included both in share/doc/node and in the place where they are now)

or just:

    tar -xzC /usr/local --strip-components 1 \
      -f /path/to/node-<version>-<platform>-<arch>.tar.gz;

(because it is not needed to exclude any file if those 3 files are moved)

The current command has to be mandatory changed in my honest opinion. It is an annoying command. First annoyance because you finish in a different folder than the one from which you run the command unless you are already on `/usr/local`. Second annoyance because you have to use an absolute path to node tarball or it won't work as any relative path would be relative to `/usr/local`.

misterdjules:
@manfer Thanks for the detailed explanations. To summarize, I think a PR containing the following changes would be accepted:

1. Move `LICENSE`, `README.md` and `ChangeLog` to `share/doc` when creating binary tarballs. This would probably need to be done in `Makefile`.

2. Change the command to untar prebuilt binaries to `tar -xzC /usr/local --strip-components 1 -f /path/to/node-<version>-<platform>-<arch>.tar.gz;`.

3. Do not mention anything about removing development files. Development files don't take much disk space, they're not breaking anything and they're useful for a lot of users. We don't want to end up making the README too long or difficult to maintain to accommodate small use cases.

@ljharb Would moving these files break nvm in any way? I'm pretty confident it wouldn't, but I'd rather ask before any work is done.

@chorrell Same question for the Node.js docker image, do you rely on the precise location of any of these files?

ljharb:
@misterdjules moving which files? moving the archives themselves would break `nvm` 100%, utterly. The location of all the archives and the naming conventions are hardcoded into `nvm` since the beginning of both `node` and `io.js`.

As for the contents of the archive - the source installs simply rely on `configure` and `make`. However, the binary installs explicitly `tar -xzf $tarball -C $tmpdir --strip-components 1` and then `mv $tmpdir/* $version_directory`, so that seems like it would absolutely break.

Let's please not change installation commands that have been the same since the first version just because it's "annoying", at least not without a lot of thought, and time for a breaking-change warning.

misterdjules:
@ljharb Moving `README.md`, `LICENSE` and `ChangeLog` from `$TARGET_DIR` to `$TARGET_DIR/share/doc` where $TARGET_DIR is the root of the target directory when untaring the binary tarball, not moving the tarballs themselves.

What would break if we move these 3 files to `$TARGET_DIR/share/doc`?

ljharb:
@misterdjules ok thanks for clarifying (it's a long thread and i just woke up). `nvm` does not care about those three files in any way - put them wherever you like :-)

misterdjules:
@ljharb No worries, and thanks for the feedback! :+1:

chorrell:
Hi @misterdjules,

For the Docker Node.js image we should be fine. It doesn't rely on the precise location of any of these files.

RE: the tar command.

This is what is typically used in the Dockerfile to installed Node.js via the tar file:

```
tar -xzf "node-v$NODE_VERSION-linux-x64.tar.gz" -C /usr/local --strip-components=1
```

For node 0.12.* you can see that in context here:

https://github.com/joyent/docker-node/blob/master/0.12/Dockerfile

misterdjules:
@chorrell Sounds good, thank you :+1:

manfer:
@misterdjules

Following your suggestions in a [previous comment](https://github.com/joyent/node/pull/25547#issuecomment-118108751) I changed the pull request.

(I totally removed with a force push the commit that was giving instructions on how to avoid installing development files. Hope it is not a problem as I bet noone has checked out this branch from my fork. Sorry if that was not the best idea)

jasnell:
@misterdjules @orangemocha ... what do you want to do with this one?

rmg:
@jasnell could the decision be deferred if @manfer was willing to open this as a new PR against [nodejs/node](https://github.com/nodejs/node).

With my @nodejs/build hat on, I'm in favour if adopting this change, but I'm not sure this PR has gotten enough attention to be merged considering it changes the layout of the binary distribution.

jasnell:
oh yeah, absolutely. That would be the best approach if this is still an open item.

PR #25560, opened 2015-06-23
IP family is integer 4 or 6, instead of string values. Fixes #4873

### LINKED ISSUE #4873: Uniform IP family values across DNS, Net and OS; as integers
I think it would of great benefit to have a uniform way of reporting IP family for IPv4 and IPv6 across all modules.

Currently _dns.lookup(..)_ returns on a callback with 'family' parameter being integer **4** or **6**, which is good in my opinion.

However, _net.Socket.address()_ and _net.Server.address()_ return objects where the 'family' property is a string 'IPv4' or 'IPv6'.

Also, _os.networkInterfaces()_ returns objects where 'family' is string 'IPv4' or 'IPv6'.

I believe it would be best to turn all _net[Socket|Server].address()_ and _os.networkInterfaces()_ to return the 'family' value as integer **4** or **6**.

**Main reasons for this request:**
- would make things consistent across results from the different modules
- would remove the need of extra parseInt()
- would allow for easy use in mask creation and bitwise operators

Example:
FLAG_NOT_IP = 0x0;
FLAG_IPv4 = 0x1;
FLAG_IPv6 = 0x2;
FAMILY_MASK = ~~(family / 3); _//similar to Math.floor(family / 3)_
// => 0 for family=null, 1 for family=4 (IPv4), 2 for family=6 (IPv6)

mpotra:
Modifies the family property from string, 'IPv4' and 'IPv6', to numeric values 4 respectively 6, and updates tests.
Currently dns.lookup() returns on a callback with 'family' parameter being integer 4 or 6.
However, net.Socket.address(), net.Server.address() and os.networkInterfaces() return objects where the 'family' property is a string, 'IPv4' or 'IPv6'.
os.networkInterfaces() will additionally return the undefined value, for unknown sa family.


cjihrig:
I would recommend opening this on the io.js `next` branch. joyent/node is becoming LTS.

jasnell:
Agree with @cjihrig

jasnell:
Closing this here in favor of the discussion opened on nodejs/node already

PR #25566, opened 2015-06-24
fix example for vm.runInThisContext

jalexanderfox:
vm.createContext(sandbox); returns a contextifiedSandbox which was not being stored in a variable to be used as a parameter in vm.runInContext method.


jasnell:
Thank you but this is not necessary. The `createContext` method modifies the `sandbox` passed into the method. The `contextifiedSandbox` returned by the method is the only relevant if the sandbox parameter is undefined (in which case `createContext` will create it's own sandbox.

jalexanderfox:
Sorry, it helps when you are using the latest version of node when doing the examples... I was using: 0.10.35_2

PR #25567, opened 2015-06-24
Fix msi installer for Visual Studio versions != 2013

misterdjules:
After landing e7c84f82c730bddb7450dc56065aab11b93ffd30, the release process on the Windows build machine was broken. The problem was that the following Visual Studio versions and tools had been installed on that machine, in the following order:
1. Visual Studio 2010
2. WiX
3. Visual Studio 2013

As a result, The WiX SDK was not installed for Visual Studio 2013, because that's taken care of during the WiX installation process. At that time, Visual Studio 2013 was not installed on that machine.

However, `vcbuild.bat` would use Visual Studio 2013 anyway because it could find the corresponding `vcvarsall.bat` file. So it would detect that it can run the [`custom_actions`](https://github.com/joyent/node/blob/v0.10/tools/msvs/msi/custom_actions.vcxproj) project (which has a requirement on the platform toolset that corresponds to Visual Studio 2013), but when it would be time to include the WiX SDK's header for Visual Studio 2013, it would not find it and the Windows installer generation process would fail.

This commit makes the `custom_actions` project not use any hardcoded value related to the Visual Studio version used to generate the Windows installer. Instead, `msbuild` passes the Visual Studio version number that `vcbuild.bat` detects to the `custom_actions` project.

In my case, I had to re-run the WiX installer process to install the WiX SDK for Visual Studio 2013, because I wanted to release the next v0.10.x release with the same compiler than the one that had been used for a while now (since September 2014, when Visual Studio 2013 was installed on that build machine). However this PR fixes the problem for users who don't have access to Visual Studio 2013 for some reason, and I think it just improves the user experience in general (better error messages, etc.). It will come in handy when/if we update the compiler used to build official releases to Visual Studio 2015.

This PR also backports 3122052890e838220b49794de924b323675b3b18 and a58b17456a7fa4ff846c4883b96d6dee7e46a62f as they deal and fix issues about generating installers and binaries on Windows machines with different Visual Studio versions.

/cc @joyent/node-collaborators @joyent/node-tsc

### DIFF
--- a/common.gypi
+++ b/common.gypi
@@ -138,8 +138,21 @@
       },
       'VCLinkerTool': {
         'conditions': [
+          ['target_arch=="ia32"', {
+            'TargetMachine' : 1, # /MACHINE:X86
+            'target_conditions': [
+              ['_type=="executable"', {
+                'AdditionalOptions': [ '/SubSystem:Console,"5.01"' ],
+              }],
+            ],
+          }],
           ['target_arch=="x64"', {
-            'TargetMachine' : 17 # /MACHINE:X64
+            'TargetMachine' : 17, # /MACHINE:AMD64
+            'target_conditions': [
+              ['_type=="executable"', {
+                'AdditionalOptions': [ '/SubSystem:Console,"5.02"' ],
+              }],
+            ],
           }],
         ],
         'GenerateDebugInformation': 'true',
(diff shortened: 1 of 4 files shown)


misterdjules:
Closing and retargeting to v0.10, my apologies.

PR #25568, opened 2015-06-24
fix example for script.runInContext

jalexanderfox:
object literal stored in sandbox was throwing an error, vm.createContext(sandbox); returns a contextifiedSandbox which was not being used in script.runInContext(sandbox);

This modification fixes the example by creating the context and passing it as a parameter to script.runInContext

### DIFF
--- a/doc/api/vm.markdown
+++ b/doc/api/vm.markdown
@@ -222,14 +222,16 @@ execute the code multiple times. These globals are contained in the sandbox.
       animal: 'cat',
       count: 2
     };
+    
+    var contextifiedSandbox = vm.createContext(sandbox);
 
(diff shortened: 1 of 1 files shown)


jasnell:
This is not necessary. The `contextifiedSandbox` returned by the `createContext` method is the exact same sandbox object that is passed in. You can verify that by doing a quick:

```javascript
var vm = require('vm');
var sandbox = {};
console.log(sandbox === vm.createContext(sandbox));
```

jalexanderfox:
Same as #25566, I was using an old version of node 0.10.35_2, my apologies.

jasnell:
No need to apologize! PR's are always welcome!

PR #25572, opened 2015-06-24
Update child_process.markdown, fix VIP example

jalexanderfox:
The original logic would have sent the socket to both child processes instead of just one. By adding the else statement, it should now only send to one child process.

### DIFF
--- a/doc/api/child_process.markdown
+++ b/doc/api/child_process.markdown
@@ -308,9 +308,10 @@ socket to a "special" child process. Other sockets will go to a "normal" process
       if (socket.remoteAddress === '74.125.127.100') {
         special.send('socket', socket);
         return;
-      }
+      } else {
       // just the usual dudes
-      normal.send('socket', socket);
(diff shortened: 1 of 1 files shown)


doc/api/child_process.markdown
```diff
@@ -308,9 +308,10 @@ socket to a "special" child process. Other sockets will go to a "normal" process
       if (socket.remoteAddress === '74.125.127.100') {
         special.send('socket', socket);
         return;
-      }
+      } else {
       // just the usual dudes
```
cjihrig: Can you remove this line. I'm amazed the word `dudes` is here actually.

jasnell:
The addition of the else is unnecessary. The return statement within the if takes care of the potential control flow issue. That said, @cjihrig is right about the "dudes" comment. That should go.

jasnell:
Closing this here. I've opened an PR with a commit that removes the `dudes` from the example

Issue #9269, opened 2015-02-23
Include OpenSSL headers with node distributions


swaj:
When node is installed (or a pre-built binary from nodejs.org is used), there are several header files that are packaged with Node which make native addon development possible.  Amongst them are headers for v8, c-ares, libuv, etc.  I've noticed that there's even an `openssl` subdirectory that contains a single header file:  `opensslconf.h`.

Well it turns out that OSX ships with OpenSSL 0.9.8 by default, and when Node is compiled using its own bundled version of OpenSSL, the verison is much more recent and includes some features that I'd like to use, specifically.

So this feature request is relatively simple:  When Node is built using it's bundled version of OpenSSL (no `--shared-openssl`), can the headers for the included version be shipped in that `include/node/openssl` subdirectory?

tjfontaine:
A related issue is #5112

Keep in mind that in part because we don't have a well defined API/ABI it means that if you link against our compiled openssl you will be stuck using only the symbols we consume. So if there's a deficiency in our existing consumption of openssl, it's likely that just linking against ours won't be sufficient.

Some of this can be fixed with map files etc, but we have to decide what we want our public symbols for node to be. Or you can ship your own copy of openssl with your module, but you want to make sure it's scoped local to your binary addon such that our static openssl doesn't interfere with yours.

swaj:
For me this is perfectly acceptable since I don't actually link statically to OpenSSL.  I always end up using whatever node bundles.

misterdjules:
@swaj It seems to me https://github.com/joyent/node/pull/14089 fixes this issue, could you please confirm? Thanks!

swaj:
Confirmed.  I was able to build 9d19dfbfdb2253e0bf42b55c7c1bbd618f40bf3c and the headers I'm looking for are now included.  Thanks Julien!  Feel free to close this one.

misterdjules:
Thank you very much @swaj :+1: That fix and https://github.com/joyent/node/commit/4028669531bcd30d6082bf0dfdb8d73ddb0717c6 will be included in the upcoming v0.10.39 and v0.12.5 releases. When that will be depends on our progress on fixing recent issues with OpenSSL and logjam. More information here: github.com/joyent/node/issues/25509.

Closing then!

swaj:
@misterdjules I think there may be some more headers missing, unfortunately.  It seems the OpenSSL headers rely on other headers, which aren't present.  A good example is `openssl/evp.h`:

```
[...]

PR #25582, opened 2015-06-26
deps: copy all openssl header files to include dir

shigeki:
On upgrading openssl, all symlinks in pulic header files are replaced with nested include files. The issue was raised that installing them leads to lost its references to real header files.
To avoid this, all public header files are copied into the `deps/openssl/openssl/include/openssl/` directory.
As a result, we have duplicated header files under `deps/openssl/openssl/` but copied files are refereed in build as specified to include path in openssl.gyp.

Fixes: https://github.com/joyent/node/issues/9269

I made a test to install node in my Linux server and confirmed that all real headers of openssl exist in `/usr/local/include/node/openssl/`


misterdjules:
@shigeki Why not removing the openssl header files that are the target of the symlinks (or targets of the #include directives in node's codebase) and keep only `deps/openssl/openssl/include/openssl/*`?

shigeki:
@misterdjules The current master of openssl has already remove them in https://github.com/openssl/openssl/commit/dee502be89e78e2979e3bd1d7724cf79daa6ef61. You can see that it also needs to change `Configure`, `Makefile` and related scripts. Although these files are not used in building with gyp, I think it is better to keep library sources healthy.
Note that iojs uses `Configure` to generate opensslconf.h for various platforms as https://github.com/nodejs/io.js/blob/master/deps/openssl/config/Makefile#L25.

misterdjules:
@shigeki I see what you mean, thank you for the clarification :+1:

My concern is more that by keeping both set of files in our repository, if the OpenSSL project makes changes to the files that are symlink targets, we won't pickup these changes unless we explicitly watch for that. Is there a chance that https://github.com/openssl/openssl/commit/dee502be89e78e2979e3bd1d7724cf79daa6ef61 make it to the 1.0.1 branch at some point?

shigeki:
@misterdjules I wouldn't bet on it. In order to avoid missing changes of header files in the future, I revised the upgrading docs of https://github.com/nodejs/io.js/blob/master/deps/openssl/doc/UPGRADING.md#3-replace-openssl-header-files-in-depsopensslopensslincludeopenssl. 
I made it with a small bash script as
```bash
#!/bin/bash
for h  in *.h
do
  a=`readlink $h`
  echo $a
  mv $h ${h}.org
  cp $a .
done
````

misterdjules:
@shigeki Would it be better to add in `tools/install.py` something that automatically copies the actual header files to `include/node/openssl` at the install step (and thus when building the binary tarball too) instead of doing it manually when upgrading OpenSSL?

That way we do not float another patch on OpenSSL, and we do not run the risk of forgetting to run this step manually during future upgrades of OpenSSL.

That's just a suggestion, you know the OpenSSL code base and the way we integrate it in this repository much better than I do.

shigeki:
@misterdjules Replacing all symlink of header files with real header files is necessary with/without executing `tool/install.py` since symlinks cause issues in extracting source files on Windows. 
We should rather make `tools/openssl_upgrade.py` but it's a long way to go :-)

misterdjules:
@shigeki Right, I'm not questioning removing symlinks, I understand we need that on Windows. What I'm suggesting is that we could replace a manual, error prone, process to copy header files with an automated one in `tools/install.py`.

If that's not an option, then we'll need to mention the additional step of copying header files in the one bundled in official releases in [the OpenSSL upgrade guide](https://github.com/joyent/node/wiki/OpenSSL-upgrade-process).

misterdjules:
This has landed in v0.10 with 82778223444d9717bcd2629443c93677035fc40f.

PR #25586, opened 2015-06-28
doc: Clarified read method with specified size argument.

### LINKED ISSUE #7273: Surprising behavior in `stream.read(n)` when stream ends
It seems that when a stream ends, and it still has unread data in its buffers, a final `readable` event gets emitted, and `stream.read(n)` where `n > unread.length` happily returns the remaining bytes even though it normally wouldn't.

Example of said behavior:

```javascript
var fs = require('fs')
var out = fs.createReadStream(__filename)

out.on('readable', function() {
  console.log('readable', out.read(10000))
})

out.on('end', function() {
  console.log('end')
})
```
(note that the file is considerably smaller than requested 10000 bytes)

Produces:

```
readable null
readable <Buffer 76 61 72 20 66 73 20 3d 20 72 65 71 75 69 72 65 28 27 66 73 27 29 0a 76 61 72 20 6f 75 74 20 3d 20 66 73 2e 63 72 65 61 74 65 52 65 61 64 53 74 72 65 61 ...>
end
```
(i.e. just before ending, it returned what it had, instead of the requested amount)

This behavior is surprising because the [documentation](http://nodejs.org/api/stream.html#stream_readable_read_size) lets you believe that `stream.read(n)` can only ever return the requested amount of bytes:

	If you pass in a size argument, then it will return that many bytes. If size bytes are not available, then it will return null.

I was able to dig a removed [explanation](https://github.com/joyent/node/commit/8acb416ad0532d249a0342f39103dd09fbbeeb2e#diff-ba6a0df0f5212f5cba5ca5179e209a17L86) for this behavior, and I'm not necessarily questioning the validity of it, but it seems very odd and confusing that none of this is documented and even the useful comment has been removed.

plafer:
Made explicitely clear that when size bytes are not available, it will
return null, unless we've ended, in which case the data remaining in the
buffer will be returned.

Fixes #7273

### DIFF
--- a/doc/api/stream.markdown
+++ b/doc/api/stream.markdown
@@ -227,7 +227,9 @@ returns it.  If there is no data available, then it will return
 `null`.
 
 If you pass in a `size` argument, then it will return that many
-bytes.  If `size` bytes are not available, then it will return `null`.
+bytes.  If `size` bytes are not available, then it will return `null`,
+unless we've ended, in which case it will return the data remaining 
+in the buffer.
 
 If you do not specify a `size` argument, then it will return all the
 data in the internal buffer.


jasnell:
LGTM

jasnell:
Will land with https://github.com/joyent/node/pull/25592

jasnell:
Landed with 8e1e18f

PR #25587, opened 2015-06-28
REPL: Fix for multiline input when using custom eval function

### LINKED ISSUE #8640: Multiline input doesn't work if user overrides "eval" in REPL.start
As of b5175003bca45e23bd20d9559096dafda3e0ad7c, multiline input is triggered when the `eval` callback is executed with an instance of `Recoverable`.  However, `Recoverable` is not exported from the "repl" module.  It is therefore no longer possible (AFAICT) to trigger multiline input when overriding "eval" in `REPL.start`.

mikeybox:
Multiline input doesn't work if user overrides "eval" in
REPL.start. Overriding default eval will stop multiline
inputs on syntax errors because of the current error checking
in the finish callback.

This fixes that issue and allows for a custom recoverableError
check function to be passed the the REPL server for more
customisable use.

fixes #8640


doc/api/repl.markdown
```diff
@@ -49,6 +49,8 @@ the following values:
 
  - `eval` - function that will be used to eval each given line. Defaults to
    an async wrapper for `eval()`. See below for an example of a custom `eval`.
+ 
+ - `recoverable` - function that will return a `bool` when passed an error and report if it is recoverable. Use if your `eval` returns none standard errors but you still want the benefits of multiline input.
```
davidchambers: I think this should read _non-standard_ rather than _none standard_.

mikeybox:
thanks @davidchambers updated the copy.

thefourtheye:
Can we see an example to reproduce the problem?

mikeybox:
Hey @thefourtheye this fix was submitted because it breaks multi-line support in Babel.js [#1741](https://github.com/babel/babel/issues/1741)

To reproduce the error you can simply use a custom eval function passed to the REPL server,
you will see that syntax errors are no longer caught in the finish method of the server.
The problem is the event upgrade that happens in the `defaultEval` handler is local to REPL so overriding this means that no syntax errors created in a custom eval will be promoted to `Recoverable`. 

in the `finish` method bellow at line 300 it checks for errors to be an instance of `Recoverable`
this will never happen for errors thrown in a custom eval method.

```javascript
// If error was SyntaxError and not JSON.parse error
      if (e) {
        if (e instanceof Recoverable) {
          // Start buffering data like that:
          // {
          // ...  x: 1
          // ... }
          self.bufferedCommand += cmd + '\n';
          self.displayPrompt();
          return;
        } else {
          self._domain.emit('error', e);
        }
      }
```

jasnell:
@joyent/node-tsc ... this change proposes adding a new option, which is technically an API change, but it's for a good reason. Thoughts on this one?

lib/repl.js
```diff
@@ -94,6 +94,7 @@ function REPLServer(prompt, stream, eval_, useGlobal, ignoreUndefined) {
     ignoreUndefined = options.ignoreUndefined;
     prompt = options.prompt;
     dom = options.domain;
+    recoverableCheck = options.recoverable || isRecoverableError;
```
cjihrig: This would be better as a ternary based on whether `options.recoverable` is a function or not.

lib/repl.js
```diff
@@ -952,9 +953,4 @@ function isRecoverableError(e) {
   return e &&
       e.name === 'SyntaxError' &&
       /^(Unexpected end of input|Unexpected token)/.test(e.message);
-}
-
-function Recoverable(err) {
```
cjihrig: I think there is at least one place left in the code that calls `new Recoverable()`.

test/simple/test-repl-options.js
```diff
@@ -66,6 +67,7 @@ var r2 = repl.start({
   useGlobal: true,
   ignoreUndefined: true,
   eval: evaler,
+  recoverable: recoverTester,
```
cjihrig: I don't think this is really enough testing for a new feature.

cjihrig:
@jasnell I left a few comments. I don't really have an opinion on whether this gets in or not.

mikeybox:
@cjihrig I'll add your suggestions to the PR, in terms of testing you're right it isn't really tested but neither was the eval function (unless I'm missing something??) so I just kept it in line with that as it directly affects that.

smasher164:
@mikeybox Here's the relevant issue on **nodejs/node#2939**

CestDiego:
are there any news on this?

mikeybox:
@CestDiego there is a similar fix in a PR which is further along https://github.com/nodejs/node/pull/3488.

Issue #25087, opened 2015-04-29
x64 Windows installer installs to incorrect path


sbmelvin:
I just downloaded the x64 Windows installation MSI and it wants to install to C:\Program Files (x86)\nodejs  instead of C:\Program Files\nodejs

Is this correct? I thought the default path for 64-bit installers was C:\Program Files\ ?

sbmelvin:
I uninstalled my previous x86 nodejs installation. Now when I run the x64 installer it uses c:\program files\.

misterdjules:
@joyent/node-coreteam Is it the desired behavior that an install of the 64 bits version when a 32 bits is already installed is considered as an upgrade, and thus the target directory is the same as the upgraded version (`C:\Program Files (x86)\` in this case)?

misterdjules:
Adding the confirmed-bug label, even though it's not clear it's a bug.

sbmelvin:
It is unexpected behavior for a 64bit installer to place an application in C:\Program Files (x86)\. If a 64 bit version of node cannot coexist with a 32 bit version of node on the same machine, I think the correct action would be for the installer to ask the user if they wish to uninstall the existing node installation prior to installing the 64 bit version. Otherwise, I think that installing the 64 bit version to C:\Program Files\ (and upgrading any existing 64 bit install) would be the expected behavior.

misterdjules:
@sbmelvin That sounds reasonable to me, thank you :+1: @orangemocha I'm not familiar with how Windows users deploy node on Windows, any thoughts?

joaocgreis:
Fixed by 14db629497d0eb97a5f0fbb70cf3e534e460deec, but this will only work on fresh installs starting from the next release. Thank you!

sbmelvin:
:+1:

Issue #7629, opened 2014-05-15
NodeJS Windows MSI installs Per User


chrpai:
The node-v0.10.28-x86.msi does not have the ALLUSERS property defined in the property table and doesn't have a UI option for selecting per-user / per-machine.  This means by default it installs as per-user.  The resources are installed in per-machine locations though.   This creates servicing problems in that only the person who originally did the install can uninstall it.  I discovered this when I was creating an installer and I wanted to use the Upgrade table to detect the NodeJS install and block installation if it wasn't present.  Since another developer did the NodeJS install my MSI couldn't find the NodeJS installed product.   The following code should be added to the WiX source:

(tag) Property Id="ALLUSERS" Value="1" (endtag)

On a side note, it's not possible to perform a major upgrade and switch installation context.  Deployed versions of NodeJS will have to be first uninstalled.

joaocgreis:
Fixed by 8e80528453aca0354422162e09c7c9f4700ddb1e. Thank you!

heifer1981:
Hi all. I know this issue is closed and that Node has moved on significantly since this. Just to put things into context for this comment we use Meteor but to deploy easily onto Windows hardware we demeteorize our projects so it will run on node and node v 0.10.4 is the only version so far we have found this works for. I spent days trying to find why our WIX Uninstallation would not work and this thread let me to what it was. In my WIX package I force the per machine setting to yes and now the nodejs package uninstalls without error and without needing custom actions to perform this cleanup. 
Eg.             MsiPackage Id="NodeJSMSI"
                  DisplayName="Node JS Framework Installer"
                  Cache="yes"
                  Compressed="no"                  
                  Permanent="no"                  
                  Vital="yes"  
                  DisplayInternalUI="yes"
                  EnableFeatureSelection="yes"
                  ForcePerMachine="yes"
                  SourceFile=".\node-v0.10.40-x64.msi"                 

Hope this helps someone else trying to figure out why nodejs is breaking your uninstallation chain and the settings it requires to perform this operation automagically :)

Issue #5849, opened 2013-07-14
Installation failed: node-v0.10.13-x64.msi 


Spetnik:
When installign node-v0.10.13-x64.msi on my Windows 7 64bit machine, it fails with the message "node.js setup wizard ended prematurely". This is the install log: http://pastebin.com/NYT18xkA

Any help would be greatly appreciated.

Thanks
Aaron

jasnell:
Appears to no longer be an issue. Can reopen if new information is presented.

JimF42:
I just downloaded node-v0.12.4-x64.msi and can confirm that this is indeed **still an issue**, albeit a slight different error message is presented, but it is still the same problem that I posted above on 1/6/2014.

The messages now are:
* MSI (s) (D4:68) [16:36:54:599]: Executing op: ActionStart(Name=CreateFolders,Description=Creating folders,Template=Folder: [1])
* MSI (s) (D4:68) [16:36:54:599]: Executing op: FolderCreate(Folder=\\ad.local\system\users\\[myAcctName]\Start Menu\Programs\Node.js\,Foreign=0,SecurityDescriptor=BinaryData,)
* Info 1926.Could not set file security for file '\\ad.local\system\users\\[myAcctName]\Start Menu\Programs\Node.js\'. Error: 0.  Verify that you have sufficient privileges to modify the security permissions for this file.
* MSI (s) (D4:68) [16:36:54:654]: Error in rollback skipped.	Return: 3

If you like, I can get you the entire log file. Also, due to the nature of the error (at least in my case) I understand that this is not easily recreated by most developers as you need an environment where your user documents (normally located at C:\Users\\[myAcctName] is relocated to a network server. See https://technet.microsoft.com/en-us/library/cc732275.aspx for details.

I do not know if I am allowed to reopen this issue myself, so if I can't, then the owner may need to.

jasnell:
hmm... @orangemocha ... thoughts?

JimF42:
Hey @jasnell, @orangemocha, I think I have a clue. I looked at the file properties of your msi to see if I could determine what "brand" it was (appears to be Microsoft's installer toolkit) to see if I could search for the issue via the installer docs.

[...]

Issue #5592, opened 2013-05-29
Incorrect default install path for x64 node .msi


natecain:
Hello,

On my windows 8 laptop the 64-bit msi installers from the node download page show a default install path of "\Program Files (x86)\" instead of the expected "\Program Files\" path.  This adds a counter-intuitive step in the install workflow, and is especially problematic when both architectures are to be installed side-by-side.

joaocgreis:
Fixed by 14db629497d0eb97a5f0fbb70cf3e534e460deec, but this will only work on fresh installs starting from the next release. Thank you!

PR #25590, opened 2015-06-29
win,msi: change InstallScope to perMachine

joaocgreis:
Currently, installing with the MSI always asks for Administrator permissions, as mentioned in #7629 . Explicitly installing per machine only adds the benefit that shortcuts are placed in ProgramData thus being accessible to all users and also solving #5849 .

The second commit moves the installation path in the registry to Local Machine. The registry stores HKLM in different places for 32 and 64 bit applications, so the installer will not suggest the old path when upgrading from 32 to 64 bit version, solving #5592 and #25087 .

### DIFF
--- a/tools/msvs/msi/product.wxs
+++ b/tools/msvs/msi/product.wxs
@@ -18,7 +18,7 @@
            Manufacturer="$(var.ProductAuthor)"
            UpgradeCode="1d60944c-b9ce-4a71-a7c0-0384eb884baa">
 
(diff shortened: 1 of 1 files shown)


joaocgreis:
Found a problem, will reopen as soon as it's ready for review.

joaocgreis:
Cannot reopen because of https://github.com/isaacs/github/issues/361 . Recreated as 25640.

PR #25591, opened 2015-06-29
doc: apply multiple doc updates from master to v0.12

jasnell:
Applies the several doc updates made to master to v0.12.

### DIFF
--- a/doc/api/stream.markdown
+++ b/doc/api/stream.markdown
@@ -164,6 +164,9 @@ readable.on('readable', function() {
 Once the internal buffer is drained, a `readable` event will fire
 again when more data is available.
 
+The `readable` event is not emitted in the "flowing" mode with the
+sole exception of the last one, on end-of-stream.
+
 #### Event: 'data'
 
 * `chunk` {Buffer | String} The chunk of data.
@@ -181,6 +184,9 @@ readable.on('data', function(chunk) {
   console.log('got %d bytes of data', chunk.length);
 });
 ```
+Note that the `readable` event should not be used together with `data`
+because the assigning the latter switches the stream into "flowing" mode,
+so the `readable` event will not be emitted.
 
 #### Event: 'end'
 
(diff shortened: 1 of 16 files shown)


jasnell:
@joyent/node-tsc ... can I get a quick review on this? These previously already landed in master, just backporting them to v0.12

jasnell:
Added a few additional doc merges that had landed in master

jasnell:
Added a few more. These have all been reviewed and landed in master.

doc/api/stream.markdown
```diff
@@ -181,6 +209,9 @@ readable.on('data', function(chunk) {
   console.log('got %d bytes of data', chunk.length);
 });
 ```
+Note that the `readable` event should not be used together with `data`
+because the assigning the latter switches the stream into "flowing" mode,
```
misterdjules: @jasnell  "the assigning the latter" should probably be reworded. Maybe just "because listening on the latter"?

doc/api/tls.markdown
```diff
@@ -694,6 +694,11 @@ This is an encrypted stream.
 A proxy to the underlying socket's bytesWritten accessor, this will return
 the total bytes written to the socket, *including the TLS overhead*.
 
+## Class: CleartextStream
+
+The CleartextStream class in Node.js version v0.10.39 and prior has been
```
misterdjules: @jasnell Do we mean in all v0.10.x versions? Node v0.10.40 was just released, and this seems a bit confusing as it would mean that `CleartextStream` is available in that version.

misterdjules:
@jasnell Added a few comments.

In general, I'm not sure that the method of landing several commits in master to have them back ported in batch via a PR is a good workflow. It seems that it adds to the manual steps needed to land changes. More reviews of bigger changes are needed, and it also seems like it would be made worse if we needed to back port some of these changes to v0.10.

It seems that landing these changes one by one in the oldest supported branch where they apply and then letting them bubble up along with the merges from "older" branches up to master and via v0.12 doesn't have these drawbacks.

Did you experience any benefit in doing it by back porting?

jasnell:
In general, yes, doing it this was ends up being a significant time saver. These were all already previously reviewed and signed off as PR's against master.

srl295:
LGTM. Some good updates.

jasnell:
Landed

Issue #9074, opened 2015-01-21
_http_agent does not reuse keepAlive sockets if maxSockets = 2


andrasq:
node v0.11 and v0.12 reuse sockets for requests configured as keepAlive = true (yay!)

However, if the Agent is configured for maxSockets = 2, a new socket is used for each connection, in spite of keepAlive = true (as seen with `netstat -na | grep TIME_WAIT | wc -l`).  Each run of the script below will put 10 sockets into TIME_WAIT state.  Changing maxSockets away from 2, to 1 or 3 or 4 or 5 or 20, each run will use only 1 or 2 sockets.

To reproduce:

```
http = require('http');

reqOptions = {
    host: 'google.com',
    port: 80,
    path: '/',
    agent: new http.Agent({
        keepAlive: true,
        maxSockets: 2,
    }),
};

nloops = 10;
function requestLoop() {
    req = http.request(reqOptions, function(res) {
        res.on('error', function(err) {
            console.log(err);
        });
        res.on('data', function(chunk){
            // ignore
        });
        res.on('end', function(){
            if (nloops-- > 0) requestLoop();
        });
    });
    req.end();
};

requestLoop();
```

PR #25600, opened 2015-06-30
Fix keepAlive/maxSockets #9074

wcherif84:
Total number of sockets (used + free) must be less than or equal to maxSockets parameters. So the "count" variable must be greater than maxSockets, and not greater than or equal to maxSockets.


jasnell:
@wcherif84 .. thank you for the patch... A test case along with this change would be ideal.

PR #25603, opened 2015-06-30
Deprecate RC4 ciphers from the default ciphers list

misterdjules:
This PR consolidates previous PRs from @jasnell (https://github.com/joyent/node/pull/14413, https://github.com/joyent/node/pull/14572) and myself (https://github.com/joyent/node/pull/23947) and allows us to have one single place where we can take a look at the current status of deprecating RC4 from the default ciphers list used by node v0.10.x, discuss it, and make changes.

The list of commits hasn't been squashed to preserve the history behind these changes. If this PR lands, its commits will be squashed. Before that, we want to make sure we understand all the implications of these changes.

### DIFF
--- a/lib/crypto.js
+++ b/lib/crypto.js
@@ -134,7 +134,11 @@ exports.createCredentials = function(options, context) {
 
   if (options.cert) c.context.setCert(options.cert);
 
-  if (options.ciphers) c.context.setCiphers(options.ciphers);
+  if (options.ciphers) {
+    c.context.setCiphers(options.ciphers);
+  } else {
+    c.context.setCiphers(binding.DEFAULT_CIPHER_LIST);
+  }
 
   if (options.ca) {
     if (Array.isArray(options.ca)) {
--- a/lib/crypto.js
+++ b/lib/crypto.js
@@ -135,9 +135,9 @@ exports.createCredentials = function(options, context) {
 
   if (options.ciphers) {
     c.context.setCiphers(options.ciphers);
-  } else if (!(process._usingV1038Ciphers() && options.ciphers === undefined)) {
+  } else if (!(process._usingV1040Ciphers() && options.ciphers === undefined)) {
     // Set the ciphers to the default ciphers list unless
-    // --enable-legacy-cipher-list=v0.10.38 was passed on the command line and
+    // --enable-legacy-cipher-list=v0.10.40 was passed on the command line and
     // no ciphers value was passed explicitly. In that case, we want to
     // preserve the previous buggy behavior that existed in v0.10.x until
     // v0.10.39, otherwise, a lot of client code might be broken. Server
--- a/lib/crypto.js
+++ b/lib/crypto.js
@@ -41,6 +41,7 @@ var constants = process.binding('constants');
 
 var stream = require('stream');
 var util = require('util');
+var tls = require('tls');
 
 // This is here because many functions accepted binary strings without
 // any explicit encoding in older versions of node, and we don't want
@@ -114,7 +115,6 @@ function Credentials(secureProtocol, flags, context) {
 
 exports.Credentials = Credentials;
 
-
 exports.createCredentials = function(options, context) {
   if (!options) options = {};
 
@@ -136,7 +136,14 @@ exports.createCredentials = function(options, context) {
 
   if (options.ciphers) {
     c.context.setCiphers(options.ciphers);
-  } else {
+  } else if (!(tls.usingV1038Ciphers() && options.ciphers === undefined)) {
+    // Set the ciphers to the default ciphers list unless
+    // --enable-legacy-cipher-list=v0.10.38 was passed on the command line and
+    // no ciphers value was passed explicitly. In that case, we want to
+    // preserve the previous buggy behavior that existed in v0.10.x until
+    // v0.10.39, otherwise, a lot of client code might be broken. Server
+    // side TLS/HTTPS code always sets a default cipher list explicitly so it
+    // never reaches this, even for versions < v0.10.39.
     c.context.setCiphers(binding.DEFAULT_CIPHER_LIST);
   }
 
--- a/lib/crypto.js
+++ b/lib/crypto.js
@@ -41,7 +41,6 @@ var constants = process.binding('constants');
 
 var stream = require('stream');
 var util = require('util');
-var tls = require('tls');
 
 // This is here because many functions accepted binary strings without
 // any explicit encoding in older versions of node, and we don't want
@@ -136,7 +135,7 @@ exports.createCredentials = function(options, context) {
 
   if (options.ciphers) {
     c.context.setCiphers(options.ciphers);
-  } else if (!(tls.usingV1038Ciphers() && options.ciphers === undefined)) {
+  } else if (!(process._usingV1038Ciphers() && options.ciphers === undefined)) {
     // Set the ciphers to the default ciphers list unless
     // --enable-legacy-cipher-list=v0.10.38 was passed on the command line and
     // no ciphers value was passed explicitly. In that case, we want to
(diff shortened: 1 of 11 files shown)


jasnell:
@misterdjules ... awesome, will review in detail this evening.

lib/crypto.js
```diff
@@ -134,7 +134,18 @@ exports.createCredentials = function(options, context) {
 
   if (options.cert) c.context.setCert(options.cert);
 
-  if (options.ciphers) c.context.setCiphers(options.ciphers);
+  if (options.ciphers) {
+    c.context.setCiphers(options.ciphers);
+  } else if (!(tls.usingV1038Ciphers() && options.ciphers === undefined)) {
```
misterdjules: This looks like a slippery slope, but that's the current state of the discussion. Basically, I'm concerned that having to maintain this type of changes will be a major headache in the future. At the same time, the current consensus was that this would prevent a lot of TLS client code from breaking, since it currently doesn't set a default cipher suite if not passed explicitly.

lib/tls.js
```diff
@@ -1327,6 +1330,21 @@ function normalizeConnectArgs(listArgs) {
...
+function usingV1038Ciphers() {
+  var argv = process.execArgv;
+  if ((argv.indexOf('--enable-legacy-cipher-list=v0.10.38') > -1 ||
+       process.env['NODE_LEGACY_CIPHER_LIST'] === 'v0.10.38') &&
+      DEFAULT_CIPHERS === _crypto.getLegacyCiphers('v0.10.38')) {
+        return true;
+  }
+  return false;
+}
+exports.usingV1038Ciphers = usingV1038Ciphers;
```
misterdjules: I'm wondering if we could find a better place to add that as a public API, or even better _not_ make this a public API.

misterdjules:
Worth noting: 

1. `test/external/ssl-options/test.js` passes (it was updated in this PR, so one thing we want to make sure is that the changes to these tests make sense and do not hide potential regressions).

2. Local `make test` passes on OSX with no regression.

3. Tests on all other supported platforms are running on Jenkins for [UNIX](http://jenkins.nodejs.org/job/node-review-unix/) and [Windows](http://jenkins.nodejs.org/job/node-review-windows/).

misterdjules:
@jasnell Thank you very much! We're getting there :+1:

misterdjules:
Confirmed no regression on [UNIX](http://jenkins.nodejs.org/job/node-review-unix/338/) or [Windows](http://jenkins.nodejs.org/job/node-review-windows/200/).

lib/tls.js
jasnell: @misterdjules ... Yes, my original intent was for this to be internal only but given the need to expose this is crypto.js also, this was the quick fix. Perhaps what we need is to push the usingV1038Ciphers into the `process.binding('crypto')` API instead.

jasnell:
btw, given that we're now on `v0.10.40`, the references to `V1038` should be updated.

lib/crypto.js
misterdjules: @jasnell Actually, this is specific to v0.10.x versions, so we won't have to maintain it in the v0.12.x branch and later. If we're on the same page regarding this then I'm fine with it.

lib/tls.js
```diff
@@ -1335,6 +1338,16 @@ exports.connect = function(/* [port, host], options, cb */) {
   var defaults = {
     rejectUnauthorized: '0' !== process.env.NODE_TLS_REJECT_UNAUTHORIZED
   };
+  if (!process._usingV1040Ciphers()) {
```
misterdjules: @jasnell I don't think that this change is needed given that we have https://github.com/joyent/node/pull/25603/files#diff-437b078f0e3aa5b13790315b3813fd4fR138 too. Or am I missing something?

That would have the nice side effect of not needing to expose `_usingV1040Ciphers` publicly.

lib/crypto.js
```diff
@@ -134,7 +133,18 @@ exports.createCredentials = function(options, context) {
 
   if (options.cert) c.context.setCert(options.cert);
 
-  if (options.ciphers) c.context.setCiphers(options.ciphers);
+  if (options.ciphers) {
+    c.context.setCiphers(options.ciphers);
+  } else if (!(process._usingV1040Ciphers() && options.ciphers === undefined)) {
```
misterdjules: @jasnell Per https://github.com/joyent/node/pull/23947#issuecomment-101849036, we could replace this with:
```
} else if (!(process._usingV1040Ciphers()) {
```
since we said that `null` and `undefined`would mean "use the default ciphers list`. That would also mean that the empty string would make node use the empty string unless the v0.10.40 legacy cipher is used, but that also sounds reasonable.

misterdjules:
@jasnell Added another round of comments, and pushed a few cleanup commits: https://github.com/misterdjules/node/commit/9783a82ac3c5e739b43d79008de53b852e9c83f5, https://github.com/misterdjules/node/commit/bd74d90a1adcb51f27f32a7d30a3996793d40335 and https://github.com/misterdjules/node/commit/ca77baeaf4a97c97b3d30bdb02b8c8fa4724ca27.

jasnell:
@misterdjules ... sorry for the delay on reviewing. This LGTM.

jasnell:
@misterdjules ... @mhdawson and I were discussing this change further. At this point, given the issues, we may want to just back off on the legacy cipher list switch and omit that entirely as it seems to be the most contentious piece. While having the legacy switch there helped us out with out particular situation, it's not critical and does increase the potential maintenance cost. So, with that in mind, let's split this into separate PRs: one that changes the default cipher list to remove RC4 in v0.10 and one that adds only the `--cipher-list` switch for overriding the default. I will also open a new PR against nodejs/node master that adds the `--cipher-list` switch to master.

src/node.cc
```diff
@@ -2566,6 +2566,8 @@ static void PrintHelp() {
          "  --max-stack-size=val set max v8 stack size (bytes)\n"
          "  --enable-ssl2        enable ssl2\n"
          "  --enable-ssl3        enable ssl3\n"
+         "  --cipher-list=val    specify the default TLS cipher list\n"
+         "  --enable-legacy-cipher-list=v0.10.40 \n"
```
thefourtheye: Extra space before \n

src/node.cc
```diff
@@ -2577,6 +2579,8 @@ static void PrintHelp() {
          "NODE_MODULE_CONTEXTS   Set to 1 to load modules in their own\n"
          "                       global contexts.\n"
          "NODE_DISABLE_COLORS    Set to 1 to disable colors in the REPL\n"
+         "NODE_CIPHER_LIST       Override the default TLS cipher list\n"
+         "NODE_LEGACY_CIPHER_LIST=v0.10.40\n"
```
thefourtheye: Shouldn't the description explain what this does?

src/node_crypto.cc
```diff
@@ -4190,6 +4197,43 @@ static void array_push_back(const TypeName* md,
...
+// (see http://v8.googlecode.com/svn/trunk/samples/shell.cc)
+const char* ToCString(const node::Utf8Value& value) {
+  return *value ? *value : "<string conversion failed>";
+}
+
+const char* LegacyCipherList(const char * ver) {
+  if (ver == NULL) {
+    return NULL;
+  }
+  if (strncmp(ver, "v0.10.40", 8) == 0) {
```
thefourtheye: In other places in the PR, length + 1 is used

jasnell:
@thefourtheye... This entire PR is likely to be refactored significantly.

jasnell:
Specifically, the legacy cipher list switch will probably be dropped entirely.

thefourtheye:
@jasnell Ah, good that you told me :-)

jasnell:
@misterdjules ... the `--tls-cipher-list` switch has landed in nodejs/node master. Let's close this PR here and open a new one that just does the RC4 deprecation. We can backport the `--tls-cipher-list` change separately. We can drop the legacy switch entirely at this point. Going to go ahead and close but reopen if you feel it's necessary

PR #25610, opened 2015-07-01
configure: fix empty prefix

### LINKED ISSUE #5197: configure script fails with empty prefix
``` sh
./configure --prefix=
```

is the canonical way to use the prefix "/" without getting nasty double-slashes everywhere.

with node's configure, it will silently create /local directory structure in the DESTDIR.

rofl0r:
autoconf-based configure scripts (which this one is modeled after)
support to pass an empty prefix via --prefix=.
this is required for systems that have binaries directly in /bin
rather than /usr/bin etc (one example is sabotage linux).

the fix is easy: rather than testing options.prefix for "if var",
which treats an empty string the same as "None", we check for
"if var not None".

fixes #5197

### DIFF
--- a/configure
+++ b/configure
@@ -1020,7 +1020,7 @@ config = {
   'PYTHON': sys.executable,
 }
 
-if options.prefix:
+if options.prefix != None:
   config['PREFIX'] = options.prefix
 
 config = '\n'.join(map('='.join, config.iteritems())) + '\n'


jasnell:
@orangemocha and @misterdjules ... thoughts?

orangemocha:
LGTM

thefourtheye:
We should use `is not` to check the None type. Please refer [this Stackoverflow answer](http://stackoverflow.com/a/23086405/1903116)

rofl0r:
@thefourtheye this seems to be only a stylistic difference, since both inequality tests evaluate to the same result.

orangemocha:
Will land this on v0.10 first.

thefourtheye:
I am sorry, if my vote matters, -1 as it is.

orangemocha:
@rofl0r mind to make the change?

orangemocha:
... and open a new PR against v0.10?

rofl0r:
branch v0.10 is missing the snippet the patch applies to and i don't know how to fix it there :/

orangemocha:
That snipped was added in https://github.com/joyent/node/commit/85e4fc4306055656f0acab0052ae6da73bfbeb1f

Doess adding the following work?

```
if options.prefix is not None:
   config['PREFIX'] = options.prefix
```

PR #25635, opened 2015-07-06
doc: several documentation improvements

jasnell:
bundled together to make landing them easier.

### DIFF
--- a/doc/api/stream.markdown
+++ b/doc/api/stream.markdown
@@ -228,7 +228,7 @@ returns it.  If there is no data available, then it will return
 
 If you pass in a `size` argument, then it will return that many
 bytes.  If `size` bytes are not available, then it will return `null`,
-unless we've ended, in which case it will return the data remaining 
+unless we've ended, in which case it will return the data remaining
 in the buffer.
 
 If you do not specify a `size` argument, then it will return all the
@@ -250,6 +250,9 @@ readable.on('readable', function() {
 
 If this method returns a data chunk, then it will also trigger the
 emission of a [`'data'` event][].
+
+Note that calling `readable.read([size])` after the `end` event has been
+triggered will return `null`. No runtime error will be raised.
 
 #### readable.setEncoding(encoding)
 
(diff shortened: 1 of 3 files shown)


jasnell:
@joyent/node-collaborators

doc/api/stream.markdown
```diff
@@ -167,6 +167,33 @@ again when more data is available.
...
+data to be consumed. The callback passed to handle the `'readable'`
+event must be prepared to handle a `null` response from
+`readable.read([size])`. For instance, in the following example, `foo.txt`
+is an empty file:
+
+```javascript
+var fs = require('fs');
+var rr = fs.createReadStream('foo.txt');
+rr.on('readable', function() {
+  var chunk;
```
trevnorris: Is this variable only for demonstration? Not being used.

trevnorris:
One question. Otherwise LGTM

doc/api/stream.markdown
jasnell: Ah...right, just mentally zoned on that one ;-). I'll remove the variable really quick

doc/api/stream.markdown
```diff
@@ -167,6 +167,32 @@ again when more data is available.
 The `readable` event is not emitted in the "flowing" mode with the
 sole exception of the last one, on end-of-stream.
 
+Note that the `'readable'` event indicates only that data *can* be
+read from the stream. It does not indicate whether there is actual
+data to be consumed. The callback passed to handle the `'readable'`
```
chrisdickinson: The first two sentences are pretty confusing. Maybe:

> The `'readable'` event indicates that the stream has new information: either new data is available or the end of the stream has been reached. In the former case, `.read()` will return that data. In the latter case, `.read()` will return `null`. For instance, in the following example, ...

jasnell: :+1: will make that change

jasnell: @chrisdickinson ... done... if it looks good to you I'll get it landed

jasnell:
Landed in f91fa52, f7af915, 9faae7a, 423c433, daed421, 340e9f0 and 6036e4f

PR #25638, opened 2015-07-07
build: make release process usable by multiple users

misterdjules:
This is a first step in making the release process easier to use by multiple users.

Instead of using `/home/tj` on `nodejs.org` as a staging directory, it uses `/home/staging`, which we also can control access to via ssh keys.

Instead of pushing tags and branches directly to joyent/node, it pushes them to the personal fork of the person doing the release, so there's a chance to fix errors and mistakes before pushing tags and releases branches to the reference repository.

Finally it also now distinguishes between stable, maintenance and unstable versions to mirror the current v0.10/v0.12/v0.13 release lines.

These changes are in sync with the latest changes I made recently to the [nodejs-release Jenkins job](http://jenkins.nodejs.org/job/nodejs-release/) and [Node.js release guide](https://github.com/joyent/node/wiki/Node.js-release-guide).

### DIFF
--- a/tools/getstability.py
+++ b/tools/getstability.py
@@ -1,4 +1,4 @@
-import os,re
+import os, re, sys
 
 node_version_h = os.path.join(os.path.dirname(__file__), '..', 'src',
     'node_version.h')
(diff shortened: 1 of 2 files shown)


sam-github:
I haven't run this, but the change LGTM. I'll give it a run next week.

cjihrig:
LGTM too from a glance

misterdjules:
Landed in ec861f6f90dc1c28f8d28231bac03c06ad85238e, thank you @sam-github and @cjihrig :+1:

PR #25640, opened 2015-07-07
win,msi: change InstallScope to perMachine

joaocgreis:
Currently, installing with the MSI always asks for Administrator permissions, as mentioned in #7629 . Explicitly installing per machine only adds the benefit that shortcuts are placed in ProgramData thus being accessible to all users and also solving #5849 .

The second commit moves the installation path in the registry to Local Machine. The registry stores HKLM in different places for 32 and 64 bit applications, so the installer will not suggest the old path when upgrading from 32 to 64 bit version, solving #5592 and #25087 .

/cc @piscisaureus

### DIFF
--- a/test/simple/simple.status
+++ b/test/simple/simple.status
@@ -14,6 +14,7 @@ test-fs-readfile-error            : PASS,FLAKY
 test-net-GH-5504                  : PASS,FLAKY
 test-stdin-script-child           : PASS,FLAKY
 test-util-debug                   : PASS,FLAKY
+test-signal-unregister            : PASS,FLAKY
 
 [$system==macos]
 test-fs-watch                     : PASS,FLAKY
--- a/tools/msvs/msi/custom_actions.c
+++ b/tools/msvs/msi/custom_actions.c
@@ -1,9 +1,59 @@
-
 #define WIN32_LEAN_AND_MEAN
 
 #include <windows.h>
 #include <msiquery.h>
 #include <wcautil.h>
+
+#define GUID_BUFFER_SIZE 39 // {8-4-4-4-12}\0
+
+
+UINT WINAPI SetInstallScope(MSIHANDLE hInstall) {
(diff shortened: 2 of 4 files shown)


orangemocha:
LGTM, but would still be good if someone else (@piscisaureus ?) could take a look.

piscisaureus:
The problem I see with this is the following:

In order for NPM to work, `%appdata%\npm` needs to be added to the path.
That path is user-specific.
Setting the installation scope to global would add the wrong path.

joaocgreis:
@piscisaureus Thanks for reviewing. In all my tests, the `%appdata%\npm` still gets added to the user path, but only for the user that installs. This is the same behavior that happens without these changes. Do you have a specific situation in mind in which the wrong path is added?

orangemocha:
> In order for NPM to work, %appdata%\npm needs to be added to the path.

@piscisaureus , thanks for reviewing! As @joaocgreis said, %appdata%\npm still gets added to the user path as before. OK to land this?

piscisaureus:
> @piscisaureus , thanks for reviewing! As @joaocgreis said, %appdata%\npm still gets added to the user path as before. OK to land this?

If that's true, then yes.

orangemocha:
Triple-checked that it does. Will land shortly...

orangemocha:
FYI, adding a commit to this PR mark test-signal-unregister as flaky. It failed on Ubuntu and that's obviously unrelated to this change:
https://jenkins-iojs.nodesource.com/job/node-test-commit-linux/28/nodes=ubuntu1504-64/tapTestReport/test.tap-576/

orangemocha:
Actually, I will mark that test as flaky in a separate PR.

orangemocha:
Landed in 14db629497d0eb97a5f0fbb70cf3e534e460deec.

orangemocha:
This should probably go in the converged repo. @jasnell what do we need to do to make it so?

PR #25641, opened 2015-07-07
Add support for building on ARM

ghost:
This patch was sourced from a FreeBSD PR

Source: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=201227
Author: Mikael Urankar

### DIFF
--- a/deps/v8/src/arm/assembler-arm.cc
+++ b/deps/v8/src/arm/assembler-arm.cc
@@ -50,6 +50,35 @@ bool CpuFeatures::initialized_ = false;
 unsigned CpuFeatures::supported_ = 0;
 unsigned CpuFeatures::found_by_runtime_probing_ = 0;
 
+#ifdef __arm__
+
+bool OS::ArmCpuHasFeature(CpuFeature feature) {
+  return false;
+}
+
+CpuImplementer OS::GetCpuImplementer() {
+  static bool use_cached_value = false;
+  static CpuImplementer cached_value = UNKNOWN_IMPLEMENTER;
+  if (use_cached_value) {
+    return cached_value;
+  }
+  cached_value = ARM_IMPLEMENTER;
+
+  use_cached_value = true;
+  return cached_value;
+}
+
+
+bool OS::ArmUsingHardFloat() {
+#if defined(__ARM_PCS_VFP)
+  return true;
+#else
+  return false;
+#endif
+}
+
+#endif  // def __arm__
+
 
 // Get the CPU features enabled by the build. For cross compilation the
 // preprocessor symbols CAN_USE_ARMV7_INSTRUCTIONS and CAN_USE_VFP3_INSTRUCTIONS
@@ -749,7 +778,7 @@ static bool fits_shifter(uint32_t imm32,
                          Instr* instr) {
   // imm32 must be unsigned.
   for (int rot = 0; rot < 16; rot++) {
-    uint32_t imm8 = (imm32 << 2*rot) | (imm32 >> (32 - 2*rot));
+    uint32_t imm8 = rot == 0 ? imm32 : ((imm32 << 2*rot) | (imm32 >> (32 - 2*rot)));
     if ((imm8 <= 0xff)) {
       *rotate_imm = rot;
       *immed_8 = imm8;
--- a/deps/v8/src/arm/cpu-arm.cc
+++ b/deps/v8/src/arm/cpu-arm.cc
@@ -64,7 +64,7 @@ void CPU::FlushICache(void* start, size_t size) {
   // None of this code ends up in the snapshot so there are no issues
   // around whether or not to generate the code when building snapshots.
   Simulator::FlushICache(Isolate::Current()->simulator_i_cache(), start, size);
-#else
+#elif defined(__linux__)
   // Ideally, we would call
   //   syscall(__ARM_NR_cacheflush, start,
   //           reinterpret_cast<intptr_t>(start) + size, 0);
@@ -103,6 +103,10 @@ void CPU::FlushICache(void* start, size_t size) {
         : "0" (beg), "r" (end), "r" (flg), "r" (__ARM_NR_cacheflush)
         : "r3");
   #endif
+#elif defined(__FreeBSD__)
+  __clear_cache(start, reinterpret_cast<char*>(start) + size);
+#else
+#error "No cache flush implementation on this platform"
 #endif
 }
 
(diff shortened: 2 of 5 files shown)


deps/v8/src/arm/assembler-arm.cc
```diff
@@ -50,6 +50,35 @@ bool CpuFeatures::initialized_ = false;
 unsigned CpuFeatures::supported_ = 0;
 unsigned CpuFeatures::found_by_runtime_probing_ = 0;
 
+#ifdef __arm__
+
+bool OS::ArmCpuHasFeature(CpuFeature feature) {
```
misterdjules: Why defining these functions in the ARM assembler for all OSes? Wouldn't that change break ARM support on platforms other than the one for which it was specifically written?

deps/v8/src/arm/assembler-arm.cc
```diff
@@ -50,6 +50,35 @@ bool CpuFeatures::initialized_ = false;
...
+  return false;
+}
+
+CpuImplementer OS::GetCpuImplementer() {
+  static bool use_cached_value = false;
+  static CpuImplementer cached_value = UNKNOWN_IMPLEMENTER;
+  if (use_cached_value) {
+    return cached_value;
+  }
+  cached_value = ARM_IMPLEMENTER;
```
misterdjules: Does FreeBSD support querying the implementer code instead of hardcoding it?

deps/v8/src/arm/assembler-arm.cc
```diff
@@ -749,7 +778,7 @@ static bool fits_shifter(uint32_t imm32,
                          Instr* instr) {
   // imm32 must be unsigned.
   for (int rot = 0; rot < 16; rot++) {
-    uint32_t imm8 = (imm32 << 2*rot) | (imm32 >> (32 - 2*rot));
+    uint32_t imm8 = rot == 0 ? imm32 : ((imm32 << 2*rot) | (imm32 >> (32 - 2*rot)));
```
misterdjules: Is this change really needed? It seems that both shift operations would zero-fill when applied to a `uint32_t`, and thus when `rot == 0`, this operation would be equivalent to `imm32 | 0`, or `imm32`, but I may be missing something.

deps/v8/src/arm/cpu-arm.cc
```diff
@@ -64,7 +64,7 @@ void CPU::FlushICache(void* start, size_t size) {
   // None of this code ends up in the snapshot so there are no issues
   // around whether or not to generate the code when building snapshots.
   Simulator::FlushICache(Isolate::Current()->simulator_i_cache(), start, size);
-#else
+#elif defined(__linux__)
```
misterdjules: Would using `#elif !defined(__FreeBSD__)` be more appropriate here? If I understand correctly, before this change this code is compiled for all platforms where `deps/v8/src/arm/cpu-arm.cc` is compiled, and thus not necessarily only on Linux.

deps/v8/src/atomicops.h
```diff
@@ -160,8 +160,10 @@ Atomic64 Release_Load(volatile const Atomic64* ptr);
 #elif defined(__GNUC__) && \
   (defined(V8_HOST_ARCH_IA32) || defined(V8_HOST_ARCH_X64))
 #include "atomicops_internals_x86_gcc.h"
-#elif defined(__GNUC__) && defined(V8_HOST_ARCH_ARM)
+#elif defined(__GNUC__) && defined(__linux__) && defined(V8_HOST_ARCH_ARM)
```
misterdjules: Same comment as [above](https://github.com/joyent/node/pull/25641/files#r34118403).

deps/v8/src/platform-freebsd.cc
```diff
@@ -415,7 +415,7 @@ void* VirtualMemory::ReserveRegion(size_t size) {
   void* result = mmap(OS::GetRandomMmapAddr(),
                       size,
                       PROT_NONE,
-                      MAP_PRIVATE | MAP_ANON | MAP_NORESERVE,
+                      MAP_PRIVATE | MAP_ANON,
```
misterdjules: For the record, this seems to be a backport of https://codereview.chromium.org/1025823003/, is that correct?

misterdjules:
@feld Thank you for your interest in running node v0.10 on ARMv6 and FreeBSD. I added a few comments, mostly so that I can understand some of the details in this patch.

Let's see what this patch would look like at the end of the review process. At that time we'll be in a better place to evaluate whether it could land in this repository or should be kept as a patch in FreeBSD's ports.

deps/v8/src/arm/assembler-arm.cc
MIkaelU: This is an upstream patch (we are using clang on FreeBSD):
https://codereview.chromium.org/979633002/

deps/v8/src/platform-freebsd.cc
MIkaelU: Correct

deps/v8/src/arm/cpu-arm.cc
MIkaelU: From my understanding __ARM_NR_cacheflush is linux specific

deps/v8/src/atomicops.h
MIkaelU: code in atomicops_internals_arm_gcc.h is really linux/arm specific

deps/v8/src/arm/cpu-arm.cc
misterdjules: Ok, makes sense :+1:

deps/v8/src/atomicops.h
misterdjules: Right, sounds good :+1:

misterdjules:
@MIkaelU Thank you very much for the feedback! :+1: Any comment on https://github.com/joyent/node/pull/25641#discussion_r34117960 and https://github.com/joyent/node/pull/25641#discussion_r34118231?

feld:
good collaboration guys :+1:

deps/v8/src/arm/assembler-arm.cc
MIkaelU: I've screwed up this part of the patch... These changes should go in deps/v8/src/platform-freebsd.cc

deps/v8/src/arm/assembler-arm.cc
MIkaelU: No

jasnell:
@misterdjules ... is this something we're going to land in v0.10?

rvagg:
@nodejs/lts & @nodejs/build I'm inclined to say we shouldn't land this given that v4+ has proper ARM support and adding it back to v0.10 is just adding additional maintenance burden, thoughts?

jasnell:
Agreed

cjihrig:
I'd say this isn't appropriate for LTS or maintenance releases.

rvagg:
done

sorry @feld, hopefully you're sorted on v4.x now though

PR #25642, opened 2015-07-07
tls: Closing parent socket also closes the tls sock

### LINKED ISSUE #25365: Access violation in nodejs when restarting a listening server

dnakamura:
Fixes #25365 .  At the moment if you close the parent socket of a a TLS socket the TLS socket is not aware of this and will segfault when you attempt to perform and IO operation on it.  Further, if you close the parent socket and then the TLS socket, the parent socket will emit the close event twice (and the TLS socket wont fire at all). If you change the order the results are reversed, ie. closing the TLS then the parent socket will result in the TLS socket firing the close event twice and the parent socket wont fire at all. 

This patch moves toward linking the parent and tls sockets. Calling `destroy()` on either of the the sockets causes both sockets to be destroyed and the close event emitted for both of them

### DIFF
--- a/lib/_tls_wrap.js
+++ b/lib/_tls_wrap.js
@@ -228,6 +228,7 @@ function onocspresponse(resp) {
 function TLSSocket(socket, options) {
   // Disallow wrapping TLSSocket in TLSSocket
   assert(!(socket instanceof TLSSocket));
+  var self = this;
 
   net.Socket.call(this, {
     handle: socket && socket._handle,
@@ -238,6 +239,9 @@ function TLSSocket(socket, options) {
 
   if (socket) {
     this._parent = socket;
+    socket._destroy = function(exception) {
+      self._destroy(exception);
+    };
     // To prevent assertion in afterConnect()
     this._connecting = socket._connecting;
   }
--- a/lib/net.js
+++ b/lib/net.js
@@ -468,12 +468,10 @@ Socket.prototype._destroy = function(exception, cb) {
     return;
   }
 
-  self._connecting = false;
-
-  this.readable = this.writable = false;
-
-  for (var s = this; s !== null; s = s._parent)
+  for (var s = this; s !== null; s = s._parent) {
     timers.unenroll(s);
+    s._connecting = s.readable = s.writable = false;
+  }
 
   debug('close');
   if (this._handle) {
@@ -482,16 +480,20 @@ Socket.prototype._destroy = function(exception, cb) {
     var isException = exception ? true : false;
     this._handle.close(function() {
       debug('emit close');
-      self.emit('close', isException);
+      for (var s = self; s !== null; s = s._parent)
+        s.emit('close', isException);
     });
     this._handle.onread = noop;
-    this._handle = null;
+    for (var s = this; s !== null; s = s._parent)
+      s._handle = null;
   }
 
   // we set destroyed to true before firing error callbacks in order
   // to make it re-entrance safe in case Socket.prototype.destroy()
   // is called within callbacks
-  this.destroyed = true;
+  for (var s = this; s !== null; s = s._parent)
+    s.destroyed = true;
+
   fireErrorCallbacks();
 
   if (this.server) {
--- a/test/simple/test-tls-destroy-socket.js
+++ b/test/simple/test-tls-destroy-socket.js
@@ -1,0 +1,55 @@
+if (!process.versions.openssl) {
+  console.error('Skipping because node compiled without OpenSSL.');
+  process.exit(0);
+}
+
+var common = require('../common');
+var assert = require('assert');
+var tls = require('tls');
+var fs = require('fs');
+
+var options = {
+  key: fs.readFileSync(common.fixturesDir + '/keys/agent2-key.pem'),
+  cert: fs.readFileSync(common.fixturesDir + '/keys/agent2-cert.pem')
+};
+
+var normalSock = null;
+var hasError = false;
+var tlsCloseCount = 0;
+var sockCloseCount = 0;
+
+var server = tls.createServer(options, function(secureSock) {
+  secureSock.on('error', function() {
+    hasError = true;
+  });
+  secureSock.on('close', function() {
+    tlsCloseCount++;
+  });
+  normalSock.on('close', function() {
+    sockCloseCount++;
+  });
+
+  normalSock.destroy();
+  secureSock.write('Test!', function(err) {
+    assert(err);
+  });
+});
+
+server.on('connection', function(sock) {
+  normalSock = sock;
+});
+
+server.listen(common.PORT, function() {
+  var c = tls.connect(common.PORT, {rejectUnauthorized: false});
+  c.on('error', function() {}); // ignore socket hangup error
+
+  c.on('end', function() {
+    server.close();
+  });
+
+  process.on('exit', function() {
+    assert(hasError);
+    assert.equal(tlsCloseCount, 1);
+    assert.equal(sockCloseCount, 1);
+  });
+});


lib/_tls_wrap.js
```diff
@@ -238,6 +238,7 @@ function TLSSocket(socket, options) {
 
   if (socket) {
     this._parent = socket;
+    socket._destroy = function(exception){self._destroy(exception)};
```
thefourtheye: Why not simply `this._destroy.bind(this);`?

lib/net.js
```diff
@@ -482,16 +482,22 @@ Socket.prototype._destroy = function(exception, cb) {
...
+        s._handle = null;
+    }
   }
 
   // we set destroyed to true before firing error callbacks in order
   // to make it re-entrance safe in case Socket.prototype.destroy()
   // is called within callbacks
-  this.destroyed = true;
+  for(var s = this; s !== null; s = s._parent){
+    s.destroyed = true;
```
thefourtheye: This can be moved to the loop above this.

lib/_tls_wrap.js
dnakamura: I'm newish to javascript and I didn't know that was a thing

mhdawson:
@indutny could you review this as well since one of your recent commits ( https://github.com/dnakamura/node/commit/bada87bd666f103cc00d63ace48ef300b0caaabf#diff-04c3a6bcf355f2e05e7700c1b253d475) looks like it addressed a similar parent/child related issue.

lib/net.js
```diff
@@ -456,7 +456,8 @@ Socket.prototype._destroy = function(exception, cb) {
     if (cb) cb(exception);
     if (exception && !self._writableState.errorEmitted) {
       process.nextTick(function() {
-        self.emit('error', exception);
+        for(var s = self; s !== null; s = s._parent)
```
indutny: `for (`

indutny:
@dnakamura I wonder if we may consider backporting https://github.com/nodejs/io.js/commit/9b6b05556 instead?

dnakamura:
@induty As far as I can tell, nodejs/io.js@9b6b055 does not fix the problem (still able to reproduce on latest io.js). Also I have to double check the code but I think it may already be backported

lib/net.js
```diff
@@ -482,16 +482,22 @@ Socket.prototype._destroy = function(exception, cb) {
     var isException = exception ? true : false;
     this._handle.close(function() {
       debug('emit close');
-      self.emit('close', isException);
+      for(var s = self; s !== null; s = s._parent)
+        s.emit('close', isException);
     });
     this._handle.onread = noop;
-    this._handle = null;
+    for(var s = this; s !== null; s = s._parent){
```
indutny: `for (...)`

lib/net.js
```diff
@@ -482,16 +482,22 @@ Socket.prototype._destroy = function(exception, cb) {
...
+        s._handle = null;
+    }
```
indutny: No need in braces.

lib/net.js
```diff
@@ -482,16 +482,22 @@ Socket.prototype._destroy = function(exception, cb) {
...
   }
 
   // we set destroyed to true before firing error callbacks in order
   // to make it re-entrance safe in case Socket.prototype.destroy()
   // is called within callbacks
-  this.destroyed = true;
+  for(var s = this; s !== null; s = s._parent){
+    s.destroyed = true;
+  }
```
indutny: Ditto.

lib/_tls_wrap.js
indutny: I would do:

```javascript
function(err) {
  self._destroy(err);
}
```

Last time I check it - `bind` was kind of slow.

indutny:
@dnakamura you are absolutely right. It just looked strikingly similar, so I thought it might need the backport. Looked through the changes and they are looking good to me except few nits.

lib/_tls_wrap.js
thefourtheye: @indutny Oh, okay. Thanks for the info :-)

dnakamura:
@indutny  , @thefourtheye  took your changes into account, can you take another look. 
(also, I added another commit, so that I wouldn't destroy your comments on the original commit. Will squash it when everybody agrees fix is good.)

lib/net.js
```diff
@@ -468,12 +469,11 @@ Socket.prototype._destroy = function(exception, cb) {
...
-  self._connecting = false;
-
-  this.readable = this.writable = false;
-
-  for (var s = this; s !== null; s = s._parent)
-    timers.unenroll(s);
+  for (var s = this; s !== null; s = s._parent){
+     timers.unenroll(s);
+     s._connecting = false;
+     s.readable = s.writable = false;
```
thefourtheye: If we are writing like this, why not just

    s._connecting = s.readable = s.writable = false;

then?

test/simple/test-tls-destroy-socket.js
```diff
@@ -0,0 +1,55 @@
...
+var hasError = false;
+var tlsCloseCount = 0;
+var sockCloseCount = 0;
+
+var server = tls.createServer(options, function(secureSock) {
+  secureSock.on('error', function() {
+    hasError = true;
+  });
+  secureSock.on('close', function() {
+    tlsCloseCount ++;
```
thefourtheye: mmm, didn't the linter complain about the space between `++` and `tlsCloseCount`?

lib/net.js
```diff
@@ -482,16 +482,22 @@ Socket.prototype._destroy = function(exception, cb) {
...
       debug('emit close');
-      self.emit('close', isException);
+      for (var s = self; s !== null; s = s._parent)
+        s.emit('close', isException);
     });
     this._handle.onread = noop;
-    this._handle = null;
+    for (var s = this; s !== null; s = s._parent)
+        s._handle = null;
+
```
indutny: Excessive newline.

lib/_tls_wrap.js
indutny: This one is still there ;)

lib/net.js
```diff
@@ -482,16 +482,22 @@ Socket.prototype._destroy = function(exception, cb) {
...
   }
 
   // we set destroyed to true before firing error callbacks in order
   // to make it re-entrance safe in case Socket.prototype.destroy()
   // is called within callbacks
-  this.destroyed = true;
+  for (var s = this; s !== null; s = s._parent)
+    s.destroyed = true;
+
+
```
indutny: One excessive newline.

indutny:
LGTM, except some style nits. May I ask you to fix them before I will land this PR?

_(It is going to be fun to forward-port it to io.js. I should probably do it myself right after landing it in v0.12)_

dnakamura:
@indutny  Fixed style nits, can you take a look?

lib/_tls_wrap.js
indutny: This one wasn't fixed. Please make it span multiple lines.

lib/net.js
```diff
@@ -468,12 +469,10 @@ Socket.prototype._destroy = function(exception, cb) {
     return;
   }
 
-  self._connecting = false;
-
-  this.readable = this.writable = false;
-
-  for (var s = this; s !== null; s = s._parent)
-    timers.unenroll(s);
+  for (var s = this; s !== null; s = s._parent){
```
indutny: Space between `)` and `{`. Didn't `make lint` complain about it?

lib/net.js
```diff
@@ -482,16 +481,20 @@ Socket.prototype._destroy = function(exception, cb) {
...
     this._handle.close(function() {
       debug('emit close');
-      self.emit('close', isException);
+      for (var s = self; s !== null; s = s._parent)
+        s.emit('close', isException);
     });
     this._handle.onread = noop;
-    this._handle = null;
+    for (var s = this; s !== null; s = s._parent)
+        s._handle = null;
```
indutny: two space indent, not 4 space, please.

indutny:
Few last nits, thanks!

dnakamura:
@indutny ok that should be all of them

lib/_tls_wrap.js
```diff
@@ -238,6 +238,9 @@ function TLSSocket(socket, options) {
 
   if (socket) {
     this._parent = socket;
+    socket._destroy = function(exception){
```
indutny: `function(exception) {`, please

lib/net.js
```diff
@@ -468,12 +469,10 @@ Socket.prototype._destroy = function(exception, cb) {
...
   }
 
-  self._connecting = false;
-
-  this.readable = this.writable = false;
-
-  for (var s = this; s !== null; s = s._parent)
-    timers.unenroll(s);
+  for (var s = this; s !== null; s = s._parent) {
+     timers.unenroll(s);
```
indutny: Please use two space indent everywhere, it is three space here right now.

indutny:
Haha! Sorry @dnakamura but there are few more style nits :) Other than that LGTM! Thanks!

dnakamura:
@indutny  ok, hopefully now I've got it :smile:

lib/_tls_wrap.js
```diff
@@ -238,6 +239,9 @@ function TLSSocket(socket, options) {
 
   if (socket) {
     this._parent = socket;
+    socket._destroy = function(exception) {
+      self._destroy(exception)
```
thefourtheye: Semicolon at the end?

test/simple/test-tls-destroy-socket.js
```diff
@@ -0,0 +1,55 @@
...
+server.listen(common.PORT, function() {
+  var c = tls.connect(common.PORT, {rejectUnauthorized: false});
+  c.on('error', function() {}); // ignore socket hangup error
+
+  c.on('end', function() {
+    server.close();
+  });
+
+  process.on('exit', function() {
+    assert(hasError)
```
thefourtheye: Semicolon at the end.

dnakamura:
@thefourtheye  fixed.

indutny:
Awesome LGTM. @thefourtheye may I have your LGTM as well? ;)

thefourtheye:
@indutny Changes are straight forward. LGTMT :)

indutny:
@mhdawson @thefourtheye where should I land this?

thefourtheye:
@indutny I am not familiar with this repo's pr landing process but going by the tags on this PR I would say v0.12

jasnell:
I can get it landed. FWIW, we use the node-accept-pull-request job in https://jenkins-iojs.nodesource.com/ to land in joyent/node. @orangemocha has all the necessary details on that.

thefourtheye:
@jasnell Thanks for the info :-)

jasnell:
@thefourtheye ... btw, I can't seem to find your full name anywhere for the sign off ;-)

thefourtheye:
@jasnell Its in [nodejs/node's collaborator's list](https://github.com/nodejs/node/blob/master/README.md#collaborators). Sakthipriyan Vairamani :-) It is nice working with you :-)

jasnell:
The pleasure is mine.
Ok, job kicked off... assuming the CI is green, the commit will land (https://jenkins-iojs.nodesource.com/job/node-accept-pull-request/36/)

jasnell:
@dnakamura ... if you're seeing this problem on io.js master also, can I ask for you to go ahead and port this and open a new PR against master on nodejs/node if you haven't done so already? :-)

jasnell:
Hmm.. @dnakamura getting some failures in CI. Not clear if they are related to the change. Same failure on both Windows platforms. https://jenkins-iojs.nodesource.com/job/node-test-commit-windows/nodes=win2008r2/181/

dnakamura:
@jasnell: yup it does seem to be me (except test-dns), looking into the root cause now

dnakamura:
So as it turns out the issue is because now if the socket is destroyed from an error, both sockets (tls, and parent) emit an error. Since nobody is listening for errors on the parent socket it causes an unhandled error event. My first instinct is to only change the code so it only fires the error event on the lowest socket (from a parent-child view, highest from a protocol stack view). However I'm wondering if it might make more sense to always fire the error event on the same object ```_destroy()``` was called on. Thoughts?

indutny:
@dnakamura or we may call it a breaking change, land it in nodejs/node repo. And watch for the `error` events ourselves. Maybe this is better?

indutny:
Or just not emit errors on parent socket in v0.12. Technically, emitting `close` should be enough to let it handle socket state change.

dnakamura:
@jasnell Ok that should fix it now. I have a couple tests still failing here, but they also fail on official builds, so I dont think its my fault

jasnell:
As long as it passes CI, LGTM. I'll run it through the checker now.

jasnell:
Landed in 0363cf4

PR #25645, opened 2015-07-08
openssl: arm64 compile fix

obastemur:
Compiling openssl was failing on arm64 platform.

### DIFF
--- a/deps/openssl/config/opensslconf.h
+++ b/deps/openssl/config/opensslconf.h
@@ -205,7 +205,7 @@
  /* If this is set to 'unsigned int' on a DEC Alpha, this gives about a
   * %20 speed up (longs are 8 bytes, int's are 4). */
 # undef DES_LONG
-# if defined(_M_X64) || defined(__x86_64__) || defined(__arm__) || defined(__mips__)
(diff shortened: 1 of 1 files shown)


jasnell:
Given that this changes openssl, wouldn't it be better to patch this upstream? Is this error specific to node's use of openssl or a general problem with openssl?

obastemur:
AFAIK this config file is customized anyways.

jasnell:
@shigeki ... would you be able to take a look at this when you've got a moment?

shigeki:
I don't think node-v0.12.x supports arm64 yet and adding a support of a new arch to node-v0.12 is not a good idea to me. 
````
  --dest-cpu=DEST_CPU   CPU architecture to build for. Valid values are: arm,
                        ia32, x64
````

jasnell:
@shigeki .. good point :-) @obastemur ... I'm going to close this here specifically because arm64 is not supported but I would encourage you to please make sure this is working in io.js. If it's not, then we'll want to get the fix in there. Would you be able to open this against io.js if necessary?

obastemur:
> I would encourage you to please make sure this is working in io.js. If it's not, then we'll want to get the fix in there. Would you be able to open this against io.js if necessary?

@jasnell sure.

PR #25653, opened 2015-07-09
Reconcile v0.12 with new CI 

orangemocha:
..continued from https://github.com/joyent/node/pull/25541

As part of CI reconciliation, these commits make v0.12 work with the jobs at https://jenkins-iojs.nodesource.com/ 

This PR addresses all review feedback from https://github.com/joyent/node/pull/25541. In particular:
- I left out the commit that changed the default for snapshots, based on feedback from @misterdjules . Instead, I introduced a new run-ci makefile rule which invokes configure internally with --without-snapshot. On Windows, the same is accomplished by calling vcbuild test-ci
- I made it so that Jenkins jobs can control whether to care about flaky tests or not, with an environment variable on *nix and a vcbuild argument on Windows.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,6 +244,11 @@ docopen: out/doc/api/all.html
 docclean:
 	-rm -rf out/doc
 
+run-ci:
+	$(PYTHON) ./configure --without-snapshot $(CONFIG_FLAGS)
+	$(MAKE)
+	$(MAKE) test-ci
+
 RAWVER=$(shell $(PYTHON) tools/getnodeversion.py)
 VERSION=v$(RAWVER)
 NODE_DOC_VERSION=$(VERSION)
@@ -443,4 +448,9 @@ cpplint:
 
 lint: jslint cpplint
 
-.PHONY: lint cpplint jslint bench clean docopen docclean doc dist distclean check uninstall install install-includes install-bin all staticlib dynamiclib test test-all test-addons build-addons website-upload pkg blog blogclean tar binary release-only bench-http-simple bench-idle bench-all bench bench-misc bench-array bench-buffer bench-net bench-http bench-fs bench-tls
+.PHONY: lint cpplint jslint bench clean docopen docclean doc dist distclean \
+	check uninstall install install-includes install-bin all staticlib \
+	dynamiclib test test-all test-addons build-addons website-upload pkg \
+	blog blogclean tar binary release-only bench-http-simple bench-idle \
+	bench-all bench bench-misc bench-array bench-buffer bench-net \
+	bench-http bench-fs bench-tls run-ci
--- a/Makefile
+++ b/Makefile
@@ -127,6 +127,9 @@ test-all-http1: test-build
 test-all-valgrind: test-build
 	$(PYTHON) tools/test.py --mode=debug,release --valgrind
 
+test-ci:
+	$(PYTHON) tools/test.py -p tap --logfile test.tap --mode=release --arch=$(DESTCPU) simple message internet
+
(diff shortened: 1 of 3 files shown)


vcbuild.bat
```diff
@@ -81,7 +84,6 @@ goto next-arg
 
 :args-done
 if defined upload goto upload
-if defined jslint goto jslint
```
misterdjules: Does `vcbuild jslint` still work? It seems that with this change it would only work when tests are run too.

vcbuild.bat
```diff
@@ -202,7 +204,11 @@ if "%test%"=="" goto exit
 if "%config%"=="Debug" set test_args=--mode=debug
 if "%config%"=="Release" set test_args=--mode=release
 
+set test_args=%test_args% --arch=%target_arch% 
+
+
```
misterdjules: Nitpicking, one blank line is enough.

Makefile
```diff
@@ -240,6 +244,11 @@ docopen: out/doc/api/all.html
 docclean:
 	-rm -rf out/doc
 
+run-ci:
+	$(PYTHON) ./configure --without-snapshot $(CONFIG_FLAGS)
```
misterdjules: Just an observation and a potential follow-up to this PR: if the goal of `run-ci` is to test builds of node with the same configuration as official releases, we should maybe add other configuration options used for them (such as `--with-intl=small-icu`). We could probably factor that out in a variable and reuse that in other targets used to build official releases (like `binary` and `pkg`).

Same for the `test-ci` command line option supported in `vcbuild.bat` below.

misterdjules:
Thank you @orangemocha! Added a few comments.

Makefile
orangemocha: Thanks for pointing that out. It wasn't in Jenkins so that why I didn't put it here (I just copied what was in Jenkins).
I will try to add this to this PR.

vcbuild.bat
orangemocha: Good catch. This change enables building and testing, when jslint is specified. But it has the side effect of not getting to jslint unless you run tests. Will fix.

vcbuild.bat
orangemocha: Will fix.

Makefile
orangemocha: Actually, I don't fully understand the implications on the download argument and whether it's a good idea to download the tarball in Jenkins. I will leave it to a follow-up PR as originally suggested.

orangemocha:
Updated to fix extra line and jslint issue.

Sneak preview CI run: https://jenkins-iojs.nodesource.com/job/orangemocha-test-commit/4/

misterdjules:
LGTM. That's great! Do we plan to backport this to v0.10 too and test + build v0.10 releases from the converged CI?

orangemocha:
I might give it a shot, but given that we currently don't use node-accept-pull-request for v0.10, I will probably settle for just porting the current workflow, and then work on enabling node-accept-pull-request on v0.10 later.

What job do you currently use on v0.10, BTW?

Thanks for reviewing!

orangemocha:
Landed in https://github.com/joyent/node/commit/c9ef4b3ae2a3d0a670a10621cdc79173bf54a4f3.

misterdjules:
@orangemocha Sounds good! Currently we use the `node-review` job to test changes targeted to v0.10, and we land them with `git-apply-pr`. Ideally we would use `node-accept-pull-request` and mark currently v0.10 flaky tests as such, and create issues to fix them.

PR #25660, opened 2015-07-11
doc: fix typo

edwinlin1987:
"Copyright" misspelled


KenanY:
@edwinlin1987 That's the license of libtabe. I'm not sure if this fix should go to them instead?

edwinlin1987:
Good point, I'll try to let them know.

PR #25671, opened 2015-07-13
OpenSuse Tumbleweed gcc (v. 5.1.1) -dumpversion issues

Rynaro:
Hello,
I found some difficulties when attempting to configure and compile NodeJS packages in openSUSE Tumbleweed.
The error reported when running ./configure

```
Traceback (most recent call last):
  File "./configure", line 992, in <module>
    configure_node(output)
  File "./configure", line 530, in configure_node
    o['variables']['gcc_version'] = 10 * cc_version[0] + cc_version[1]
IndexError: tuple index out of range
```

In the configure file, specifically the compiler_version function. It returned a tuple to cc_version variable. However, in OpenSUSE, if we execute:

```
henrique@aegis~>: gcc -dumpversion
5
```

It returns a single integer value, making the variable that contains the tuple store (5,), resulting in IndexError exception.

I made this change in my configure file, allowing the compilation of NodeJS in my Tumbleweed environment.

ps: Excuse me for my English, it is not my native language. I'm learning it yet. :)

### DIFF
--- a/configure
+++ b/configure
@@ -527,8 +527,11 @@ def configure_node(o):
   o['variables']['clang'] = 1 if is_clang else 0
 
   if not is_clang and cc_version != 0:
-    o['variables']['gcc_version'] = 10 * cc_version[0] + cc_version[1]
-
+    try:
+        o['variables']['gcc_version'] = 10 * cc_version[0] + cc_version[1]
+    except IndexError:
+        o['variables']['gcc_version'] = 10 * cc_version[0]
+        
   # clang has always supported -fvisibility=hidden, right?
   if not is_clang and cc_version < (4,0,0):
     o['variables']['visibility'] = ''


fusion809:
I received the same error you reported here and with your modification to the configure file I was able to install node, thanks.

jasnell:
This looks fine to me, but let's see what @orangemocha or @misterdjules think

orangemocha:
Interesting. So there is no other way of telling the minor version (.1) on this platform?

orangemocha:
The change per se LGTM. I probably would have implemented it with an if statement but the try-except works.

Rynaro:
@orangemocha. By default gcc 5.1+ dont returns to me a minor version of their compiler. I dont know if they have plans to reinsert the minor version in forward releases.

configure
```diff
@@ -527,8 +527,11 @@ def configure_node(o):
   o['variables']['clang'] = 1 if is_clang else 0
 
   if not is_clang and cc_version != 0:
-    o['variables']['gcc_version'] = 10 * cc_version[0] + cc_version[1]
-
+    try:
+        o['variables']['gcc_version'] = 10 * cc_version[0] + cc_version[1]
+    except IndexError:
```
thefourtheye: If we are expecting at least one element then we can do `if len(cc_version) == 2: ... else: ...`

Rynaro:
@thefourtheye . Really, the `if ... else` statement, is a great solution for this issue. But, ` try ... except` is better performatic than `if ... else`. And, Python encourages the use of exceptions in most of cases.
The `if ... else` works too, but I used `try ... except` in this commit, and I have no problem to change this code `try .. except` to `if .. else`. 
:smile:

thefourtheye:
I was worried that if the number of elements is zero then we ll not know which access throws the error.

configure
orangemocha: Or:
```
o['variables']['gcc_version'] = 10 * cc_version[0]
if len(cc_version) >= 2:
  o['variables']['gcc_version'] += cc_version[1]
else:
  print 'Warning: no gcc minor version found, assuming 0'
```

thefourtheye: Ya something like this. I didn't have a computer at that moment, so couldn't express myself clearly. Sorry.

Rynaro: ``o['variables']['gcc_version'] = 10 * cc_version[0]``
This line always run, because ``cc_version[0]`` refers the major version of gcc.
Now I think, we only need to check if ``cc_version[1]`` exists. If exists, use your value, if not exists show a warning message. Like the code above.
I've used ``try .. except`` because Python encourages use of  ``try``. 
I've tested ``if  .. else`` way and my ``try ... except`` way, both ways have solved this issue.

thefourtheye: I agree that Python encourages [EAFP: Easier to ask for forgiveness than permission.](https://docs.python.org/2/glossary.html#term-eafp). But I wanted to make sure that the code is easily understandable. Anyway I am not going to hold this PR with this discussion. Please continue.

orangemocha: Agreed. Will land as is.

orangemocha:
Just noticed that this PR is against master. If it is meant to go in v0.12, then it should be against the v0.12 branch. Otherwise, it probably makes more sense to do it in io.js master. @jasnell @misterdjules is that right?

jasnell:
Yes. Commits here need to be targeted at v0.12 or v0.10. Commits against master need to go into nodejs/io.js moving forward.

orangemocha:
@jasnell : makes sense, thank you. Assuming that 'being able to compile' is one of those fixes that we want to include in v0.10, this should land in v0.10 first. I can do the cherry-picking.

orangemocha:
Landed in 8564a9f5f75ad9069c0a1fb6a90ccadc83e0067b.

PR #25672, opened 2015-07-13
Fix util extend issue

itkoren:
When the "origin" object argument is null or not defined, _extend
throws exception.
Fix _extend to allow undefined "origin" object argument and return
the "add" object argument.


jasnell:
Unfortunately, I believe this ends up qualifying as an API change (even if only a small one). It would likely be better to target this at http://github.com/nodejs/io.js or http://github.com/nodejs/node. Also, the commits should be squashed. /cc @joyent/node-tsc

itkoren:
Thanks, I've created a new PR with this fix at http://github.com/nodejs/node

PR #25674, opened 2015-07-13
crypto: Show exponent in decimal and hex

### LINKED ISSUE #17202: TLS Peer certificate exponent listed as "10001"

iamthechad:
Show the exponent in a manner that matches OpenSSL
Fixes #17202


jasnell:
Definitely can appreciate the intent on this, but it does involve a bit of an API change, which we should likely avoid in v0.12 and v0.10. This should likely be targeted at http://github.com/nodejs/io.js or http://github.com/nodejs/node instead.

iamthechad:
I'll admit I don't know much about the timelines around which repo to use when. Are future changes not going to be applied to this repo any more? We knew this was a breaking change, which is why it's targeted at `master` and not a specific release.

jasnell:
For breaking changes, we want to target `nodejs/io.js` or `nodejs/node` as that is where the next major release will come from. `joyent/node` is largely intended for LTS of the `v0.12` and `v.10` streams.

iamthechad:
Okay. Is there a preference as to which repo - `nodejs/iojs` vs `nodejs/node`?

jasnell:
Likely io.js would be best.

iamthechad:
Resubmitted to `io.js`. Should this PR be closed?

jasnell:
Yes, thank you!

PR #25679, opened 2015-07-14
Only apply drain listener when callback provided

CraigCav:
When using the [compression](https://github.com/expressjs/compression) middleware in express with Server-Sent Events, I've noticed a `possible EventEmitter memory leak detected. 11 drain listeners added.` message appearing from Gzip.addListener:

``` term
at Gzip.addListener (events.js:179:15)
at Gzip.Readable.on (_stream_readable.js:671:33)
at Gzip.once (events.js:204:8)
at Gzip.Zlib.flush (zlib.js:451:10)
at ServerResponse.flush (/app/node_modules/compression/index.js:63:16)
```

The error can be reproduced using the SSE example on the expressjs/compression [readme](https://github.com/expressjs/compression#server-sent-events).

When `stream.flush()` is called without a callback, an empty listener is being added. Since flush may be called multiple times to push SSE's down to the client, multiple noop listeners are being added. This in turn causes the `memory leak detected` message.

Judging by [line 449](https://github.com/joyent/node/blob/6036e4f5a833b27fb474f3bf891c405606fdce19/lib/zlib.js#L449), an attempt has been made to avoid this situation, however, not all code paths have had the empty listener check applied.

### DIFF
--- a/lib/zlib.js
+++ b/lib/zlib.js
@@ -450,9 +450,11 @@ Zlib.prototype.flush = function(kind, callback) {
       this.once('end', callback);
   } else if (ws.needDrain) {
     var self = this;
-    this.once('drain', function() {
-      self.flush(callback);
-    });
+    if (callback) {
+      this.once('drain', function() {
+        self.flush(callback);
+      });
+    }
   } else {
     this._flushFlag = kind;
     this.write(new Buffer(0), '', callback);


cjihrig:
LGTM, although it would be better if the code actually checked that `callback` was a function instead of any truthy value.

jasnell:
hmm... according to the docs, `callback` is not optional, so I'm wondering if this is actually doing us a bit of a disservice. If the intent was to allow `callback` to be optional, then the docs need to be updated as well to reflect that. If the intent was for `callback` to be required, I'd rather see an explicit throw if `callback` is not a function.

CraigCav:
> If the intent was for callback to be required, I'd rather see an explicit throw if `callback` is not a function.

Agreed, although I'd be concerned that this was a breaking change for anyone currently calling `stream.flush()`. 

In my specific case, requiring a callback would also mean that the `possible EventEmitter memory leak detected. 11 drain listeners added.` message would be unavoidable. That said, given that the frequent calling of flush is undesirable anyway (since it limits the effectiveness of the compression algorithm) my case maybe better resolved by disabling the compression middleware for SSE routes.

I can't speak for the original intent of the `callback` parameter. Given `callback` isn't called until the `drain` event is emitted (which doesn't seem to necessarily coincide with the completion of the flush), I'm not even sure I can provide a use case for having the `callback` parameter that wouldn't be better suited to having the called add the listener themselves.

I'm happy to update the docs to reflect `callback` being optional if that is appropriate?

jasnell:
@joyent/node-tsc ... what do you think?

lib/zlib.js
```diff
@@ -450,9 +450,11 @@ Zlib.prototype.flush = function(kind, callback) {
       this.once('end', callback);
   } else if (ws.needDrain) {
     var self = this;
-    this.once('drain', function() {
-      self.flush(callback);
-    });
+    if (callback) {
```
cjihrig: If we're going to add this check, might as well make it a check for a function.

jasnell: Agreed

usernameisalreadytaken2014:
I'm also running into this problem (node v4.2.1).

Also using flush() with SSE.

Disagree with @CraigCav that compression is not useful.  My stream goes from 600 KiB down to 34 KiB with compression enabled on an SSE stream.

CraigCav:
@usernameisalreadytaken2014 I wasn't saying that compression is not useful. I was saying that in my specific circumstances, disabling the compression middleware for SSE routes might be an acceptable work-around. Disabling compression for all SSE routes is not going to be an appropriate solution for everyone.

I would much rather have the listener only applied if an optional callback is provided.

usernameisalreadytaken2014:
@CraigCav - check!  I misunderstood, apologies.

Did you find a workaround?  Does attaching a callback help?

(And if so, is it necessary to hold back further write()s until the callback has been activated, or?...)

CraigCav:
> Does attaching a callback help?

I don't think so. The problem as I see it is that, regardless of whether a callback is provided, a `drain` listener is attached. Given that with flush is being called multiple times to push SSE's down to the client, multiple event handlers are being added.

I haven't found a workaround (other than removing compression which doesn't sound helpful for your case). It might be possible to remove all `drain` listeners after calling flush, although I haven't explored the side-effects of doing so.

MylesBorins:
@CraigCav would it make sense to resubmit this PR against nodejs/node?

Fishrock123:
@TheAlphaNerd I was planning to, but my dev machine is broken for a couple days anyways. Feel free to take it if @CraigCav isn't around. :)

MylesBorins:
The fix is now on master and v4.x-staging

PR #25683, opened 2015-07-15
Add the CESU-8 encoding support

pmq20:
In our daily work it frequently occurs that we need to let Node.js apps communicate in binary with Java apps, which necessitates the CESU-8 encoding. Details about this encoding is observed in [1]. This pull request is a quick implementation of it. One test case has been added and passed.

Since it modifies the upstream v8 code, I am not sure about how to handle such kind of patches. It seems problematic to make changes naively onto the node repository. Please guide me on how to merge this code. Thank you very much!
- CESU-8 stands for Compatibility Encoding Scheme for UTF-16: 8-Bit
- Function VisitTwoByteString() was made virtual to call in CESU-8 mode
- Added a new class CesuVisitor inheriting from Visitor
- Added Smiling Face symbol as a test case
- New API - String::Cesu8Length()
- New String::WriteOptions - ENCODE_AS_CESU_8

[1] https://en.wikipedia.org/wiki/CESU-8

### DIFF
--- a/deps/v8/include/v8.h
+++ b/deps/v8/include/v8.h
@@ -1616,6 +1616,12 @@ class V8_EXPORT String : public Primitive {
   int Utf8Length() const;
 
   /**
+  * Returns the number of bytes in the CESU-8 encoded
+  * representation of this string.
+  */
+  int Cesu8Length() const;
+
+  /**
    * Returns whether this string is known to contain only one byte data.
    * Does not read the string.
    * False negatives are possible.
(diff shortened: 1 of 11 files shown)


pmq20:
See also, https://github.com/nodejs/io.js/pull/2190

jasnell:
While I certainly can appreciate it, if it lands it would likely need to be done over in http://github.com/nodejs/io.js or http://github.com/nodejs/node. Because of the changes involved, it's not likely that this would ever land in v0.12. Since there's the other conversation already going on the io.js side, I'm going to close this here.

PR #25707, opened 2015-07-17
configure: add --without-mdb flag

### LINKED ISSUE #6439: Compilation of v0.11.8 fails on Solaris 10 and some Illumos distros due to missing libproc.h

misterdjules:
Fixes build issues on Solaris based platforms where libproc.h is not
available or not compatible with the one shipped by SmartOS.

Fixes #6439.

This is just a PR that fixes the commit message from https://github.com/joyent/node/pull/9349. It's otherwise strictly identical.

### DIFF
--- a/configure
+++ b/configure
@@ -282,6 +282,11 @@ parser.add_option('--without-etw',
     dest='without_etw',
     help='build without ETW')
 
+parser.add_option('--without-mdb',
+    action='store_true',
+    dest='without_mdb',
+    help='build without mdb')
+
 parser.add_option('--without-npm',
     action='store_true',
     dest='without_npm',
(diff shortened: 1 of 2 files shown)


misterdjules:
This was already reviewed at https://github.com/joyent/node/pull/9349, landing in v0.12 now.

misterdjules:
Landed in 8b81f98c41158856c09b57afc4cf879e9b65efe9, thank you @cgalibern :+1:

Issue #25736, opened 2015-07-18
secureOptions SSL_OP_NO_TICKET causes 0.12 SSL server to hang


tlivings:
Using the SSL_OP_NO_TICKET will not allow server to complete handshake.

Example:

```javascript
var Https = require('https');
var Fs = require('fs');
var C = require('constants');

var server = Https.createServer({
    secureOptions: C.SSL_OP_NO_TICKET,
    cert: Fs.readFileSync('/usr/local/ssl/server.crt'),
    key: Fs.readFileSync('/usr/local/ssl/server.key')
}, function (req, res) {
    res.writeHead(200);
    res.end("hello world\n");
});

server.on('resumeSession', function (id, callback) {
    console.log('resume requested');
    callback();
});

server.listen(8000);
```

tlivings:
cc @indutny

indutny:
It is actually a manifestation of https://github.com/joyent/node/issues/25735. Works like a charm in io.js

PR #25739, opened 2015-07-18
tls: do not hang without `newSession` handler

indutny:
When listening for client hello parser events (like OCSP requests), do
not hang if `newSession` event handler is not present.

### DIFF
--- a/test/simple/test-tls-new-session-hang.js
+++ b/test/simple/test-tls-new-session-hang.js
@@ -1,0 +1,42 @@
+var common = require('../common');
+
+if (!process.features.tls_ocsp) {
+  console.error('Skipping because node compiled without OpenSSL or ' +
+                'with old OpenSSL version.');
+  process.exit(0);
+}
+
+var assert = require('assert');
+var tls = require('tls');
+var constants = require('constants');
+var fs = require('fs');
+var join = require('path').join;
+
+var keyFile = join(common.fixturesDir, 'keys', 'agent1-key.pem');
+var certFile = join(common.fixturesDir, 'keys', 'agent1-cert.pem');
+var caFile = join(common.fixturesDir, 'keys', 'ca1-cert.pem');
+var key = fs.readFileSync(keyFile);
+var cert = fs.readFileSync(certFile);
+
+var server = tls.createServer({
+  cert: cert,
+  key: key
+}, function (socket) {
+  socket.destroySoon();
+});
+
+server.on('resumeSession', common.mustCall(function() {
+  // Should not be actually called
+}, 0));
+
+server.listen(common.PORT, function() {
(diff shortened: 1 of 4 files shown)


indutny:
cc @joyent/node-collaborators

test/simple/test-tls-new-session-hang.js
```diff
@@ -0,0 +1,43 @@
...
+
+var server = tls.createServer({
+  cert: cert,
+  key: key
+}, function (socket) {
+  socket.destroySoon();
+});
+
+// Should not be actually called
+server.on('resumeSession', function (id, callback) {
```
thefourtheye: Can we use `common.mustCall(function() {}, 0)`?

test/simple/test-tls-ocsp-callback.js
```diff
@@ -98,12 +105,19 @@ function test(testOptions, cb) {
   });
 
   process.on('exit', function() {
+    if (testOptions.ocsp === false) {
+      assert.equal(requestCount, clientSecure);
+      assert.equal(requestCount, 1);
```
thefourtheye: I would use `common.mustCall` to do the number of invocation checks.

test/simple/test-tls-new-session-hang.js
indutny: Fixed.

tlivings:
Can we get some action on this - it's a pressing concern for us.

sam-github:
LGTM

@indutny  Does this effect io.js, too?

@tlivings Wouldn't a workaround be to just establish a newSession listener? You don't have to save the session, just callback immediately.

indutny:
@sam-github it is effectively a backport from io.js, except the new test case. @sam-github should I just land the PR by hand as we do in io.js?

tlivings:
@sam-github simply adding a listener doesn't work since we use `SSL_OP_NO_TICKET`. See also issue: https://github.com/joyent/node/issues/25736

indutny:
CI: http://jenkins.nodejs.org/job/node-accept-pull-request/159/ (thanks @misterdjules )

sam-github:
@indutny you've committed **way** more to joyent/node than I have!

I don't think the procedure is changed much from when you were working all the time over here, but I don't recall from the onboarding, and can't find anything specific to how to land in https://nodejs.org/contribute/code_contributions/workflow.html

@misterdjules ?

indutny:
Cancelled that CI job, new one is:http://jenkins.nodejs.org/job/node-accept-pull-request/160/

indutny:
Thanks @sam-github figured it out with @misterbisson

indutny:
Looks like it landed in v0.12.

PR #25748, opened 2015-07-22
test: fix test-debug-port-from-cmdline

joaocgreis:
`test-debug-port-from-cmdline` became an issue in io.js in https://github.com/nodejs/io.js/issues/2094 , when it was noticed that it failed frequently on Windows 2008. The error messages displayed were not clear and threw investigation off from the real problem.

The problem turned out to be with the test itself, and it failed also on Linux but silently, which is also the case for v0.12. For example here: http://jenkins.nodejs.org/job/node-test-commit-unix/132/DESTCPU=ia32,label=linux/tapTestReport/simple.tap-126/ the output should have been 2 lines:

```
> Starting debugger agent.
> Debugger listening on port 12346
```

This fix changes the test so that it runs as expected and always asserts the expected result. Further details in the original PR: https://github.com/nodejs/io.js/pull/2186

### DIFF
--- a/test/simple/test-debug-port-from-cmdline.js
+++ b/test/simple/test-debug-port-from-cmdline.js
@@ -24,11 +24,11 @@ var assert = require('assert');
 var spawn = require('child_process').spawn;
 
 var debugPort = common.PORT;
-var args = ['--debug-port=' + debugPort];
+var args = ['--interactive', '--debug-port=' + debugPort];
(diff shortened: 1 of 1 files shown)


orangemocha:
:+1:

cjihrig:
LGTM

joaocgreis:
Thanks! Landed in ceb6a8c131e53036f041bd2c5ff834ef1398981e.

PR #25752, opened 2015-07-22
Make http.request correctly parse {host: "hostname:port"}

### LINKED ISSUE #25751: http.request fails when host contains port
`host` is different from `hostname` in that it may contain a port number. For example:

```js
require('url').parse('http://localhost:8080')
{ protocol: 'http:',
  slashes: true,
  auth: null,
  host: 'localhost:8080',
  port: '8080',
  hostname: 'localhost',
  hash: null,
  search: null,
  query: null,
  pathname: '/',
  path: '/',
  href: 'http://localhost:8080/' }
```
However, `http.request({options})` treats both fields as `hostname` and attempts an invalid DNS lookup when given a `host` with a port. For example:

```js
http.get({host: 'localhost:8080', path: '/'}, console.log)
```
```bash
> Error: getaddrinfo ENOTFOUND localhost:8080
    at errnoException (dns.js:44:10)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:94:26)
```

nfriedly:
Fixes #25751 so that `http.request()` doesn't attempt an invalid dns lookup of `hostname:port` when it should only be looking up `hostname`.

### DIFF
--- a/lib/_http_client.js
+++ b/lib/_http_client.js
@@ -80,10 +80,10 @@ function ClientRequest(options, cb) {
 
   var defaultPort = options.defaultPort || self.agent && self.agent.defaultPort;
 
-  var hostParts = options.host && options.host.split(':');
+  var hostParts = (options.host && options.host.split(':')) || [];
 
-  var port = options.port = options.port || (hostParts && hostParts[1]) || defaultPort || 80;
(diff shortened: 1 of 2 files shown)


jasnell:
Going to close this here. If it lands, io.js or the converged branch will be the right place. This shouldn't land in v0.12 or v0.10

PR #25753, opened 2015-07-22
net: ensure Socket reported address is current

### LINKED ISSUE #25621: net: Wrong localPort on using the same socket object for multiple connect/destroy
When the same socket object is used for multiple .connect(), the socket.localPort is not refreshing with new localPort used by the OS.

See the discussion on mailing list:
https://groups.google.com/d/topic/nodejs/eT3xanyP83U/discussion

rmg:
Any time the connection state or the underlying handle itself changes,
the socket's name (aka, local address) can change.

To deal with this we need to reset the cached sockname any time we
set or unset the internal handle or an existing handle establishes a
connection.

Fixes #25621
Same PR on iojs: nodejs/io.js#2095

### DIFF
--- a/lib/net.js
+++ b/lib/net.js
@@ -124,6 +124,7 @@ function initSocketHandle(self) {
   self.destroyed = false;
   self.bytesRead = 0;
   self._bytesDispatched = 0;
+  self._sockname = null;
 
   // Handle creation may be deferred to bind() or connect() time.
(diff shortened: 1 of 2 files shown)


jasnell:
@rmg .. this should be closed here. I believe the PR landed over on the nodejs/node side. Can reopen this if necessary

PR #25755, opened 2015-07-23
test: make test-abort-fatal-error non flaky

### LINKED ISSUE #25720: Investigate v0.10 flaky test test-abort-fatal-error
https://github.com/joyent/node/commit/a2f879f197f76d91a42ec0f8322f18948911a552 marked this test as flaky in v0.10 so that it doesn't block merges with Jenkins and node-accept-pull-request.

As a follow up, we should investigate why this test is failing in v0.10. Some of the tests marked as flaky in https://github.com/joyent/node/commit/a2f879f197f76d91a42ec0f8322f18948911a552 fail intermittently, while others fail persistently. For the latter, an explicit skip for the specific platform might be more appropriate.

My guess is that a good percentage of these tests fail because of issues with the test itself, not node. It's likely that those issues were fixed in v0.12, so it might be a good idea to look for changes to the same test in git history in v0.12, or even simply trying the v0.12 version of the test with node v0.10.

/cc: @joyent/node-collaborators

misterdjules:
Before this change, test/simple/test-abort-fatal-error.js would fail in
some environments for reasons I wasn't able to fully understand. It was
marked as flaky on some systems, but not on others on which it was
failing sometimes (OSX).

This change basically syncs test-abort-fatal-error with how it's
implemented in v0.12. It back ports 429b587 (or rather the parts that
apply to it since it's a merge commit), 2f5e77f and 114bff4.

After backporting these changes in v0.10, test-abort-fatal-error is not
flaky anymore in environments for which it was flaky. It also has the
added benefit of being more robust because it checks exit codes and
signals instead of error messages.

Tested on OSX and SmartOS, the only platforms on which I could reproduce
the flakiness of this test.

This change also removes test-abort-fatal-error from the list of flaky
tests in test/simple/simple.status.

Fixes #25720.

### DIFF
--- a/test/simple/simple.status
+++ b/test/simple/simple.status
@@ -17,7 +17,6 @@ test-tls-securepair-server              : PASS,FLAKY
 test-timers-first-fire                  : PASS,FLAKY
  
 [$system==linux]
-test-abort-fatal-error                  : PASS,FLAKY
 test-debugger-client                    : PASS,FLAKY
 test-process-argv-0                     : PASS,FLAKY
 test-cluster-master-kill                : PASS,FLAKY
@@ -29,7 +28,6 @@ test-cluster-master-error               : PASS,FLAKY
 test-http-exit-delay                    : PASS,FLAKY
 
 [$system==solaris]
-test-abort-fatal-error                  : PASS,FLAKY
 test-debugger-repl                      : PASS,FLAKY
 test-debugger-repl-break-in-module      : PASS,FLAKY
 test-debugger-repl-utf8                 : PASS,FLAKY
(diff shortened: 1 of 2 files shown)


misterdjules:
This change is needed to land https://github.com/joyent/node/pull/25569.

joaocgreis:
@misterdjules Commit message: you mention `429b5870q` (extra `q`), but also include changes made in 2f5e77f55bb7af8ba0b9643a83abdb81b4255d9b and 114bff467eb67def0dbe3baddf06969bcadd0ecc. Maybe change that paragraph to something like (edit as you see fit):

```
This change back ports this test as it was in v0.12 at 114bff46. The
test was changed to adapt to changes in V8 in the v0.12 branch.
```

joaocgreis:
Tested on SmartOS, LGTM.

misterdjules:
Good catch and thank you @joaocgreis :+1: Updated the commit message, will land asap.

misterdjules:
Landed in e10892cccc6e64be413bf41c4c33af4557fd7a69.

Issue #9325, opened 2015-03-04
test-net-error-twice.js seems to not behave as expected


misterdjules:
It seems that [`test-net-error-twice.js`](https://github.com/joyent/node/blob/v0.10/test/simple/test-net-error-twice.js) does not behave as expected. Its goal is to test [`fireErrorCallbacks`](https://github.com/joyent/node/blob/v0.10/lib/net.js#L437-L445), but it doesn't do it correctly, leading to false negatives on some platforms and failures on others .

For a `net.Socket` object, each write operation are [setup to call `afterWrite`](https://github.com/joyent/node/blob/v0.10/lib/net.js#L648), which itself [calls `net.Socket.prototype_destroy`](https://github.com/joyent/node/blob/v0.10/lib/net.js#L719-L723) if the write operation fails.

`net.Socket.prototype_destroy` in turn calls `fireErrorCallbacks`, whether the `net.Socket` object [has been already destroyed](https://github.com/joyent/node/blob/v0.10/lib/net.js#L447-L451) or [not](https://github.com/joyent/node/blob/v0.10/lib/net.js#L475-L476).

This test seems to have been written so that it triggers [more than one write](https://github.com/joyent/node/blob/v0.10/test/simple/test-net-error-twice.js#L33) by [having a buffer that is large enough](https://github.com/joyent/node/blob/v0.10/test/simple/test-net-error-twice.js#L26).

It expects to trigger several errors with these writes [by closing the connection on the client side](https://github.com/joyent/node/blob/v0.10/test/simple/test-net-error-twice.js#L45-L47).
and makes sure that [`fireErrorCallbacks` will only emit the same write error once](https://github.com/joyent/node/blob/v0.10/test/simple/test-net-error-twice.js#L34-L38).

However what seems to happen is that at most one write operation fails (which leads to the test succeeding although `fireErrorCallbacks` has been called only once, and so hasn't been really tested), and sometimes they all succeed (like [on SmartOS](http://jenkins.nodejs.org/job/node-review-unix-julien/DESTCPU=x64,label=smartos/lastBuild/tapTestReport/simple.tap-364/)) which leads to the test failing, but not necessarily for the good reasons.

misterdjules:
[...]

PR #25760, opened 2015-07-23
test: mark test-net-error-twice flaky on SmartOS

misterdjules:
It seems that test-net-error-twice.js does not behave as expected. Its
goal is to test fireErrorCallbacks, but it doesn't do it correctly,
leading to false negatives on some platforms and failures on others.
This change marks this test as flaky so that we can use our CI to land
changes in the v0.10 branch until we can fix it properly.

See the corresponding issue at
https://github.com/joyent/node/issues/9325 for more details.


joaocgreis:
LGTM

misterdjules:
Landed in 7df57703dd2449382263419a266efa43f403734c, thank you @joaocgreis!

PR #25763, opened 2015-07-24
timers: fix processing of nested same delay timers

### LINKED ISSUE #25607: test: microtask-queue-run(-domain) should be made non-flaky
`test-microtask-queue-run` and `test-microtask-queue-run-domain` appear to be flaky (PR: https://github.com/joyent/node/pull/25606), failing very rarely. Testing 10k runs on the platforms I have available, I got failure rates as high as:

- Linux x64: `OK: 9964   NOT OK: 36`

- Windows 7 x64: `OK: 9843   NOT OK: 157`

- SmartOS x86 (gcc 4.7): `OK: 9904   NOT OK: 96`

The results appear to be influenced by the machine load, I got much lower results on an otherwise idle machine.

Quick test:
```
time (OK=0; NOK=0; for i in `seq 10000`; do ./out/Release/node test/simple/test-microtask-queue-run.js && OK=$(($OK+1)) || NOK=$(($NOK+1)); echo "$i   OK: $OK   NOT OK: $NOK"; done)
```

Other tests from `test-microtask-queue-*` seem okay.

whitlockjc:
Whenever a timer with a specific timeout value creates a new timer with
the same timeout, the newly added timer might be processed immediately
in the same tick of the event loop instead of during the next tick of
the event loop at the earliest.

Fixes #25607

### DIFF
--- a/lib/timers.js
+++ b/lib/timers.js
@@ -87,13 +87,35 @@ function listOnTimeout() {
 
   var first;
   while (first = L.peek(list)) {
-    // If the previous iteration caused a timer to be added,
-    // update the value of "now" so that timing computations are
-    // done correctly. See test/simple/test-timers-blocking-callback.js
-    // for more information.
-    if (now < first._monotonicStartTime) {
-      now = Timer.now();
-      debug('now: %d', now);
+    // This handles the case of a timer that was created within a timers
+    // callback with the same timeout value. For instance, when processing the
+    // timer that would call `bar` in such code:
+    //
+    //   setTimeout(function foo() { setTimeout(function bar() {}, 0) }, 0);
+    //
+    // or
+    //
+    //   setTimeout(function foo() { setTimeout(function bar() {}, 500) }, 500);
+    //
+    // We want to make sure that newly added timer fires in the next turn of the
+    // event loop at the earliest. So even if it's already expired now,
+    // reschedule it to fire later.
+    //
+    // At that point, it's not necessary to process any other timer in that
+    // list, because any remaining timer has been added within a callback of a
+    // timer that has already been processed, and thus needs to be processed at
+    // the earliest not in the current tick, but when the rescheduled timer will
+    // expire.
+    //
+    // See: https://github.com/joyent/node/issues/25607
+    if (now <= first._monotonicStartTime) {
+      var timeRemaining = msecs - (Timer.now() - first._monotonicStartTime);
+      if (timeRemaining < 0) {
+        timeRemaining = 0;
+      }
+      debug(msecs + ' list wait because timer was added from another timer');
+      list.start(timeRemaining, 0);
+      return;
     }
 
     var diff = now - first._monotonicStartTime;
--- a/test/simple/test-timers-nested.js
+++ b/test/simple/test-timers-nested.js
@@ -1,0 +1,58 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+var common = require('../common');
+var assert = require('assert');
+
+// Make sure we test 0ms timers, since they would had always wanted to run on
+// the current tick, and greater than 0ms timers, for scenarios where the
+// outer timer takes longer to complete than the delay of the nested timer.
+// Since the process of recreating this is identical regardless of the timer
+// delay, these scenarios are in one test.
+var scenarios = [0, 100];
+
+scenarios.forEach(function (delay) {
+  var nestedCalled = false;
+
+  setTimeout(function A() {
+    // Create the nested timer with the same delay as the outer timer so that it
+    // gets added to the current list of timers being processed by
+    // listOnTimeout.
+    setTimeout(function B() {
+      nestedCalled = true;
+    }, delay);
+
+    // Busy loop for the same timeout used for the nested timer to ensure that
+    // we are in fact expiring the nested timer.
+    common.busyLoop(delay);
+
+    // The purpose of running this assert in nextTick is to make sure it runs
+    // after A but before the next iteration of the libuv event loop.
+    process.nextTick(function() {
+      assert.ok(!nestedCalled);
+    });
+
+    // Ensure that the nested callback is indeed called prior to process exit.
+    process.on('exit', function onExit() {
+      assert.ok(nestedCalled);
+    });
+  }, delay);
+});
--- a/test/common.js
+++ b/test/common.js
@@ -22,6 +22,7 @@
 var path = require('path');
 var fs = require('fs');
 var assert = require('assert');
+var Timer = process.binding('timer_wrap').Timer;
 
 exports.testDir = path.dirname(__filename);
 exports.fixturesDir = path.join(exports.testDir, 'fixtures');
@@ -220,9 +221,9 @@ exports.hasMultiLocalhost = function hasMultiLocalhost() {
 };
 
 exports.busyLoop = function busyLoop(time) {
-  var startTime = new Date().getTime();
+  var startTime = Timer.now();
   var stopTime =  startTime + time;
-  while (new Date().getTime() < stopTime) {
+  while (Timer.now() < stopTime) {
     ;
   }
 };


lib/timers.js
```diff
@@ -87,13 +87,18 @@ function listOnTimeout() {
...
   var first;
   while (first = L.peek(list)) {
-    // If the previous iteration caused a timer to be added,
-    // update the value of "now" so that timing computations are
-    // done correctly. See test/simple/test-timers-blocking-callback.js
-    // for more information.
+    // Whenever a timer creates a new timer in the current list being processed
+    // you can end up processing the new timer immediately instead of waiting
+    // the scheduled time.  Whenever we encounter this situation stop
+    // processing this list and reschedule.  (Issue 25607)
```
Fishrock123: Could you put the full issue URL here? Thanks. :)

whitlockjc: Of course, consider it done.

lib/timers.js
```diff
@@ -87,13 +87,20 @@ function listOnTimeout() {
 
   var first;
   while (first = L.peek(list)) {
-    // If the previous iteration caused a timer to be added,
```
misterdjules, line 90: We want to keep these comments because they are still relevant: we still need to update `now` or use `Timer.now()` to make sure timing computations are done correctly. However we may want to move these comments to were we use `Timer.now()` right below.

whitlockjc, line 90: I am already using `Timer.now()` to do the computation and since we didn't need to update `now`, it's not used anywhere else and we quick return, I didn't think the previous comments made sense sticking around since we were no longer updating `now`.  Am I misunderstanding?

lib/timers.js
```diff
@@ -87,13 +87,20 @@ function listOnTimeout() {
...
-    // update the value of "now" so that timing computations are
-    // done correctly. See test/simple/test-timers-blocking-callback.js
-    // for more information.
+    // Whenever a timer creates a new timer in the current list being processed
```
misterdjules: `you can end up processing the new timer immediately instead of waiting the scheduled time.`
is a bit too vague. I would suggest rephrasing it as following:

```
This handles the case of a timer that was created within a timers callback with the same timeout value.
For instance, when processing the timer that would call `bar` in such code :

setTimeout(function foo() { setTimeout(function bar() {}, 0) }, 0);

or 

setTimeout(function foo() { setTimeout(function bar() {}, 5000) }, 5000);

We want to make sure that newly added timer fires in the next turn of the event loop
at the earliest. So even if it's already expired now, reschedule it to fire later.

At that point, it's not necessary to process any other timer in that list,
because any remaining timer has been added within a callback of a timer
that has already been processed, and thus needs to be processed at the
earliest not in the current tick, but when the rescheduled timer will expire.
```

Basically we want to be as explicit and precise as possible, so that when we come back to this piece of code
in 6 months, we understand why we made this change.

lib/timers.js
```diff
@@ -87,13 +87,20 @@ function listOnTimeout() {
...
+    // you can end up processing the new timer immediately instead of waiting
+    // the scheduled time.  Whenever we encounter this situation stop
+    // processing this list and reschedule.
+    //
+    // See: https://github.com/joyent/node/issues/25607
     if (now < first._monotonicStartTime) {
-      now = Timer.now();
-      debug('now: %d', now);
+      var delay = msecs - (Timer.now() - first._monotonicStartTime);
```
misterdjules: I would suggest renaming `delay` to `timeRemaining`, as it seems closer to what this value represents, but I'm not a native English speaker, so I may be wrong.

lib/timers.js
whitlockjc: You got it.  Should I update the commit message as well?  The code comment would make an excellent description in the Git commit message if you ask me.

lib/timers.js
whitlockjc: Works for me.

lib/timers.js
misterdjules: I think your commit message is fine, with the exception that it says that we process timers added within timers' callbacks immediately in _all cases_ when the two timers' timeouts are equal. Actually that happens _only_ if the time to run through all previously scheduled timers with the same timeout value exceeds the timers' timeout.

This is likely to happen with timers with short timeouts, like with `setTimeout(function() { setTimeout(function() {}, 0); }, 0);`. However it would not happen with longer ones, unless there's blocking code that takes a while to run in the first timer's callback.

So I would rephrase the beginning of your commit message as such:

```
Whenever a timer with a specific timeout value creates a new timer with the same
timeout, the newly added timer might be processed immediately in the same tick
of the event loop instead of during the next tick of the event loop at the earliest.
```

lib/timers.js
misterdjules, line 90: Agreed, it's probably self-explanatory in the current state of this PR. Thank you for questioning my comment :) :+1:

misterdjules:
Added the P-1 label because this PR fixes a few flaky tests (`test-microtask-queue-run` and `test-microtask-queue-run-domain.js`) in v0.10 and v0.12.

misterdjules:
@whitlockjc I would actually suggest removing `test-microtask-queue-run` and `test-microtask-queue-run-domain` from `test/simple/simple.status` in the same commit.

misterdjules:
@whitlockjc Thank you very much for submitting this PR. I left a few comments after a first review. This definitely looks like it's going in the right direction. Please let me know when my comments are addressed, and I'll do another review.

whitlockjc:
@misterdjules I originally wrote this code on `master` which does have those tests and the test suite ran without error.  As for the flaky test reported in #25607, I ran the same 10K recipe in the report and it ran all 10K tests without a single error.  Of course, when I ported this work over to `v0.10` those tests aren't there so I just relied on the `make test` passing.

misterdjules:
@whitlockjc Ok, that sounds good!

whitlockjc:
@misterdjules I do believe I've addressed your above requests:

* Changed variable in `lib/timers.js` from `delay` to `timeRemaining`
* More verbose documentation in `lib/timers.js`
* Updated commit message details

misterdjules:
In the absence of `test-microtask-queue-run.js` and `test-microtask-queue-run-domain.js` in v0.10, we'd need at least one test for this change. Even with these two tests in v0.12, I'm not familiar with the semantics of the microtasks queue, and `process.nextTick`'s behavior compared to `setTimeout` callbacks is [clearly defined](https://nodejs.org/api/process.html#process_process_nexttick_callback), so having specific tests for this change in v0.12 would also be valuable.

I would suggest starting with making sure that with the following code:
```
setTimeout(function foo() { setTimeout(function bar() {}, 0); process.nextTick(function baz() {}); }, 0);
```
`baz` is called before `bar`.

I would add another variant with nested `setTimeout` calls with the same timeout but > 0. Then you can go crazy and add even more nesting, anything you can think of basically :)

misterdjules:
@whitlockjc Another test you can add too is something like the following:
```
var TIMEOUT = 100;

var barCalled = false;

setTimeout(function foo() {
    setTimeout(function bar() {
        barCalled = true;
    }, TIMEOUT);

    // busy loop for timeout miliseconds so that
    // when listOnTimeout is done calling foo,
    // it processes the inner timer and find that its
    // timer has expired
    common.busyLoop(timeout);

    process.nextTick(function baz() {
        // make sure this nextTick callback
        // is called before the inner setTimeout
        // callback
        assert(barCalled === false);
    });
}, TIMEOUT);
```

I believe that would fail without the changes in this PR.

misterdjules:
@tunniclm Continuing the discussion regarding [your latest comment](https://github.com/joyent/node/issues/25607#issuecomment-125133997) here, because this way we can look at the changes that are being discussed.

> What I found in my exploration of nested setTimeout() calls is that it is possible to add a timeout in a timeout callback such that it fires on the same iteration of the event loop in which it was added. 

Yes, that's the same conclusion @whitlockjc and I arrived at in this PR.

> I didn't see a note in the documentation to say whether this should be allowed or not (whereas there is such a note for setImmediate()), but perhaps I missed it?

I don't think you missed anything, unless we both missed it (which is possible :) ).

> The above PR assumes that this scenario should not be allowed and if that is correct, then it looks like a good fix for this issue. And in that case I think we should update the doc to describe this contract under https://nodejs.org/api/timers.html#timers_settimeout_callback_delay_arg .

My opinion is there are at least two things to consider:

1. The previous behavior of Node.js.
2. Conformance to browsers' behavior/specs.

For 1), the previous behavior of Node.js was definitely to schedule nested timers' callbacks in different ticks of the event loop, which is why `test-microtask-queue-run.js` and `test-microtask-queue-run-domain.js` started to fail after d38e865fce93c00cb128034202fb1e26987efdbc.

As of 2), it seems that all browsers I have access to (Chrome, Firefox, Opera), with the following code:
```
setTimeout(function () {
    setTimeout(function() {
        console.log('timer callback called');
    }, 0);
    
    Promise.resolve().then(function() {
        console.log('promise callback called');
    });
}, 0);
```
consistently produce the following output:
```
promise callback called
timer callback called
```

The HTML spec also indicates that timers are [queued in their own tasks queue](https://html.spec.whatwg.org/multipage/webappapis.html#timer-initialisation-steps) (see item number 14).

In other words, it seems that:

1. Regardless of the timeout value, timers should never fire in the same tick of the event loop, and we should document it as you suggested.
2. This would not be a breaking change, so this can land in v0.10.

whitlockjc:
As requested, I just updated the PR to include a unit test.  I wrote this test against `v0.10` without my changes to ensure the test would fail consistently.  With the logic changes in `lib/timers.js`, this test passes consistently.

**Note:** This commit also includes a changes to `test/common.js#busyLoop` so that it uses the monotonic clock just like timers do.

jasnell:
@misterdjules ... did you have any further thoughts on this one?
@chrisdickinson ... do you happen to have any thoughts on it?

lib/timers.js
```diff
@@ -87,13 +87,35 @@ function listOnTimeout() {
...
+    //
+    // See: https://github.com/joyent/node/issues/25607
+    if (now <= first._monotonicStartTime) {
+      var timeRemaining = msecs - (Timer.now() - first._monotonicStartTime);
+      if (timeRemaining < 0) {
+        timeRemaining = 0;
+      }
+      debug(msecs + ' list wait because timer was added from another timer');
+      list.start(timeRemaining, 0);
+      return;
```
Fishrock123, line 118: Would it be better to just track if we are in a timeout and do the fix in like `timer.active()`?

In general, touching the actual timeout loop seems pretty hairy.

whitlockjc, line 118: Let me wrap my head around this.  I'm not sure what extra state, etc. would be required for your approach.  Do you think the edge case, `now <= timer._monotonicStartTime` would still be necessary to catch in `listOnTimeout`.  If so, it might not be worth doing thing differently as we'd still have the edge case to deal with.

whitlockjc:
I don't want this to drop through the cracks so can we update this with the next steps?  Really hoping to either find out that this isn't necessary in the new node repository or be told what the next steps are.  If it's still useful/important, I'll gladly port this to the new repository via a PR.

/cc @nodejs/collaborators and @nodejs/tsc

Fishrock123:
I'm mostly away right now (vacation) so ping me every few weeks until I am able to take a look at it, if no-one else does first.

As a good fist step, could you move this to the nodejs/node repo? thanks!

whitlockjc:
Done.  I fixed the merge conflicts but the PR is primarily the same.  I did move `test/simple/test-timers-nested.js` to `test/timers/test-timers-nested.js` to be consistent with other timers tests.  The other changes were trivial.

lib/timers.js
misterdjules, line 118: @Fishrock123 What did you have in mind more specifically? Some actual code would help.

test/simple/test-timers-nested.js
```diff
@@ -0,0 +1,50 @@
...
+    setTimeout(function B() {
+      nestedCalled = true;
+    }, delay);
+
+    // Busy loop for the same timeout used for the nested timer to ensure that we
+    // are in fact expiring the nested timer.
+    common.busyLoop(delay);
+    
+    process.nextTick(function() {
+      assert.ok(!nestedCalled);
```
misterdjules, line 50: I'd suggest adding a comment that this `assert` is done within a `nextTick` callback so that it runs after `A` completed, but before the next turn of libuv's event loop. Otherwise it's not clear for anyone but who wrote the code why it's done this way.

test/simple/test-timers-nested.js
```diff
@@ -0,0 +1,50 @@
...
+var scenarios = [0, 100];
+
+scenarios.forEach(function (delay) {
+  var nestedCalled = false;
+
+  setTimeout(function A() {
+    // Create the nested timer with the same delay as the outer timer so that it
+    // gets added to the current list of timers being processed by listOnTimeout.
+    setTimeout(function B() {
+      nestedCalled = true;
```
misterdjules, line 40: Also, I would add a `process.on('exit')` listener that makes sure that the nested callback is actually called before the process exits:

```
process.on('exit', function onExit() {
  assert.ok(nestedCalled);
});
```

misterdjules:
@whitlockjc This will look good to me once my latest two comments are addressed.

@nodejs/tsc Is it actually a rule now to submit PRs against master before landing changes in v0.12/v0.10?

If so that would be very tedious, since there's a lot of difference between some internal modules between v0.10/v0.12 and nodejs/master, and that would essentially mean forward-porting changes to nodejs/master and then back porting them again.

So unless there's a real benefit in doing that that I missed, I would suggest just landing this PR in v0.10, merging it in v0.12 and then forward-porting it in nodejs/master with a separate PR.

rvagg:
@misterdjules no need to add in master first where it doesn't make sense, sometimes things are just not going to belong in master at all and sometimes it's completely different code so separate PRs makes sense. So, no strict rules, don't jump through too many hoops if it's too complex, _particularly_ for 0.10 and 0.12 which are far detached now. Just keep a record in the commit of what happened so when we're cherry-picking things that we know what's going on.

misterdjules:
@rvagg Sounds good.

whitlockjc:
For posterity, I went ahead and updated this PR with the changes @misterdjules requested in https://github.com/nodejs/node/pull/3063

oleksiyk:
Can this still be a problem if the clock in the process is not monotonically updated at some moment? For example when the computer hibernates and then wakes up or the system time being updated by NTP daemon? 

This problem https://github.com/petkaantonov/bluebird/issues/1034 with bluebird promise looks to be related.

whitlockjc:
To test, you could build node with this PR applied to see if it still happens: https://github.com/nodejs/node/pull/3063  Long story short, the whole purpose of this PR is to make it so that a timer added within a timer gets fired on the next tick instead of the tick it was added.  That being said, I'm not sure why that would cause the issue you're seeing but I'd be willing to look deeper.

oleksiyk:
Thanks, I tried 0.10.41 which as I understand already has this PR applied and the issue with promises call stack still happens. Its hard to reproduce so I'm trying to narrow down the circumstances that can replicate it.

whitlockjc:
I don't think it's been merged to any stream yet.  Might tray a custom build.  The easiest thing would be to clone [my repo](https://github.com/whitlockjc/node), checkout the `merge-archive-pr-25763` branch and build it from there.  Let me know what you find out.

PR #25767, opened 2015-07-25
doc: buffer, adding missing backtick

dyanarose:
adding a missing backtick to fix the formatting in buf.toString


jasnell:
LGTM

jasnell:
Landed in 1a82c83

PR #25772, opened 2015-07-27
https: globalAgent keeps the connection alive.

### LINKED ISSUE #25749: regression: https.get(url, ...) returns partial content
Hi,

It seems that the https module is broken for Node.js 0.12.7, it may also be the case for other 0.12.x versions of Node, but I haven't built them yet.

I am getting some content which is available via HTTP or HTTPS, and using the `http` and the `https` modules does not give me the same result, while Firefox, wget and Node.js 0.10.39 do.

Here is the piece of code reproducing the bug:

``` js
var http = require("http");
var https = require("https");

var httpString;
var httpsString;

var onend = function() {
  if(httpString && httpsString) {
    console.log("httpsString equals to httpString: " + (httpsString == httpString));
    console.log("httpsString is smaller than httpString: " + (httpsString.length < httpString.length));
    console.log("httpsString is the beginning of httpString: " + (httpString.indexOf(httpsString) == 0));
  }
};

http.get("http://ccapi-preprod.cleverapps.io/v2/application.wadl", function(res) {
  var data = "";

  res.on("data", function(chunk) {
    data += chunk;
  });

  res.on("end", function() {
    httpString = data;
    onend();
  });
});

https.get("https://ccapi-preprod.cleverapps.io/v2/application.wadl", function(res) {
  var data = "";

  res.on("data", function(chunk) {
    data += chunk;
  });

  res.on("end", function() {
    httpsString = data;
    onend();
  });
});
```

We expect the following output (and this is what we get with Node.js 0.10.39 and 0.10.40):

```
httpsString equals to httpString: true
httpsString is smaller than httpString: false
httpsString is the beginning of httpString: true
```

And this is what we actually get with Node.js 0.12.7:

```
httpsString equals to httpString: false
httpsString is smaller than httpString: true
httpsString is the beginning of httpString: true
```

Technical information:
OS: Linux mufasa 4.0.7-2-ARCH #1 SMP PREEMPT Tue Jun 30 07:50:21 UTC 2015 x86_64 GNU/Linux
Node.js: 0.12.7, installed with nvm (as for Node.js 0.10.39)

rbelouin:
Some https requests returns truncated content, although everything works
when requesting the same resource with http. This regression appears
since 15a5a4a, and making the https globalAgent keep the connection
alive fix it.

Closes #25749

### DIFF
--- a/lib/https.js
+++ b/lib/https.js
@@ -122,6 +122,7 @@ Agent.prototype.getName = function(options) {
 };
 
 var globalAgent = new Agent();
+globalAgent.keepAlive = true;
 
 exports.globalAgent = globalAgent;
 exports.Agent = Agent;
--- a/test/simple/test-regress-GH-25749.js
+++ b/test/simple/test-regress-GH-25749.js
@@ -1,0 +1,81 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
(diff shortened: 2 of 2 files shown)


rbelouin:
There is a bit more information about this bug in this issue: #25749 

I need some help for mocking the HTTPS call (I know making a test dependent from a remote service is not a good idea).

If @isaacs could take a look at it, that would be great, as this PR is related to one of his commit.

jasnell:
@indutny @trevnorris @chrisdickinson ... any thoughts on this one?

crazyohpooh (approved):

Steven-hsm:
i don't know the reason

PR #25780, opened 2015-07-29
Node v0.12.7 ported to Visual Studio 2015

springmeyer:


common.gypi
```diff
@@ -89,7 +89,7 @@
         ],
         'msvs_settings': {
           'VCCLCompilerTool': {
-            'RuntimeLibrary': 0, # static release
```
piscisaureus: Why did you switch to a dll version of the crt?

springmeyer:
Very sorry. I did not meant to create this pull here (against joyent/node), but rather against mapbox/node - purely to keep track of the changes we need to carry (like the dll version).

common.gypi
springmeyer: @piscisaureus The Mapbox Studio app (https://www.mapbox.com/mapbox-studio/) uses Node. And it depends on node c++ modules that need to be linked to the dll version (and not statically linked to the c++ runtime) like node-mapnik. So, I build node against the same dll to ensure all C++ code is using the same exact one. Anyway, I realize this is not what node core is going to do and I did not mean for this pull to be against joyent/node as per my comment at https://github.com/joyent/node/pull/25780#issuecomment-126083384. Sorry for the noise.

BTW, this pull does have some fixes needed to compile against Visual 2015 which I will - at some point when I have time - extract and provide as fixes separately.

Makefile
```diff
@@ -343,7 +343,7 @@ tar: $(TARBALL)
 $(BINARYTAR): release-only
 	rm -rf $(BINARYNAME)
 	rm -rf out/deps out/Release
-	$(PYTHON) ./configure --prefix=/ --download=all --with-intl=small-icu \
+	$(PYTHON) ./configure --prefix=/ --download=all --with-intl=none \
```
srl295: I know this is closed, but why `none` out of curiosity? (saves only 4Mb, opposite from how Node binaries ship, turns off Internationalization API…)

PR #25819, opened 2015-08-06
REPL: tab complete for Arrays

jasnell:
Slight modified version of @seebees original commit:

https://github.com/seebees/node-1/commit/d624a683c160422ca03e7fd78a16686a5154756f

The original message:
  The tab complete for Arrays will
  include any integer indexes.  This is
  suboptimal because obj.21 is not
  valid syntax and a large array will
  contain a very cluttered list.

This alternative tweaks the original commit to use
Array.isArray and isNaN for the filter. Everything
else in untouched.

### DIFF
--- a/lib/repl.js
+++ b/lib/repl.js
@@ -602,7 +602,19 @@ REPLServer.prototype.complete = function(line, callback) {
           // if (e) console.log(e);
 
           if (obj != null) {
-            if (util.isObject(obj) || util.isFunction(obj)) {
+            if (Array.isArray(obj)) {
+              // Array elements will contain many integer values
+              // if we return them in memberGroups they will appear
+              // as obj.0 obj.100 etc. which is not valid syntax.
+              // Additionally for large arrays this can make for a
+              // very cluttered list.  So I filter them out.
+              memberGroups
+                .push(Object
+                      .getOwnPropertyNames(obj)
+                      .filter(function(item) {
+                        return isNaN(parseInt(item));
+                      }));
+            } else if (util.isObject(obj) || util.isFunction(obj)) {
               memberGroups.push(Object.getOwnPropertyNames(obj));
             }
             // works for non-objects
--- a/test/simple/test-repl-tab-complete-array-remove-integer-elements.js
+++ b/test/simple/test-repl-tab-complete-array-remove-integer-elements.js
@@ -1,0 +1,107 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+var common = require('../common');
+var assert = require('assert');
+var util = require('util');
+
+var repl = require('repl');
+
+// A stream to push an array into a REPL
+function ArrayStream() {
+  this.run = function(data) {
+    var self = this;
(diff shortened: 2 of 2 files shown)


jasnell:
@joyent/node-tsc .. quick review?

lib/repl.js
```diff
@@ -602,7 +602,19 @@ REPLServer.prototype.complete = function(line, callback) {
...
+              // Array elements will contain many integer values
+              // if we return them in memberGroups they will appear
+              // as obj.0 obj.100 etc. which is not valid syntax.
+              // Additionally for large arrays this can make for a
+              // very cluttered list.  So I filter them out.
+              memberGroups
+                .push(Object
+                      .getOwnPropertyNames(obj)
+                      .filter(function(item) {
+                        return isNaN(parseInt(item));
```
cjihrig: Maybe use `+item` instead of `parseInt()`. For example, `parseInt('3c')` returns 3.

lib/repl.js
```diff
@@ -602,7 +602,19 @@ REPLServer.prototype.complete = function(line, callback) {
           // if (e) console.log(e);
 
           if (obj != null) {
-            if (util.isObject(obj) || util.isFunction(obj)) {
+            if (Array.isArray(obj)) {
+              // Array elements will contain many integer values
+              // if we return them in memberGroups they will appear
+              // as obj.0 obj.100 etc. which is not valid syntax.
+              // Additionally for large arrays this can make for a
+              // very cluttered list.  So I filter them out.
```
cjihrig: Can you change the language on the last two sentences to get rid of the `I`.

test/simple/test-repl-tab-complete-array-remove-integer-elements.js
```diff
@@ -0,0 +1,107 @@
+// Copyright Joyent, Inc. and other Node contributors.
```
cjihrig: Do we still do the header in the joyent/node branches?

test/simple/test-repl-tab-complete-array-remove-integer-elements.js
```diff
@@ -0,0 +1,107 @@
...
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+var common = require('../common');
+var assert = require('assert');
+var util = require('util');
+
```
cjihrig: No need for the blank line here.

lib/repl.js
jasnell: Yeah, that's actually a 'feature' ;-) ... consider.. `var m = []; m['123abc'] = 'a'` .. if you do tab completion, the `123abc` and use `+item`, the `123abc` would still appear, which ends up having the same problem this patch is trying to fix. Using `parseInt` filters those out, even if they are valid keys for the object.

lib/repl.js
jasnell: heh, yes, I completely ignored the comments in the original patch.. I'll edit those.

test/simple/test-repl-tab-complete-array-remove-integer-elements.js
jasnell: Good point

lib/repl.js
cjihrig: Ah, carry on then :-)

seebees:
@jasnell How cool is that!  thanks!  What's the best way to get a PR reviewed?

jasnell:
In an ideal world, all PR's will be reviewed ;-) ... some just may take longer than others. This particular one is being actively reviewed now. Barring any issues, it should land shortly.

test/simple/test-repl-tab-complete-array-remove-integer-elements.js
```diff
@@ -0,0 +1,107 @@
...
+  'asdf.' ];
+
+var object_elements = [
+  [ 'asdf.__defineGetter__', 'asdf.__defineSetter__',
+    'asdf.__lookupGetter__', 'asdf.__lookupSetter__',
+    'asdf.__proto__', 'asdf.constructor',
+    'asdf.hasOwnProperty', 'asdf.isPrototypeOf',
+    'asdf.propertyIsEnumerable',
+    'asdf.toLocaleString', 'asdf.toString',
+    'asdf.valueOf', '', 'asdf.1'],
```
thefourtheye: Hmmm `asdf.1` is still invalid

jasnell:
Will be updating this shortly...

test/simple/test-repl-tab-complete-array-remove-integer-elements.js
jasnell: Yes, for consistency, all the other tests still have this header.

test/simple/test-repl-tab-complete-array-remove-integer-elements.js
jasnell: Looking into the test... it appears that somewhere along the line, test-repl-tab-complete stopped working properly. The entire test is not run and exits with a success without executing all of the code. Investigating now

cjihrig: The issue with the test exiting part of the way through has been fixed in the converged repo.

jasnell:
Closing here in favor of https://github.com/nodejs/node/issues/2409, which is a more complete approach. That will need to be cherry picked back to v0.12 and v0.10 at some point

PR #25822, opened 2015-08-07
 { encodeURIComponent: 'gbkEncodeURIComponent' }

zhenhua32:
{ encodeURIComponent: gbkEncodeURIComponent }, need ' '.
and I am not sure, I get it when I run the code in win10
w=%E4%B8%AD%E6%96%87&foo=bar

### DIFF
--- a/doc/api/querystring.markdown
+++ b/doc/api/querystring.markdown
@@ -29,7 +29,7 @@ Example:
     // Suppose gbkEncodeURIComponent function already exists,
     // it can encode string with `gbk` encoding
     querystring.stringify({ w: '中文', foo: 'bar' }, null, null,
-      { encodeURIComponent: gbkEncodeURIComponent })
+      { encodeURIComponent: 'gbkEncodeURIComponent' })
     // returns
     'w=%D6%D0%CE%C4&foo=bar'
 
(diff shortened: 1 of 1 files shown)


doc/api/querystring.markdown
```diff
@@ -29,7 +29,7 @@ Example:
     // Suppose gbkEncodeURIComponent function already exists,
     // it can encode string with `gbk` encoding
     querystring.stringify({ w: '中文', foo: 'bar' }, null, null,
-      { encodeURIComponent: gbkEncodeURIComponent })
+      { encodeURIComponent: 'gbkEncodeURIComponent' })
```
cjihrig: I think this is intentional. The comment above says that this is supposed to be a function that already exists.

zhenhua32:
// Suppose gbkEncodeURIComponent function already exists
sorry, I didn't see it, maybe someday it wiil support other encoding

zh-h:
@zhenhua32 Because default encodeURIComponent   is `escape()`,This is my [solution](http://applehater.cn/2015/08/31/node-post-urlencode-gbk/):

``` javascript
querystring.stringify(form, null, null,  
  { encodeURIComponent: function (str){
    var chinese = new RegExp(/[^\x00-\xff]/g);
      var gbkBuffer = null;
      var i = 0;
      var tempStr = '';
      if (chinese.test(str)){
        gbkBuffer = iconv.encode(str,'gbk');
      for (i;i<gbkBuffer.length;i++){
        tempStr += '%' + gbkBuffer[i].toString(16);
      };
      tempStr = tempStr.toUpperCase();
        return tempStr;
    }else{
        return querystring.escape(str);
    }
  }
});
```

`function (str){/* */}` is the gbkEncodeURIComponen,whitch contain function likes querystring.escape for gbk.

Issue #25831, opened 2015-08-10
Incorrect loop->active_handles count when nested setTimeout calls have same value


abbr:
I maintain a c++ module [deasync](https://github.com/abbr/deasync) (originally written by @vkurchatkin) which essentially exposes `uv_run` to jsLand in order to address some use cases that cannot be handled by other async-2-sync modules. A problem was reported recently that when `deasync` is involved in nested `setTimeout` calls with same timeout value, the process hung. It is illustrated by code

```
var deasync = require('deasync');

function async(cb) {
  console.log('x');
  setTimeout(function () {
    console.log('y');
    cb(null, 'value');
    console.log('z');
  }, 8);
}

setTimeout(function () {
  console.log('A', deasync(async)());
}, 8); // changing 8 to, say 9, works

```

output

```
x
(hung)
```

Changing one of the timeout values of `setTimeout` yields expected output

```
x
y
z
A value
```

Problem is reproducible in latest Node version on at least Linux and Windows. I have run `node` in gdb session and found the problem is `loop->active_handles` has gone down to 0 after the inner `setTimeout` is called but before its handler is triggered, therefore the inner handler doesn't have a chance to run.

I think the cause is that timers of same timeout value are implemented using a linked list and only one handle is exposed to libuv. This handle is removed at the time the outer `setTimeout` handler is called.

By moving `list.start(msecs, 0)` in `/lib/timers.js` outside else block, I have verified the problem can be fixed. Will supply a PR later. What I am not clear is the impact of this change. Appreciate if someone can review the issue and my PR.

(P.S. I have read Node issue reporting guidelines and understand I am not supposed include dependencies. But `deasync` is simple, has no dependency by itself, and is easy to be re-factored into a dependency-less mixture of c++ and js code)

tunniclm:
Could be related to #25763

misterdjules:
@tunniclm Although it seems at first sight, this issue is not related to #25763 :) @abbr's analysis is spot on.

While `deasync` seems to rely too much on Node.js' internals, the fix suggested in #25832 actually looks good to me so far, and thus I think I would be inclined to merge it in.

Thank you both @abbr @tunniclm for your input!

kjhangiani:
Has there been any progress on this?  Does this fix have other potential consequences?

PR #25832, opened 2015-08-10
timer: call list.start regardless new or not

abbr:
Call start regardless whether list is new
or not to prevent incorrect active_handles
count.

Fixes issue #25831.


garrettsickles:
What a great fix! This should be included asap!

snuggles08:
+1

misterdjules:
To paraphrase my comment in #25831:

> While `deasync` seems to rely too much on Node.js' internals, the fix suggested in [this PR] actually looks good to me so far, and thus I think I would be inclined to merge it in.

jasnell:
This looks good but would you be able to include a test case for it?

abbr:
@jasnell,  a test case would require either writing a c++ module or obtaining `deasync` via npm. I cannot find any existing cases under `test/simple` setup this way. In `test/addons` there are test cases written in c++ but that folder doesn't seem to be included in `make test`. Do you have any suggestion where to put the test case?

jonathanleang:
:+1:

jonathanleang:
:shipit:

thefourtheye:
cc @Fishrock123

vkurchatkin:
If it's impossible to write a test without external dependencies, then it's not a bug

Fishrock123:
I think this is a bug. My tired self does not think that the proposed patch could actually fix this. We'll see tomorrow maybe.

abbr:
I have added a test case, without any dependencies. Confirmed test case caused existing version 0.12.7 hung but passed with the fix.

Fishrock123:
@abbr could we move this issue to https://github.com/nodejs/node? I'd prefer to review it there.

vkurchatkin:
@Fishrock123 so what's the bug? running `uv_run` is clearly not supported so it should be something else

abbr:
@Fishrock123, I have ported this pr to https://github.com/nodejs/node/pull/2788

ChALkeR:
Closing here.

Issue #8877, opened 2014-12-16
Port #8666 from v0.10 to v0.12


misterdjules:
Recently, when merging v0.10 into v0.12, the commits from #8666 could not be merged because v0.10 and v0.12 diverged a lot in the areas that this PR touches.

Thus, we should manually forward port #8666 from v0.10 to v0.12 as soon as possible.

misterdjules:
I'm available to guide anyone willing to pick this issue up and submit a PR that fixes it. Ping me on GitHub by mentioning @misterdjules or find me on IRC (jgi in #libuv on Freenode).

whitlockjc:
If no one else has jumped on this, I'd love to do this.

misterdjules:
@whitlockjc That's awesome, thank you! :+1:

Just a few recommendations before you tackle this one:

  1. I would start by getting `test/simple/test-domain-top-level-error-handler-throw.js` to work in v0.12.      Just `git checkout upstream/v0.10 test/simple/test-domain-top-level-error-handler-throw.js` from a local checkout of v0.12, build node and run `./node test/simple/test-domain-top-level-error-handler-throw.js` to reproduce the problem.

  Make sure you understand the goal of this test as described in the inline comments. The changes that fix this test in the original PR (#8666) are only the changes made to ` process._fatalException` in `src/node.js`. The implementation of domains and the way exceptions are handled by node is a bit different in v0.12 compared to v0.10, so even though the changes you'll need to make will be similar, they will probably need to be adapted/moved to the proper place.

  2. Once 1) is done, I would try to get `test/simple/test-domain-with-abort-on-uncaught-exception.js` to pass too in the same manner.  In the original PR (#8666), the changes that fix this problem are all the changes but the ones in src/node.js. Basically, in v0.10 it meant adding a `SetAbortOnUncaughtException` function to V8's APIs, which sets a callback that [is called when a JavaScript exception is thrown](https://github.com/joyent/node/pull/8666/files#diff-2f958c3c5c7c9b1aa11a1f2297548145R1164). That callback determines if either there's no domain active currently, or if the active domain is at the top of the domains stack. If so it aborts, if not it doesn't. I believe we can probably adapt this almost as is for V8 and node in v0.12.

  Once that test pass, you'll want to make sure that when the process aborts, it aborts at the right time. One way to do this is to run a node program similar to the following:

  ```
var domain = require('domain');
var d = domain.create();
var fs = require('fs');
d.on('error', function() {
[...]

PR #25835, opened 2015-08-10
domains: port fix abort on uncaught to v0.12

whitlockjc:
Port fbff705 and caeb677 from v0.10 to v0.12, original commit messages:

  fbff705
  Add v8::Isolate::SetAbortOnUncaughtException() so the user can be
  notified when an uncaught exception has bubbled.

  caeb677
  Do not abort the process if an error is thrown from within a domain,
  an error handler is setup for the domain and
  --abort-on-uncaught-exception was passed on the command line.

  However, if an error is thrown from within the top-level domain's
  error handler and --abort-on-uncaught-exception was passed on the
  command line, make the process abort.

Fixes: #8877

### DIFF
--- a/deps/v8/src/isolate.cc
+++ b/deps/v8/src/isolate.cc
@@ -1090,19 +1090,30 @@ void Isolate::DoThrow(Object* exception, MessageLocation* location) {
         thread_local_top()->pending_message_end_pos_ = location->end_pos();
       }
 
-      // If the abort-on-uncaught-exception flag is specified, abort on any
-      // exception not caught by JavaScript, even when an external handler is
-      // present.  This flag is intended for use by JavaScript developers, so
-      // print a user-friendly stack trace (not an internal one).
+      // If the abort-on-uncaught-exception flag is specified, and if the
+      // exception is not caught by JavaScript (even when an external handler is
+      // present).
       if (fatal_exception_depth == 0 &&
           FLAG_abort_on_uncaught_exception &&
           (report_exception || can_be_caught_externally)) {
-        fatal_exception_depth++;
-        PrintF(stderr,
-               "%s\n\nFROM\n",
-               MessageHandler::GetLocalizedMessage(this, message_obj).get());
-        PrintCurrentStackTrace(stderr);
-        base::OS::Abort();
+        // If the embedder didn't specify a custom uncaught exception callback,
+        // or if the custom callback determined that V8 should abort, then
+        // abort
+        bool should_abort = !abort_on_uncaught_exception_callback_ ||
+                             abort_on_uncaught_exception_callback_(
+                                 reinterpret_cast<v8::Isolate*>(this)
+                             );
+
+        if (should_abort) {
+          fatal_exception_depth++;
+          // This flag is intended for use by JavaScript developers, so
+          // print a user-friendly stack trace (not an internal one).
+          PrintF(stderr,
+                 "%s\n\nFROM\n",
+                 MessageHandler::GetLocalizedMessage(this, message_obj).get());
+          PrintCurrentStackTrace(stderr);
+          base::OS::Abort();
+        }
       }
     } else if (location != NULL && !location->script().is_null()) {
       // We are bootstrapping and caught an error where the location is set
@@ -1299,6 +1310,12 @@ void Isolate::SetCaptureStackTraceForUncaughtExceptions(
 }
 
 
+void Isolate::SetAbortOnUncaughtException(
+      v8::Isolate::abort_on_uncaught_exception_t callback) {
+  abort_on_uncaught_exception_callback_ = callback;
+}
+
+
 Handle<Context> Isolate::native_context() {
   return handle(context()->native_context());
 }
@@ -1474,7 +1491,8 @@ Isolate::Isolate()
       num_sweeper_threads_(0),
       stress_deopt_count_(0),
       next_optimization_id_(0),
-      use_counter_callback_(NULL) {
+      use_counter_callback_(NULL),
+      abort_on_uncaught_exception_callback_(NULL) {
   id_ = base::NoBarrier_AtomicIncrement(&isolate_counter_, 1);
   TRACE_ISOLATE(constructor);
 
--- a/src/node.cc
+++ b/src/node.cc
@@ -909,6 +909,33 @@ Local<Value> WinapiErrnoException(Isolate* isolate,
 #endif
 
 
+static bool IsDomainActive(const Environment* env) {
+  if (!env->using_domains()) {
+    return false;
+  }
+
+  Local<Array> domain_array = env->domain_array().As<Array>();
+  uint32_t domains_array_length = domain_array->Length();
+  if (domains_array_length == 0)
+    return false;
+
+  Local<Value> domain_v = domain_array->Get(0);
+  return !domain_v->IsNull();
+}
+
+
+bool ShouldAbortOnUncaughtException(v8::Isolate* isolate) {
+  Environment* env = Environment::GetCurrent(isolate);
+  Local<Object> process_object = env->process_object();
+  Local<String> emitting_top_level_domain_error_key =
+    env->emitting_top_level_domain_error_string();
+  bool isEmittingTopLevelDomainError =
+    process_object->Get(emitting_top_level_domain_error_key)->BooleanValue();
+
+  return !IsDomainActive(env) || isEmittingTopLevelDomainError;
+}
+
+
 void SetupDomainUse(const FunctionCallbackInfo<Value>& args) {
   Environment* env = Environment::GetCurrent(args.GetIsolate());
 
@@ -2772,6 +2799,9 @@ void SetupProcessObject(Environment* env,
 
   // pre-set _events object for faster emit checks
   process->Set(env->events_string(), Object::New(env->isolate()));
+
+  process->Set(env->emitting_top_level_domain_error_string(),
+               False(env->isolate()));
 }
 
 
@@ -3452,6 +3482,8 @@ void Init(int* argc,
   // even when we need it to access it from another (debugger) thread.
   node_isolate = Isolate::New();
   Isolate::Scope isolate_scope(node_isolate);
+
+  node_isolate->SetAbortOnUncaughtException(ShouldAbortOnUncaughtException);
 
 #ifdef __POSIX__
   // Raise the open file descriptor limit.
--- a/test/simple/test-domain-with-abort-on-uncaught-exception.js
+++ b/test/simple/test-domain-with-abort-on-uncaught-exception.js
@@ -1,0 +1,233 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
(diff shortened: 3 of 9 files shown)


misterdjules:
Thank you @whitlockjc! Tests running on [UNIX](http://jenkins.nodejs.org/job/node-test-commit-unix/150/) and on [Windows](http://jenkins.nodejs.org/job/node-test-commit-windows/148/).

misterdjules:
@orangemocha I've re-enabled the node-accept-pull-request job on jenkins.nodejs.org to test this PR because when I tried to use the node-test-pull-request job on jenkins-iojs.nodesource.com, it failed: https://jenkins-iojs.nodesource.com/job/node-test-pull-request/63/.

misterdjules:
@whitlockjc `test-domain-with-abort-on-uncaught-exception.js` is failing on [Linux](http://jenkins.nodejs.org/job/node-test-commit-unix/150/DESTCPU=ia32,label=linux/tapTestReport/simple.tap-173/) and [SmartOS](http://jenkins.nodejs.org/job/node-test-commit-unix/150/DESTCPU=ia32,label=smartos/tapTestReport/simple.tap-173/).

Fixing them should be straightforward, could you please update this PR with a fix when you have some time? Thank you!

whitlockjc:
You got it.  I'm sure their failures are similar to the fix for OS X where we have to check for different exit codes.  I'll update the PR.

orangemocha:
> @orangemocha I've re-enabled the node-accept-pull-request job on jenkins.nodejs.org to test this PR because when I tried to use the node-test-pull-request job on jenkins-iojs.nodesource.com, it failed: https://jenkins-iojs.nodesource.com/job/node-test-pull-request/63/.

The issue should be fixed: https://jenkins-iojs.nodesource.com/job/node-test-pull-request/65/

In any case, you can use node-accept-pull-request from the new Jenkins. I have re-disabled the old one now. Let me know if you still have problems. Thanks!

misterdjules:
A fix for similar issue(s) was merged in io.js a while ago: https://github.com/nodejs/io.js/pull/922. That fix has the advantage of not requiring to make any change to V8, so it would be great if we could use it instead of the changes in this PR.

However, it seems that that fix does not fix https://github.com/joyent/node/issues/8630:
```
➜  v0.12 git:(v0.12) ✗ cat ~/dev/repros/domain-top-level-error-handler-throw.js
var domainErrHandlerExMessage = 'exception from domain error handler';
var internalExMessage = 'You should NOT see me';

var d = domain.create();

d.on('error', function() {
    console.log('in domain error handler');
    throw new Error(domainErrHandlerExMessage);
});

d.run(function doStuff() {
    process.nextTick(function () {
      throw new Error(internalExMessage);
    });
});
➜  v0.12 git:(v0.12) ✗ ./node ~/dev/repros/domain-top-level-error-handler-throw.js
in domain error handler
/Users/JulienGilli/dev/repros/domain-top-level-error-handler-throw.js:14
      throw new Error(internalExMessage);
            ^
Error: You should NOT see me
    at /Users/JulienGilli/dev/repros/domain-top-level-error-handler-throw.js:14:13
    at process._tickDomainCallback (node.js:389:11)
    at Function.Module.runMain (module.js:503:11)
    at startup (node.js:129:16)
    at node.js:822:3
➜  v0.12 git:(v0.12) ✗
```

It seems to also make a node process not abort when it seems it should:

```
➜  v0.12 git:(v0.12) ✗ cat ~/dev/repros/domain-abort-on-uncaught.js
var d = domain.create();
var triggeredProcessUncaughtException = false;

process.on('uncaughtException', function onUncaughtException() {
  console.log('In uncaughtException handler');
});

d.on('error', function() {
  console.log('Throwing from top level domain error handler');
  throw new Error('Error from domain error handler');
});

d.run(function doStuff() {
  // Throwing from within different types of callbacks as each of them
  // handles domains differently
  process.nextTick(function () {
    throw new Error("Error from nextTick callback");
  });

  var fs = require('fs');
  fs.exists('/non/existing/file', function onExists(exists) {
    throw new Error("Error from fs.exists callback");
  });

  setImmediate(function onSetImmediate() {
    throw new Error("Error from setImmediate callback");
  });

  throw new Error("Error from domain.run callback");
});
➜  v0.12 git:(v0.12) ✗ ./node --abort-on-uncaught-exception ~/dev/repros/domain-abort-on-uncaught.js
Throwing from top level domain error handler
In uncaughtException handler
Throwing from top level domain error handler
In uncaughtException handler
Throwing from top level domain error handler
Throwing from top level domain error handler
In uncaughtException handler
Throwing from top level domain error handler
Throwing from top level domain error handler
Throwing from top level domain error handler
In uncaughtException handler
➜  v0.12 git:(v0.12) ✗ 
```
In the test above, I would expect node to abort, not the `uncaughtException` handler to be called. But I may misinterpret the semantics of `process.on('uncaughtException')` and `--abort-on-uncaught-exception`.

@chrisdickinson Thoughts?

whitlockjc:
I have updated the tests to pass on Linux and SmartOS.  This involved always checking the exit signal, even though SmartOS is the only one right now that gets a signal other than `null`, and updating Linux to have variable exit codes.  I also updated the comments explaining why we use arrays of possible values for exit codes/signals.

misterdjules:
Thank you @whitlockjc!

Reran the whole tests suite here: https://jenkins-iojs.nodesource.com/job/node-accept-pull-request/. We have [one failure on Windows](https://jenkins-iojs.nodesource.com/job/node-test-commit-windows/nodes=win2008r2/134/tapTestReport/test.tap-173/). This is also related to changes in the implementation of `OS::Abort` in V8 from v0.10 to v0.12.

whitlockjc:
I'll get on it.

whitlockjc:
Fixed, @misterdjules was right...this was a result of `OS::Abort` changing its exit code so we now need to check for multiple values like we do on *nix.  I also cleaned up a few places in the test where lines were greater than 80 characters.

misterdjules:
@whitlockjc Thank you! [All tests passed](https://jenkins-iojs.nodesource.com/job/node-test-commit/163/).

whitlockjc:
WOOHOO!!!

jasnell:
@misterdjules ... is this ready to land?

whitlockjc:
I don't want this to drop through the cracks so can we update this with the next steps?  Really hoping to either find out that this isn't necessary in the new node repository or be told what the next steps are.  If it's still useful/important, I'll gladly port this to the new repository via a PR.

/cc @nodejs/collaborators and @nodejs/tsc

misterdjules:
@whitlockjc I've ported this PR in nodejs/node's master with https://github.com/nodejs/node/pull/3036 because of https://github.com/nodejs/node/issues/3035.

The next step for this PR is for me to give it a final review and land it in v0.12.

whitlockjc:
Thanks a lot.  I was planning on porting this today but since you beat me to it...  ;)

deps/v8/src/isolate.cc
```diff
@@ -1090,19 +1090,30 @@ void Isolate::DoThrow(Object* exception, MessageLocation* location) {
...
-      // If the abort-on-uncaught-exception flag is specified, abort on any
-      // exception not caught by JavaScript, even when an external handler is
-      // present.  This flag is intended for use by JavaScript developers, so
-      // print a user-friendly stack trace (not an internal one).
+      // If the abort-on-uncaught-exception flag is specified, and if the
+      // exception is not caught by JavaScript (even when an external handler is
+      // present).
       if (fatal_exception_depth == 0 &&
-          FLAG_abort_on_uncaught_exception &&
+          (FLAG_abort_on_uncaught_exception) &&
```
misterdjules: There should be no parentheses around `FLAG_abort_on_uncaught_exception`.

src/node.cc
```diff
@@ -909,6 +909,33 @@ Local<Value> WinapiErrnoException(Isolate* isolate,
 #endif
 
 
+static bool IsDomainActive(const Environment* env) {
+  if (!env->using_domains()) {
+    return false;
+  }
+
+  Local<Array> domain_array = env->domain_array().As<Array>();
+  uint32_t domainsArrayLength = domain_array->Length();
```
misterdjules: Use snake case for consistency for `domainsArrayLength` -> `domains_array_length`, or just get rid of that temporary, it seems unneeded.

test/simple/test-domain-with-abort-on-uncaught-exception.js
```diff
@@ -0,0 +1,229 @@
...
+    // handles domains differently
+    process.nextTick(function () {
+      throw new Error("Error from nextTick callback");
+    });
+
+    var fs = require('fs');
+    fs.exists('/non/existing/file', function onExists(exists) {
+      throw new Error("Error from fs.exists callback");
+    });
+
```
misterdjules, line 87: I would also throw from within a `setTimeout`'s callback, as these are handled in a different code path as the others, and thus handle domains specifically too.

Something like:

```
setTimeout(function() { throw new Error("Error from setTimeout's callback"); }, 0);
```

misterdjules:
@whitlockjc Added more comments.

Also, it would be great to split this PR in two different commits, the first one for the changes in deps/v8, and the second one for the changes outside of deps/v8. This way we can isolate and document this new floating patch in https://github.com/nodejs/node-v0.x-archive/wiki/V8-upgrade-process#v012x.

When these are addressed we can land this PR in v0.12. Thank you for your work again! :+1:

whitlockjc:
You got it.  I'll get on it.

deps/v8/src/isolate.cc
whitlockjc: Yup, that was an oversight...probably a relic from previous debugging session.

src/node.cc
whitlockjc: Agreed, my bad.

test/simple/test-domain-with-abort-on-uncaught-exception.js
whitlockjc, line 87: You got it.

whitlockjc:
Since you've already got a PR for this _(https://github.com/nodejs/node/pull/3036)_, how do you want me to handle this?  Create a new PR with two separate commits and that will replace the aforementioned PR?

misterdjules:
@whitlockjc We still don't know if we'll want to go with nodejs/node#3036 or nodejs/node#3038, or something else. So I would suggest just updating this PR.

whitlockjc:
Okay.  Thanks for the guidance Julien.

whitlockjc:
I've addressed the changes @misterdjules suggested.  I will not break the singular port commit into two port commits.  When this is complete, I will update you.

whitlockjc:
I have broken the PR commit into two commits, one for `v8` and one for `domains`.  I retained the original commit messages in their respective port commits but the subject of each port commit has been shortened to fit.

misterdjules:
LGTM. Thank you once again @whitlockjc!

misterdjules:
Landing with https://ci.nodejs.org/job/node-accept-pull-request/129/.

misterdjules:
Landed in 1982ed6e6356070241d695f77888a7681fd3c23c and f0453caea2c47405625236cf42ae11ae2a857142.

whitlockjc:
Is there a next step for getting this into nodejs/node?

misterdjules:
@whitlockjc These two commits are in nodejs/node's v0.12 branch. I'll follow-up on https://github.com/nodejs/node/pull/3036 when I get some feedback about the V8 changes in https://codereview.chromium.org/1375933003/. If that takes too long, or if it turns out https://github.com/nodejs/node/pull/3036 is not the solution we want to consider, we'll consider the other options mentioned in https://github.com/nodejs/node/pull/3035.

whitlockjc:
Gotcha.  Thanks for the heads up and let me know if I can help further regardless of the decisions mentioned above.

PR #8761, opened 2014-11-21
src/node.cc: fix OpenSSL build error

joerg-krause:
Since version 0.10.33 (commit d601c76f4d728dd7adfa2fbbed2fe86de2e6b479) building nodejs without OpenSSL is broken. This issue is already reported upstream: https://github.com/joyent/node/issues/8676.

This issue is fixed by this patch.

Signed-off-by: Jörg Krause jkrause@posteo.de

Issue #8676, opened 2014-11-05
v0.10.33 does not build when openssl disabled


lovebug356:
v0.10.33 does not build when openssl disabled, v0.10.32 builds fine if openssl is disabled.

> In file included from ../src/node.cc:85:0:
> ../src/node_crypto.h:30:25: fatal error: openssl/ssl.h: No such file or directory
> #include <openssl/ssl.h>

tedu:
Still an error with v0.12.0. Same fix.

misterdjules:
@lovebug356 @joerg-krause @jsiei97 @tedu Thank you for your help, and sorry for the delay. Adding this issue to the v0.10.40 release milestone, because we want to get v0.10.39 out as quickly as possible due to OpenSSL vulnerabilities and we don't want to add anything more to that milestone.

misterdjules:
I'm available to guide anyone willing to pick this issue up and submit a PR that fixes it.

whitlockjc:
I'd like to take a shot at this.  I'll get started and will hit you up with any questions I come up with.

misterdjules:
@whitlockjc Excellent! Thank you :+1:

whitlockjc:
I can reproduce with the latest `v0.10` branch using `./configure --without-ssl && make`:

```
Undefined symbols for architecture x86_64:
  "node::SSL2_ENABLE", referenced from:
      node::Init(int, char**) in node.o
  "node::SSL3_ENABLE", referenced from:
      node::Init(int, char**) in node.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I'll get a PR together shortly.

joerg-krause:
This is the patch we use to build nodejs 0.10.40 with Buildroot: http://git.buildroot.net/buildroot/tree/package/nodejs/0.10.40/0004-fix-build-error-without-OpenSSL-support.patch

And this is the PR I already did to fix this issue: https://github.com/joyent/node/pull/8761

whitlockjc:
Cool, I didn't see your work before.  I'll stop working on this then.  I got pulled into other work yesterday so after reproducing, I didn't work on the fix.

whitlockjc:
@joerg-krause Your PR in #8761 seems to reference a deleted repository and cannot be reopened.  Do you want to close that PR and create a new one, linking to this issue?  If not, I don't mind creating a new PR to fix this issue I just don't want to do it without giving you the chance to.

[...]

PR #25862, opened 2015-08-15
src/node.cc: fix build error without OpenSSL support

joerg-krause:
Resubmit PR https://github.com/joyent/node/pull/8761 since the forked repo was closed by myself. This fixes the issue reported upstream: https://github.com/joyent/node/issues/8676

Commit d601c76f4d728dd7adfa2fbbed2fe86de2e6b479 introduced a bug which includes `node_crypto.h` twice. Once conditionally with `HAVE_OPENSSL` and once without it:

``` c
#if HAVE_OPENSSL
# include "node_crypto.h"
#endif
[...]
#include "node_crypto.h"
```

Fix this by removing the unconditional header include of `node_crypto.h` and add conditions for the OpenSSL related parts.

### DIFF
--- a/src/node.cc
+++ b/src/node.cc
@@ -82,7 +82,6 @@ typedef int mode_t;
 #include "node_script.h"
 #include "v8_typed_array.h"
 
-#include "node_crypto.h"
 #include "util.h"
 
 using namespace v8;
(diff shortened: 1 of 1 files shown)


jasnell:
LGTM

jasnell:
Landed in 16ca077

PR #25868, opened 2015-08-17
Orangemocha opensusegcc

martynbm:



orangemocha:
@martynbm it looks like this was opened by mistake?

Issue #25709, opened 2015-07-17
Investigate v0.10 flaky test test-http-pipeline-flood


orangemocha:
https://github.com/joyent/node/commit/a2f879f197f76d91a42ec0f8322f18948911a552 marked this test as flaky in v0.10 so that it doesn't block merges with Jenkins and node-accept-pull-request.

As a follow up, we should investigate why this test is failing in v0.10. Some of the tests marked as flaky in https://github.com/joyent/node/commit/a2f879f197f76d91a42ec0f8322f18948911a552 fail intermittently, while others fail persistently. For the latter, an explicit skip for the specific platform might be more appropriate.

My guess is that a good percentage of these tests fail because of issues with the test itself, not node. It's likely that those issues were fixed in v0.12, so it might be a good idea to look for changes to the same test in git history in v0.12, or even simply trying the v0.12 version of the test with node v0.10.

/cc: @joyent/node-collaborators

dnakamura:
I've already been looking at this one. I can say that the test is flaky on windows due to the very small tcp buffer sizes. Also the test is not particularly well written. I'm working on a rewrite to better test desired behavior

dnakamura:
Also, in v0.12 releases the number of requests asserted has been lowered to 250 instead of 500 (e3ec2f7dab). This seems to fix on platforms other than windows. Ultimately, asserting a specific number of requests does not make much sense, as in my testing when the desired behavior does not exists the test runs indefinitely

misterdjules:
@orangemocha @dnakamura FYI, marking the test as flaky for Windows in v0.12 with https://github.com/joyent/node/pull/25707.

misterdjules:
Created https://github.com/joyent/node/issues/25732 for Windows and v0.12.

PR #25870, opened 2015-08-17
Fix test-http-pipeline-flood.js

### LINKED ISSUE #25732: Fix test-http-pipeline-flood being flaky on Windows in v0.12
https://github.com/joyent/node/pull/25707 marked `test-http-pipeline-flood` as flaky on Windows so the rest of the changes could land.

@dnakamura started to investigate the root cause in https://github.com/joyent/node/issues/25709 and seems to have made good progress.

dnakamura:
Fixes #25732 and #25709

So, some background. This test looks to test a feature to prevent a DoS vulnerability. Essentially once native socket write buffers have filled up, the http parser should stop reading in new data. (requests in data which has already been read will still be processed)
## Current test
- sets up a http server, with a timeout which destroys the connection after 200ms of inactivity.
- creates a child which continuously fires requests and does not read responses.
- On exit asserts that:
  - Inactivity timeout has fired
  - >250 requests have been processed
  - The child process terminated
### Issues:
- Flaky on windows because of very small socket buffers (write buffers fill after only a few responses), depending on timing this means that only ~40 requests will get processed
- Does not make much sense to assert number of requests greater than a threshold when testing a DoS prevention feature.
- If feature under test not working, the test case runs until program runs out of memory 
## New test
- Set up server without inactivity timer
- Create child process like before, but with timeout to kill itself after  specified amount of time
- When server sends response, check if native buffer filled (res.write() returns false). On first time:
  - Add listener for data events on the socket which asserts false (backlog logic should have corked the stream so it should never be called)

### DIFF
--- a/test/simple/test-http-pipeline-flood.js
+++ b/test/simple/test-http-pipeline-flood.js
@@ -38,22 +38,27 @@ function parent() {
   var childClosed = false;
   var requests = 0;
   var connections = 0;
+  var backloggedReqs = 0;
 
   var server = http.createServer(function(req, res) {
     requests++;
     res.setHeader('content-length', bigResponse.length);
-    res.end(bigResponse);
+    if(!res.write(bigResponse)){
+      if(backloggedReqs == 0){
+        //No new data should arrive after our responses start backing up
+        req.socket.on('data', function(){assert(false)})
+      }
+      backloggedReqs++;
+    }
+    res.end();
   });
 
   server.on('connection', function(conn) {
     connections++;
   });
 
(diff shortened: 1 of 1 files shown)


jasnell:
@dnakamura ... hey, can I trouble you to open this against the v0.12 branch instead? And, if appropriate, against nodejs/node master if it's an issue there too? We're not going to be landing PRs on master here.

ChALkeR:
This is also reported as https://github.com/nodejs/node/pull/2862, no need to keep it open here.

PR #25875, opened 2015-08-18
Update README.md

ghost:


README.md
```diff
@@ -8,8 +8,8 @@ Prerequisites (Unix only):
     * GCC 4.2 or newer
     * G++ 4.2 or newer
     * Python 2.6 or 2.7
-    * GNU Make 3.81 or newer
-    * libexecinfo (FreeBSD and OpenBSD only)
+    * Windows and Linux and MacOS
```
KenanY: Seems a little silly to add this under "Prerequisites (Unix only)". And why remove `make`?

alexrepin:
libexecinfo is a library
that's on FreeBSD and not yet but soon OpenBSD
because this is the task
why only GNU Make linux?
There are Windows and MacOS and other linuxes aren't they?

jasnell:
Thanks for the PR. I have not yet reviewed it so I cannot comment on the details, but given that we're in the middle of transitioning to http://github.com/nodejs/node, we won't be landing PRs against master here. If you'd like, please open a new PR either against the v0.12 branch or against master on nodejs/node

jasnell:
As mentioned, this won't land on master in this repo. Please open a new PR against master on nodejs/node if the issue is still relevant there.

PR #25886, opened 2015-08-21
stream: watch removeListener on readable event

### LINKED ISSUE #7678: No 'readable' events after unshift-ing more than highWaterMark bytes
In the following test, if toUnshift is less than 16_1024, the program writes messages without end, as pretended.
But if toUnshift is greater than or equal to 16_1024, the program stops after the first 'readable' event.

This program is the simplest I could write to exhibit this strange behaviour.
It occurred to me in a real program that accepts several protocols in the same port; these protocols can be optionally compressed and encrypted with ssl.
The program explores the first bytes to determine the protocol, unshifts the read data and jumps to the found protocol.

``` javascript
var net = require('net');

var toUnshift = 16*1024-1;

var buf = new Buffer(65536);
for (var i = 0; i < buf.length; i++) buf[i] = i;

var pr;
var server = net.createServer();
server.on('connection', function(c) {
  console.log('New connection');
  c.on('end', function() {console.log('Connection ended');});
  pr = processReadable.bind(null, c);
  c.on('readable', pr);
});
server.listen(12345, function() {console.log("Server bound");});

sendTo("localhost", 12345);

var first = true;
function processReadable(c) {
  console.info("Readable");
  var data = c.read();
  console.info("Read: " + data.length);
  if (first) {
    first = false;
    c.removeListener('readable', pr);
    c.unshift(buf.slice(0, toUnshift));
    c.on('readable', processReadable.bind(null,c));
    var delayedRead = function() {
      var data = c.read();
      console.info("Delayed read: " + data.length);
    };
    process.nextTick(delayedRead);
  }
}

function sendTo(host, port) {
  var s = net.connect(port, host);
  s.on('drain', startSender);
  startSender();

  function startSender() {process.nextTick(doSend);}

  function doSend() {
    while (s.write(buf));
  }
}
```

plafer:
Removing a readable listener now updates the readable state i.e.
readableListening, needReadable and emittedReadable are set to false.
Then, if a readable listener is added at a later time, the stream will
know none are attached and will take proper action to get the stream
going back again.

Fixes #7678

### DIFF
--- a/lib/_stream_readable.js
+++ b/lib/_stream_readable.js
@@ -691,7 +691,7 @@ Readable.prototype.on = function(ev, fn) {
           self.read(0);
         });
       } else if (state.length) {
-        emitReadable(this, state);
+        setImmediate(emitReadable.bind(null, this));
       }
     }
   }
@@ -699,6 +699,17 @@ Readable.prototype.on = function(ev, fn) {
   return res;
 };
 Readable.prototype.addListener = Readable.prototype.on;
+
+Readable.prototype.removeListener = function(type, listener) {
+  Stream.prototype.removeListener.call(this, type, listener);
+
+  if (type === 'readable' && EE.listenerCount(this, type) === 0) {
+    var state = this._readableState;
+    state.readableListening = false;
+    state.emittedReadable = false;
+    state.needReadable = false;
+  }
+};
 
 // pause() and resume() are remnants of the legacy readable stream API
 // If the user uses them, then switch into old mode.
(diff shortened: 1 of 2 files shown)


jasnell:
Overall this looks like a good change. Would like to get @chrisdickinson and @trevnorris to take a look also tho. 

One thing, however, I see that this is opened against master. Because we are in mid-transition to moving things over to http://github.com/nodejs/node, you may want to open this against the v0.12 branch here and against master there. We are not going to be landing fixes into master in this repo.

plafer:
Alright, I'll make a new PR pointing to v0.12 if the feedback @chrisdickinson and @trevnorris is positive. Will also replicate changes on master branch in the new repo also.

ChALkeR:
This is also reported as https://github.com/nodejs/node/pull/2616, no need to keep it open here.

PR #25888, opened 2015-08-21
Update dns.markdown

PixnBits:
fixed heading level
see TOC at https://nodejs.org/api/dns.html


jasnell:
Thank you for the PR. Right now we're in the middle of transitioning to http://github.com/nodejs/node and we're not going to be landing PRs against master here. Could I ask you to re-open this PR against the v0.12 branch or the master on nodejs/node?

PixnBits:
Looks like I don't need to :+1:
https://github.com/nodejs/node/commit/df0d790107edf635dc233f3338b3c2e68db58cc7

PR #25898, opened 2015-08-26
add headers length check

hustxiaoc:
add headers length check


jasnell:
@hustxiaoc ... thank you for the PR. Quick question: is there a problem that you're trying to fix with this? There is nothing wrong with the check itself, but I'm just not sure it's necessary. 

Also, since we are in mid transition moving active development over to http://github.com/nodejs/node master, we're not going to be landing any PR's against master here. This PR would need to be moved over to that other repo and closed here.

hustxiaoc:
@jasnell  not really a big problem，but  an empty array still satisfies the condition

PR #25903, opened 2015-08-27
timers: Add instanceof check for all 'Timeout' methods

Sergii-Babych:
During the call of 'Timeout' member method, the call context may have
incorrect values or the  context  itself  can  be  completely  wrong.
It cause crash  or  hang  or  unknown error. In order to prevent this
situation  in  future  we  need  to  add  call  context  check.

Fixes #4261

### DIFF
--- a/lib/timers.js
+++ b/lib/timers.js
@@ -312,7 +312,18 @@ function unrefdHandle() {
 }
 
 
+var Timeout_call_context_error = "Trying to call 'Timeout' member method on non 'Timeout' context";
+
 Timeout.prototype.unref = function() {
+  // Prevent call 'Timeout' member method on non 'Timeout' context.
+  // For example this is wrong:
+  //   var t = setInterval(function() {}, 1);
+  //   process.nextTick(t.unref);
+  // But this is ok:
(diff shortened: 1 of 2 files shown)


Fishrock123:
I'm really not convinced this is necessary. Should we just add these to all prototype methods? I mean, of course it's going to fail if the context isn't a timeout. :S

Sergii-Babych:
Yes, it is going to fail and there is a test calling unref in non Timeout context. It was working differently on different platforms and we spend much time in order to figure out why, so I think it's worth having this check here.
Also it fixes test with undefined behaviour.

For more information please see last two comments here:
https://github.com/joyent/node/pull/5102#issuecomment-132719843

cjihrig:
I agree with @Fishrock123. It seems a bit excessive to add so many checks here and not do it throughout the rest of the codebase.

jasnell:
Either way, this isn't something that would land here. If this change is made, it needs to be against master in nodejs/node. Going to close this issue off here and encourage @Sergii-Babych to open a new PR against nodejs/node if he'd like to pursue it further.

PR #25915, opened 2015-09-04
deps: upgrade to npm 2.14.2

zkat:
_THIS UPDATE INCLUDES A SECURITY FIX_

Hooray, shiny new npm! See the release notes at https://github.com/npm/npm/releases/tag/v2.14.2 for details!

Aside from that, we fixed up how `preferGlobal` warnings was working because the unexpected warnings were messing with Windows PowerShell.

The extra commit you see is just a patch to our test suite to make it pass when you `make test-npm` -- a recent patch made it so `.npmrc` is never included when you pack a repo, and that broke the test suite when not testing directly in source :)

**r**: @jasnell


targos:
@zkat I think that now 0.12 will be maintained in nodejs/node.
This repo is basically read-only.

zkat:
@jasnell Where should I be putting this LTS PR?

ChALkeR:
@zkat [nodejs/node, v0.12 branch](https://github.com/nodejs/node/tree/v0.12)

